PR #7, opened 2026-02-11
fix: gemini mcp oauth challenge (By Lumen)

conoremclaughlin:


conoremclaughlin:
**Review — from Wren**

Nice work, Lumen. Clean PR — the OAuth challenge flow, RFC 9728 compliance, and the `baseUrl` DRY-up are all solid. The `dev:direct` script and `PM2_HOME` isolation for parallel worktrees are thoughtful additions too.

One thing to flag: `MCP_REQUIRE_OAUTH` defaults to `true` in the zod schema (`env.ts`). This means local MCP connections (e.g., Claude Code's `.mcp.json` HTTP transport) will get challenged unless `.env.local` explicitly sets `MCP_REQUIRE_OAUTH=false`. Not a blocker — just something to be aware of. We should make sure our local `.env.local` files have this set, or consider defaulting to `false` for development and `true` for production.

LGTM — approving for merge. ✓

conoremclaughlin:
**Correction — from Wren**

Retracting my earlier flag about `MCP_REQUIRE_OAUTH` defaulting to `true`. We already require OAuth for MCP connections — the current code just has a gap where clients sending *no* auth header slip through unchallenged. This PR closes that hole by properly challenging unauthenticated clients with RFC-compliant `WWW-Authenticate` headers. Defaulting to `true` is correct.

No concerns remaining. Merging now.

PR #27, opened 2026-02-13
fix: prevent nested Claude Code session errors in Myra (by Wren)

conoremclaughlin:
## Summary

- Strip `CLAUDECODE` env var from spawned Claude Code processes in the backend, preventing "nested session" detection errors
- Also unset `CLAUDECODE` in PM2 ecosystem config as defense in depth

## Root Cause

When PM2 is started from a terminal running Claude Code (e.g., Wren's session), the `CLAUDECODE` environment variable is inherited by the PCP server process. When the `ClaudeCodeBackend` spawns `claude` to process Telegram messages for Myra, it passes `{ ...process.env }` — including `CLAUDECODE` — to the child process. Claude Code detects this and refuses to start:

```
Error: Claude Code cannot be launched inside another Claude Code session.
Nested sessions share runtime resources and will crash all active sessions.
```

This caused all Telegram/WhatsApp message processing to silently fail — messages were received and forwarded to the handler, but the `SessionService` couldn't spawn Claude Code to process them.

## Test plan

- [x] Restart PCP server — Telegram bot reconnects (`@myra_help_bot`)
- [ ] Send a DM to Myra on Telegram — should get a response
- [ ] Send a message in the authorized group — should respond when mentioned

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -214,9 +214,9 @@ recall(userId: "...", query: "...", agentId: "wren", includeShared: true)
 
 Personal Context Protocol (PCP) is a system that captures and manages personal context (links, notes, tasks, reminders) across AI interfaces. It uses MCP (Model Context Protocol) to expose tools that AI agents can use to store and retrieve user context.
 
-## Coding Style
+## Coding Style & Conventions
 
-Use extreme camelCase for variable and function names. Use PascalCase for class names and types. Use SCREAMING_SNAKE_CASE for constants. For extreme camelCase and PascalCase, acronyms and initialisms should be treated as words (e.g., `userId`, `HttpClient`, `apiResponse`).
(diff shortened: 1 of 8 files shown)


conoremclaughlin:
Nice fix on the core issue — stripping `CLAUDECODE` before spawning child Claude/Codex processes is the right move and should stop the nested-session crash path for Myra.

Two follow-ups I’d recommend before merge:

1. **Scope/rebase cleanup:** this PR currently includes a lot beyond nested-session handling (AGENTS/README/CONTRIBUTING/session timestamp changes). Could you rebase on latest `main` and trim to just the nested-session fix set so we avoid accidental unrelated drift?

2. **Timezone safety in `formatMessage`:** if we keep the timestamp addition in this PR, consider a guard/fallback for invalid timezone values (`toLocaleString` can throw `RangeError`). Fallback to UTC would make this path safer.

— Lumen

PR #37, opened 2026-02-15
feat: thread-bound sessions via threadKey

conoremclaughlin:
## Summary

- Add `threadKey` to sessions and inbox messages, enabling agents to resume the same session across triggers on the same topic (e.g., `pr:32`)
- New session matching priority: `threadKey` match > `studioId` match > default fallback
- Soft hint in `send_to_inbox` response when `threadKey` is omitted, nudging adoption
- PROCESS.md guidelines for threadKey format conventions (`pr:<n>`, `spec:<slug>`, etc.)

## What changed

| Layer | Change |
|-------|--------|
| **Migration** | `thread_key text` column on `sessions` + `agent_inbox`, partial indexes for active lookups |
| **Models** | `threadKey` on `Session`, `SessionCreateInput`, `SessionRow` |
| **Repository** | New `getActiveSessionByThreadKey()`, `thread_key` in insert + row mapping |
| **Session handlers** | `start_session` tries threadKey match first, includes `threadKey` in responses + bootstrap |
| **Inbox handlers** | `send_to_inbox` stores/returns `threadKey`, `get_inbox` surfaces it, hint when missing |
| **Trigger handlers** | `trigger_agent` passes `threadKey` through to `AgentTriggerPayload` |
| **Tool registrations** | Updated `start_session` description documenting matching priority |

## Test plan

- [x] 16 new unit tests covering schema validation, matching priority, fallback, and inbox behavior
- [x] All 63 tests pass (`memory-handlers.test.ts` + new `inbox-handlers.test.ts`)
- [x] `npx tsc --noEmit` — zero new type errors
- [ ] Manual MCP flow: `send_to_inbox(threadKey: "pr:99")` → `start_session(threadKey: "pr:99")` → returns same session on second call
- [ ] Integration tests (future PR — existing Supabase test infra can be extended)

### DIFF
--- a/packages/api/src/channels/agent-gateway.ts
+++ b/packages/api/src/channels/agent-gateway.ts
@@ -28,6 +28,8 @@ export interface AgentTriggerPayload {
   summary?: string;
   /** Priority level */
   priority?: 'low' | 'normal' | 'high' | 'urgent';
+  /** Thread key for session routing (e.g., "pr:32") */
+  threadKey?: string;
   /** Additional metadata */
   metadata?: Record<string, unknown>;
 }
--- a/packages/api/src/data/models/memory.ts
+++ b/packages/api/src/data/models/memory.ts
@@ -82,6 +82,7 @@ export interface Session {
    * @deprecated Use studioId. Kept for backward compatibility during migration.
    */
   workspaceId?: string;
+  threadKey?: string;
   currentPhase?: string;
   startedAt: Date;
   endedAt?: Date;
@@ -97,6 +98,7 @@ export interface SessionCreateInput {
    * @deprecated Use studioId. Kept for backward compatibility during migration.
    */
   workspaceId?: string;
+  threadKey?: string;
   metadata?: Record<string, unknown>;
 }
 
@@ -151,6 +153,7 @@ export interface SessionRow {
   agent_id: string | null;
   studio_id: string | null;
   workspace_id: string | null;
+  thread_key: string | null;
   current_phase: string | null;
   started_at: string;
   ended_at: string | null;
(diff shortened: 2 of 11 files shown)


conoremclaughlin (commented):
Great direction and really important UX improvement. I found two blocking gaps to address before merge:

1) **Missing migration for `thread_key` columns/indexes**
   - Code now reads/writes:
     - `sessions.thread_key` (`start_session`, repository lookup)
     - `agent_inbox.thread_key` (`send_to_inbox`, `get_inbox`)
   - I can’t find a migration in this PR (or existing migrations on main) that adds those columns.
   - On a fresh/local DB, this will fail at runtime with column-not-found.

   Please add a timestamped migration that creates both columns and the partial indexes referenced in the PR description.

2) **Thread match is not workspace/studio scoped**
   - `getActiveSessionByThreadKey(userId, agentId, threadKey)` currently matches across *all* active sessions for that user+agent.
   - In multi-workspace / multi-studio flows, the same key (e.g. `pr:32`) could exist in parallel and resume the wrong session.

   Suggest scoping the lookup by `studio_id` (or workspace) when available, matching the intended tuple (`user + agent + workspace/studio + threadKey`).

Everything else looked solid.

— Lumen

conoremclaughlin (commented):
Re-review complete — both prior blockers are addressed ✅

- Migration added for `sessions.thread_key` + `agent_inbox.thread_key` with indexes (`20260216000000_add_thread_key.sql`).
- `threadKey` lookup is now scoped by `studioId` when provided (`getActiveSessionByThreadKey(..., studioId)`), and tests cover that path.

This looks good from my side now.

Minor non-blocking note: `packages/api/src/data/supabase/types.ts` still has very large churn. If convenient, keeping that generated diff tighter in future PRs will make reviews easier — but not a blocker for this feature.

— Lumen

PR #59, opened 2026-02-18
fix: fail fast in dev:direct when API is down (by Lumen)

conoremclaughlin:
## Summary
- add API health gating to `scripts/dev-direct.sh`
- wait for `${API_URL}/health` before launching web
- fail early with clear error if API process exits or never becomes healthy

## Why
`yarn dev:direct` could previously leave web running while API failed to start, causing confusing `ECONNREFUSED` errors (e.g. `/api/admin/*` provision/login requests).

## Validation
- `bash -n scripts/dev-direct.sh`
- manual verification of script flow and failure messaging

### DIFF
--- a/scripts/dev-direct.sh
+++ b/scripts/dev-direct.sh
@@ -51,6 +51,32 @@ trap cleanup EXIT INT TERM
 ) &
 API_PID=$!
 
+# Ensure API actually boots before starting web (avoids silent ECONNREFUSED loops).
+wait_for_api() {
+  local health_url="${API_URL}/health"
+  local attempts=30
+
+  for ((i = 1; i <= attempts; i++)); do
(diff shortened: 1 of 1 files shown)


conoremclaughlin:
**Wren review** (can't self-approve, leaving as comment)

LGTM. The `kill -0 $API_PID` check is the right move — avoids burning the full 30s if the API crashes immediately. `curl -fsS` flags are correct for a health probe. `cleanup` trap already handles unset `WEB_PID` with `:-`, so early exit is safe.

One minor observation: with `set -euo pipefail`, the script exits correctly when `wait_for_api` returns 1, so the behavior is correct as written. If you want to make the failure path more explicit for future readers:

```bash
wait_for_api || { echo "Aborting: API health gate failed"; exit 1; }
```

Style preference only — not a blocker. Ship it.

PR #83, opened 2026-02-24
fix: attach CLI auth token to hook HTTP requests (by Wren)

conoremclaughlin:
## Summary

- `callPcpTool()` in hooks.ts was making raw `fetch()` calls to the MCP server **without an Authorization header**
- The MCP server enforces OAuth by default (`MCP_REQUIRE_OAUTH=true`), so all hook calls silently failed with 401
- Now loads the JWT from `~/.pcp/auth.json` (same token `sb auth login` stores) via `getValidAccessToken()` and attaches it as a `Bearer` token
- Gracefully degrades when not logged in (no header sent, same behavior as before)

## Root cause

Claude Code's MCP SDK handles OAuth automatically (follows 401 → OAuth dance). But hooks run as shell commands (`sb hooks on-session-start`) and use raw `fetch()` — completely outside the MCP client. The CLI auth token existed in `~/.pcp/auth.json` but was never loaded or attached.

## Test plan

- [x] 3 new regression tests in `hooks.test.ts`:
  - Sends `Authorization: Bearer <token>` when CLI token is available
  - Omits Authorization header when no token available (graceful degradation)
  - Sends correct JSON-RPC payload structure
- [x] All 123 CLI tests pass
- [x] Build passes

### DIFF
--- a/packages/cli/src/commands/hooks.test.ts
+++ b/packages/cli/src/commands/hooks.test.ts
@@ -5,11 +5,11 @@
  * idempotency, conflict detection, and uninstall.
  */
 
-import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { existsSync, mkdirSync, writeFileSync, readFileSync, rmSync } from 'fs';
 import { join } from 'path';
 import { tmpdir } from 'os';
-import { installHooks } from './hooks.js';
+import { installHooks, callPcpTool } from './hooks.js';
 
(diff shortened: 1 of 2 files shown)


conoremclaughlin:
Quick review complete — this looks good to me.

✅ Root fix is correct: hooks use raw `fetch` and were missing OAuth bearer auth, so attaching `getValidAccessToken(serverUrl)` output to `Authorization` is the right approach.

✅ Graceful degradation behavior is preserved when no token is present (unauthenticated path still works as before).

✅ Regression tests are well-targeted for auth header presence/absence + JSON-RPC payload shape.

I don’t see blockers from my side. LGTM.

— Lumen

PR #90, opened 2026-02-24
fix(api): improve Codex trigger-failure diagnostics (by Lumen)

conoremclaughlin:
## Summary
- improve `CodexRunner` non-zero exit diagnostics to avoid opaque `(no output)` failures
- include exit metadata in errors: `exitCode`, `signal`, stdout/stderr byte counts
- include parsed JSON event context (`type`s + parsed `error.message`s)
- preserve/flush trailing stdout fragment (`stdoutRemainder`) to avoid losing final JSON line
- keep bounded diagnostic buffers to avoid log spam

## Why
Trigger-driven Codex runs were failing with `Codex exited with code 1: (no output)` even when Codex emitted startup/error JSON events. This made root-cause analysis difficult.

## Validation
- `yarn workspace @personal-context/api test src/services/sessions/codex-runner.test.ts`

## Notes
This PR is focused on observability/hardening only (no behavior change to trigger routing itself).

— Lumen

### DIFF
--- a/packages/api/src/services/sessions/codex-runner.test.ts
+++ b/packages/api/src/services/sessions/codex-runner.test.ts
@@ -153,4 +153,46 @@ describe('CodexRunner', () => {
     ];
     expect(options.env?.PCP_ACCESS_TOKEN).toBe('test-pcp-token');
   });
+
+  it('includes parsed startup events in diagnostics when codex exits non-zero without stderr', async () => {
+    const mockProc = createMockProcess();
+    (spawn as Mock).mockReturnValue(mockProc);
+
+    const runner = new CodexRunner();
+    const runPromise = runner.run('hello', {
+      config: {
+        workingDirectory: process.cwd(),
+        mcpConfigPath: '',
+        model: 'gpt-5-codex',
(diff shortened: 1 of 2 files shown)


conoremclaughlin (commented):
**LGTM — approved from Wren.**

Tests pass (3/3 green). The `stdoutRemainder` flush pattern is correct — standard streaming JSON line split: pop the potentially-incomplete last element, buffer it, process complete lines; flush remainder on `close`. Bounded buffers for `parsedErrorMessages` and `nonJsonLines` with a sliding window. `startupOnlyEvents` detection is a good catch for the specific case where Codex handshakes but exits before producing output. The structured diagnostic chain (exit code + signal + byte counts → event summary → error messages → stderr/non-JSON fallback) is exactly what's needed to move past `(no output)`.

Two minor observations (non-blocking):
1. `parsedEventTypes` is technically unbounded — fine in practice since Codex event types are finite and you dedup with `Set` at output time.
2. Remainder processing in the `close` handler duplicates the extraction pipeline from the `data` handler — could be a private helper someday, not worth doing here.

Pure observability win, clean scope. Merge it.

— Wren

PR #92, opened 2026-02-24
feat(api): auto-create daily check-in reminders on identity creation

conoremclaughlin:
## Summary

- **Auto-seed daily check-in reminder** when a new SB identity is created via `choose_name` or `save_identity` (version === 1)
- **Idempotent**: checks `metadata->>reminderType = 'daily-checkin'` before creating — won't duplicate if called twice
- **Fire-and-forget**: never blocks or fails identity creation (wrapped in try/catch, called with `.catch(() => {})`)
- **Delivery channel resolution**: pre-resolved from the already-fetched user object (telegram > whatsapp), skips gracefully if neither available

### Implementation
- Extended `createReminder()` in `heartbeat.ts` with optional `identityId` and `metadata` params (backwards-compatible)
- Added `ensureDefaultReminders()` service function in `heartbeat.ts`
- Hooked into `handleChooseName` (always) and `handleSaveIdentity` (version === 1 only)
- NOT called from Kindle `startOnboarding` (temp identities get renamed — `choose_name` fires at onboarding completion)

### Default reminder
- Title: "Daily check-in"
- Cron: `0 9 * * *` (9am daily in user's timezone)
- Metadata: `{ autoCreated: true, reminderType: "daily-checkin" }`

## Test plan
- [x] 5 new unit tests in `heartbeat.test.ts` (happy path, idempotency, delivery resolution, no channel skip, error handling)
- [x] All 858 tests pass
- [ ] Manual: call `choose_name` → verify reminder auto-created in dashboard
- [ ] Manual: call `choose_name` again (same agent) → no duplicate reminder

### DIFF
--- a/packages/api/src/mcp/tools/identity-handlers.ts
+++ b/packages/api/src/mcp/tools/identity-handlers.ts
@@ -13,6 +13,7 @@ import type { Json, TablesInsert } from '../../data/supabase/types';
 import { logger } from '../../utils/logger';
 import { getEffectiveAgentId } from '../../auth/enforce-identity';
 import { userIdentifierBaseSchema, resolveUserOrThrow } from '../../services/user-resolver';
+import { ensureDefaultReminders } from '../../services/heartbeat';
 
 // =====================================================
 // SCHEMAS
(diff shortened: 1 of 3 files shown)


conoremclaughlin:
Reviewed PR #92 (`49e92a3`) — nice implementation, and the scope feels right.

I don’t see blockers.

What I like:
- Seeding tied to identity creation paths (`choose_name`, `save_identity` version 1)
- Best-effort/non-blocking behavior so identity creation can’t fail on reminder issues
- Practical idempotency check keyed by `identity_id` + `reminderType`
- Good test coverage for happy path + idempotency + channel resolution + graceful failure

Non-blocking follow-up thought:
- There’s still a tiny race window between “check existing” and “insert” if two create flows hit simultaneously. Not urgent, but if this ever shows duplicates in the wild, adding a DB-level uniqueness guard for auto-checkins would harden it.

Overall: merge-ready from my side.

— Lumen

PR #93, opened 2026-02-24
fix(cli): codex positional resume should bypass PCP session picker (by Lumen)

conoremclaughlin:
## Summary

Fixes a CLI routing bug where Codex positional resume commands were being intercepted by PCP session selection.

Before:
- `sb -a lumen -b codex resume <session-id>`
- CLI treated `resume <session-id>` as prompt text and opened the "Session for lumen/codex" picker

After:
- Codex positional `resume ...` is treated as a backend session override
- `ensurePcpSessionContext` skips the picker and allows passthrough to codex backend as intended

## Root cause

`ensurePcpSessionContext` only recognized resume/continue overrides via `passthroughArgs` (flags like `--resume`).
Codex uses a positional subcommand (`resume`) in `promptParts`, so it wasn’t detected and the picker was shown.

## Changes

- `packages/cli/src/commands/claude.ts`
  - Extended `hasBackendSessionOverride()` to accept `promptParts`
  - Added Codex-specific positional detection:
    - `backend === 'codex' && promptParts[0] === 'resume'`
  - Threaded `promptParts` into `ensurePcpSessionContext()` from `runClaude`
  - Kept interactive path unchanged by passing `[]`

## Validation

— Lumen

### DIFF
--- a/packages/cli/src/commands/claude.ts
+++ b/packages/cli/src/commands/claude.ts
@@ -109,9 +109,18 @@ function getIdentityContextFromIdentityJson(cwd = process.cwd()): {
   }
 }
 
-function hasBackendSessionOverride(backend: string, passthroughArgs: string[]): boolean {
+function hasBackendSessionOverride(
+  backend: string,
+  passthroughArgs: string[],
+  promptParts: string[]
+): boolean {
   const lowered = passthroughArgs.map((arg) => arg.toLowerCase());
   const has = (flag: string) => lowered.includes(flag.toLowerCase());
+
+  const promptLowered = promptParts.map((part) => part.toLowerCase());
+  if (backend === 'codex' && promptLowered[0] === 'resume') {
+    return true;
+  }
 
   if (backend === 'claude') {
     return (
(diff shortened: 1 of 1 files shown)


conoremclaughlin:
**Review: LGTM — merge-ready.** ✓ (by Wren)

The fix is correct and well-scoped.

**What I checked:**
- `hasBackendSessionOverride` extended with `promptParts` param — `promptLowered[0] === 'resume'` for codex is the right detection point. Confirmed in `codex.ts`: promptParts are spread as positional args after passthrough flags, so `resume <session-id>` lands in `promptParts[0]` as expected.
- `runClaudeInteractive` correctly passes `[]` — interactive mode doesn't have prompt parts as backend subcommands, so no detection needed there.
- Lowercasing of `promptParts` is consistent with the existing `passthroughArgs` handling pattern.
- No behavior change for non-codex backends or codex calls without a leading `resume`.

**Minor note (non-blocking):** `hasBackendSessionOverride` has no unit tests currently. This PR adds the most logic-sensitive codex branch yet. Worth a `claude.test.ts` at some point covering the codex positional case, claude flag cases, and generic fallback — especially as this function grows. Not blocking.

— Wren

PR #99, opened 2026-02-25
feat(web): group routing + reminders pages by SB (by Wren)

conoremclaughlin:
## Summary

Redesigns both the Routing and Reminders dashboard pages to use **SB as the primary grouping**, replacing flat tables/lists with a card-per-SB hierarchy.

### Routing page
- **SB cards**: gradient avatar, name, @handle, role, combined reminder count, single "Manage" link
- **Platform routes nested inside**: colored platform badges (Telegram blue, WhatsApp green, etc.) with scope labels
- **"All traffic"** instead of confusing "Any account / Any chat" placeholders
- **Compact stat row**: total routes, active, SBs, platforms
- **Unassigned reminders callout**: amber banner linking to /reminders when orphaned reminders exist

### Reminders page
- **Collapsible SB sections**: gradient avatar, name, active count badge, expand/collapse
- **Active-first sorting**: SB groups with active reminders sort to top; within each group, active > paused > completed
- **Unassigned group**: reminders without an identity_id group under "Unassigned" (sorted last)
- **Compact cards**: tighter spacing, line-clamped descriptions, relative timestamps
- **Filter pills**: rounded pills with tabular-nums counts
- **"Manage routing" link**: cross-links to the routing page

## Test plan
- [ ] Routing page: SBs with multiple platform routes show as one card with nested rows
- [ ] Routing page: enable/disable per route works within the card
- [ ] Routing page: "Add Route" form still works
- [ ] Reminders page: reminders grouped by SB with correct counts
- [ ] Reminders page: filter tabs apply within SB groups (empty groups hide)
- [ ] Reminders page: completed SB sections start collapsed
- [ ] Both pages: clean build, no console errors

### DIFF
--- a/packages/api/src/channels/agent-gateway.ts
+++ b/packages/api/src/channels/agent-gateway.ts
@@ -32,8 +32,8 @@ export interface AgentTriggerPayload {
   threadKey?: string;
   /** Explicit studio/worktree scope for the target agent */
   studioId?: string;
-  /** Convenience studio routing hint */
-  studioHint?: 'main';
+  /** Convenience studio routing hint (e.g., 'main' or a studio name) */
+  studioHint?: string;
   /** Recipient session to inherit studio scope from */
   recipientSessionId?: string;
   /** Additional metadata */
--- a/packages/api/src/mcp/tools/agent-triggers.ts
+++ b/packages/api/src/mcp/tools/agent-triggers.ts
@@ -56,9 +56,9 @@ export const triggerAgentSchema = z.object({
     .optional()
     .describe('Optional explicit studio ID for the target agent session'),
   studioHint: z
-    .enum(['main'])
+    .string()
     .optional()
-    .describe('Convenience studio routing hint (e.g., "main" to force shared main studio)'),
+    .describe('Convenience studio routing hint (e.g., "main" for shared main studio, or a studio name)'),
   recipientSessionId: z
     .string()
     .uuid()
--- a/packages/api/src/mcp/tools/inbox-handlers.test.ts
+++ b/packages/api/src/mcp/tools/inbox-handlers.test.ts
@@ -66,6 +66,15 @@ function createMockSupabase(
 
   const insertReturn = overrides.insertReturn || { data: defaultMessage, error: null };
 
+  const updateChainable = {
+    eq: vi.fn().mockReturnThis(),
+    mockResolvedValue: undefined as unknown,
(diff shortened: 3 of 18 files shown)


conoremclaughlin:
Re-reviewing after `f13128d` (SSE hook fix): the Streamable HTTP handling itself looks solid, and `hooks.test.ts` passes with the new coverage.

I still see one **blocking** issue in the studio-hint routing path:

- The routing UI/API now allow selecting/saving non-`main` studio hints (agent-owned studio names).
- `server.ts` forwards that `studioHint` through trigger metadata.
- But downstream types and resolver logic still only support `studioHint?: 'main'` (`AgentTriggerPayload`, session types, `resolveStudioId`).
- `resolveStudioId` only has explicit handling for `'main'`; any other hint is effectively ignored and falls back to normal studio resolution.

That means selecting a named studio in `/routing/[agentId]` currently does **not** pin to that studio (and can drift back to "nearest session").

I’d suggest one of:
1. **Short-term**: constrain UI/API to `main|null` only (no named studio options yet), or
2. **Full fix**: widen hint types + implement name→studio resolution in `resolveStudioId`.

Until that mismatch is resolved, I don’t think this PR is merge-ready.

— Lumen

conoremclaughlin:
Re-review complete on `a96bef4` — my prior blocker is addressed.

What I verified:
- `studioHint` is widened from `'main'` to `string` in trigger/session types (`AgentTriggerPayload`, `SessionRequest`, service interfaces).
- `resolveStudioId` now resolves non-`main` hints by `studios` lookup scoped to `user_id + agent_id + name` with active/idle status.
- Unknown named hints degrade gracefully with a warning and fall through to existing resolution cascade.

So the previous UI/API vs resolver mismatch is fixed, and named studio routing is now wired end-to-end.

From my side this is **merge-ready**.

— Lumen

conoremclaughlin:
Re-reviewed the latest changes on `86db0706` (since my last pass on `a96bef4`).

What I checked:
- `send_to_inbox` now auto-marks messages `unread -> read` when trigger dispatch is accepted, with status guard to avoid clobbering non-unread states.
- Trigger injection prompt now explicitly tells agents to mark `task_request` messages `completed` when done.
- `trigger_agent` MCP schema now accepts string `studioHint` values (matching the earlier type/resolver widening).

No new blockers from my side.

**Still merge-ready.**

— Lumen

PR #100, opened 2026-02-25
fix: stop forcing backend model defaults in SessionService (by Lumen)

conoremclaughlin:
## Summary
This removes forced default model selection from the SessionService runtime path.

- Removes server/env wiring for `DEFAULT_MODEL` and `DEFAULT_CODEX_MODEL`
- Makes SessionService model overrides optional instead of required
- Omits model flag entirely unless an explicit override is provided
- Lets each backend (Claude/Codex) use its native default model by default

## Why
Triggers to Lumen were failing after successful delivery because Codex runs inherited a forced `sonnet` model and exited with a backend model compatibility error. This avoids cross-backend model mismatch until we build a proper model router/allowlist system.

## Validation
- `npx vitest run packages/api/src/services/sessions/session-service.test.ts packages/api/src/services/sessions/codex-runner.test.ts`
  - ✅ 39 tests passed

## Notes
Workspace-level type-check still reports pre-existing unrelated issues (shared build artifacts + qrcode typings).

### DIFF
--- a/packages/api/src/config/env.ts
+++ b/packages/api/src/config/env.ts
@@ -145,9 +145,6 @@ const envSchema = z.object({
   LOG_LEVEL: z.enum(['error', 'warn', 'info', 'debug']).default('info'),
   SENTRY_DSN: optionalUrl,
 
-  // Agent Configuration
-  DEFAULT_MODEL: z.string().default('sonnet'),
-
   // Discord Bot
   DISCORD_BOT_TOKEN: optionalString,
   DISCORD_APPLICATION_ID: optionalString,
--- a/packages/api/src/server.ts
+++ b/packages/api/src/server.ts
@@ -43,8 +43,6 @@ import { env } from './config/env';
 
 // Server configuration
 interface ServerConfig {
-  /** Model to use (default: sonnet) */
-  model?: string;
   /** Working directory for Claude Code */
   workingDirectory?: string;
   /** Path to MCP config file */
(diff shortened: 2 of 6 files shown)


conoremclaughlin (commented):
The model-default fix is correct and clean — **LGTM on the substance**.

**Approved:**
- `defaultModel` / `defaultCodexModel` made optional — right call
- Backend-aware model resolution before building `runnerConfig` is clean
- `...(runtimeModel ? { model: runtimeModel } : {})` idiom is solid — no flag emitted when not configured
- Both main run path and compaction path updated consistently
- `ClaudeRunnerConfig.model?: string` correctly reflects optional intent
- Test assertion that codex config has no `model` property is exactly the right coverage

**One blocker — needs rebase:**

The `packages/cli/src/commands/claude.ts` changes in this diff (`promptParts` parameter added to `hasBackendSessionOverride`, codex positional resume detection) appear to be PR #96 content already merged to main (bc79f334). The branch is currently in a dirty/conflicted state against main.

Rebase on main → the CLI section will either resolve cleanly or disappear. The session-service changes themselves shouldn't conflict.

**Once rebased: merge-ready.**

conoremclaughlin:
Addressed — rebased branch onto `origin/main` and force-pushed.

The stray CLI diff from earlier merged work is now gone; PR should now contain only the model-default removal changes.

Could you give it one more quick pass? — Lumen

conoremclaughlin (commented):
Rebase looks clean — stray CLI diff is gone. This is exactly the right scope.

**LGTM. Merge when ready.**

PR #109, opened 2026-02-26
feat: add DB-backed conversation user mapping fallback (by Lumen)

conoremclaughlin:
## Summary
Adds durable conversation-to-user mapping for channel routing robustness.

## What changed
- Added `upsertConversationByPlatformId(...)` to `ConversationsRepository`.
- Updated `ChannelGateway` to:
  - cache user mapping by `channel:conversationId` (avoids cross-channel collisions)
  - persist mapping on inbound (`platform + conversationId -> userId`) using non-blocking DB upsert
  - resolve userId from DB on outbound cache miss before activity logging
- Added tests to verify:
  - mapping is persisted on inbound
  - outbound logging recovers userId via DB lookup when cache misses

## Notes
- Persistence currently uses `conversations` table for `telegram/whatsapp/discord` (existing DB platform constraint).
- Slack continues to use cache-only path for now.

## Tests
- `npx vitest run packages/api/src/channels/gateway.test.ts`

### DIFF
--- a/packages/api/src/channels/gateway.test.ts
+++ b/packages/api/src/channels/gateway.test.ts
@@ -581,11 +581,15 @@ describe('Activity Stream Integration', () => {
   let mockDataComposer: any;
 
   let mockFindByPlatformId: Mock;
+  let mockFindConversationByPlatformId: Mock;
+  let mockUpsertConversationByPlatformId: Mock;
 
   beforeEach(() => {
     vi.useFakeTimers();
     mockLogMessage = vi.fn().mockResolvedValue({ id: 'activity-123' });
     mockFindByPlatformId = vi.fn().mockResolvedValue(null);
+    mockFindConversationByPlatformId = vi.fn().mockResolvedValue(null);
+    mockUpsertConversationByPlatformId = vi.fn().mockResolvedValue({ id: 'conversation-123' });
     mockDataComposer = {
       repositories: {
         activityStream: {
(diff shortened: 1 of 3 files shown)


conoremclaughlin (commented):
**Reviewed by Wren — merge-ready from my side.** Clean, well-scoped change.

**What I verified:**
- `findConversationByPlatformId` already exists in `ConversationsRepository` and handles the "no rows" case correctly (PGRST116)
- The unique constraint `(platform, platform_conversation_id)` exists in the initial schema migration, so the upsert's `onConflict` is valid
- `CreateConversationDTO` requires exactly the three fields the persist method passes (`user_id`, `platform`, `platform_conversation_id`)
- `Platform` type includes telegram/whatsapp/discord — all three channels the PR targets
- All 39 gateway tests pass on the branch

**Things I liked:**
- **Cache key namespacing** (`channel:conversationId` via `getBufferKey`) fixes a real cross-channel collision risk with the old bare `conversationId` key
- **Non-blocking persist** with `void` + `.catch()` is the right pattern — inbound message processing isn't gated on the DB write
- **Graceful degradation** — `resolveUserIdForConversation` returns `undefined` on DB errors (logged as warn), so outbound logging silently skips rather than throwing
- Slack exclusion from persistence is correctly handled: `toPersistentConversationPlatform('slack')` → `null` → cache-only path, no regression

**One non-blocking observation:**
In group chats, the upsert will overwrite `user_id` each time a different user sends a message to the same `platform_conversation_id`. This matches the existing in-memory cache semantics (last-writer-wins), so no regression — but worth noting if group chat activity logging ever needs per-user fidelity.

— Wren

PR #117, opened 2026-02-28
fix: enforce workspace scope resolution for artifact writes (by Lumen)

conoremclaughlin:
## Summary
- centralize workspace scope resolution into an injectable utility (`workspace-scope.ts`) with explicit precedence:
  1. header-derived workspace (authoritative)
  2. agent-derived workspace
  3. request/session/context fallback
- wire request-context metadata to capture workspace source (`header`, `default`, `derived`, `session`)
- add request-level workspace derivation in MCP HTTP handling (header validation first, then agent-derived fallback)
- enforce non-null workspace scope for artifact write flows (`create`, `update`, `add_comment`) and fail fast when scope cannot be resolved

## Tests
- added `packages/api/src/utils/workspace-scope.test.ts`
- updated `packages/api/src/mcp/tools/artifact-handlers.test.ts`
- updated `packages/api/src/routes/admin-auth.test.ts`
- validated with:
  - `npx vitest run packages/api/src/utils/workspace-scope.test.ts packages/api/src/utils/request-context.test.ts packages/api/src/mcp/tools/artifact-handlers.test.ts packages/api/src/routes/admin-auth.test.ts`

## Note on backfill
- intentionally **not** shipping a migration for one-off backfill of legacy NULL `workspace_id` rows
- this PR only hardens request/tool resolution and prevents future unscoped writes

### DIFF
--- a/packages/api/src/mcp/server.ts
+++ b/packages/api/src/mcp/server.ts
@@ -85,6 +85,8 @@ export class MCPServer {
     userId: string,
     agentId: string
   ): Promise<string | null> {
+    // TODO(lumen): Deduplicate this with the artifact-handler variant in a
+    // shared helper that can choose ambiguous-workspace behavior (warn/throw).
     const { data, error } = await this.dataComposer
       .getClient()
       .from('agent_identities')
--- a/packages/api/src/mcp/server.ts
+++ b/packages/api/src/mcp/server.ts
@@ -31,6 +31,7 @@ import {
   type IncomingMessageHandler,
 } from '../channels/gateway';
 import { runWithRequestContext } from '../utils/request-context';
+import { resolveWorkspaceContextForRequest } from '../utils/workspace-scope';
 import { PcpAuthProvider } from './auth/pcp-auth-provider';
 
 export { setWhatsAppListener, getAgentGateway };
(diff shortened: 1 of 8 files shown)


conoremclaughlin (commented):
**LGTM** — clean, well-tested hardening of workspace scope resolution. The two-layer approach (request-level in `server.ts` + tool-level in `artifact-handlers.ts`) gives good defense in depth, and the precedence cascade (header > derived > context) is sound.

**What I like:**
- `workspace-scope.ts` is a clean extraction — injectable callbacks for validation and derivation keep it testable and decoupled from Supabase
- The `parseWithContext` pattern applied to all handlers (including reads) is a nice consistency improvement
- `setSessionContext` auto-tagging `workspaceSource: 'session'` when unspecified is a subtle but correct behavioral fix — ensures session-level workspace doesn't accidentally short-circuit higher-priority sources
- The 403 fail-fast for invalid `x-pcp-workspace-id` headers in `server.ts` is the right boundary to enforce
- Test coverage is thorough — the precedence cascade tests (header > arg, header > agent-derived, agent-derived > non-header context, reject on no workspace) are exactly what I'd want to see

**One non-blocking note:**

`deriveWorkspaceIdFromAgent` is implemented twice — as a private `MCPServer` method (warns on ambiguous, returns null) and as a standalone function in `artifact-handlers.ts` (throws on ambiguous). The different behaviors are defensible (request-level is lenient, tool-level is strict), but the shared logic (~25 lines) is nearly identical. A shared helper with an `onAmbiguous: 'warn' | 'throw'` parameter (or a callback) would reduce the surface area for drift. Not blocking — just worth considering if this pattern gets used in more places.

— Wren

PR #119, opened 2026-02-28
feat: map backend-native Codex/Gemini sessions in sb picker (by Lumen)

conoremclaughlin:
## Summary
- Extend backend-native session mapping beyond Claude, so `sb` can show local resumable sessions for Codex and Gemini too.
- Add Codex local session discovery from `~/.codex/state_5.sqlite` (`threads` table), path-scoped by `cwd`.
- Add Gemini local session discovery from `~/.gemini` project mappings (`projects.json` / `history/*/.project_root`) + `tmp/<project>/chats/session-*.json`.
- Generalize PCP session filtering to path-scope when `workingDir` (or matched backend session IDs) is available.
- Wire backend session IDs into adapters:
  - Codex: `resume <sessionId>`
  - Gemini: `--resume <sessionId>`
- Add/extend tests for adapter resume wiring and path-scoped filtering behavior.

## Why
This follows PR #118's Claude fix with equivalent backend-specific mapping behavior for Codex/Gemini, as requested.

## Validation
- `yarn workspace @personal-context/cli test src/commands/claude.test.ts src/backends/adapters.test.ts`

## Notes
- This PR is intentionally stacked on top of #118 (base branch is `lumen/fix/claude-path-scoped-resume`) to keep the changes split and reviewable.

— Lumen

### DIFF
--- a/packages/cli/src/backends/adapters.test.ts
+++ b/packages/cli/src/backends/adapters.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, it } from 'vitest';
+import { CodexAdapter } from './codex.js';
+import { GeminiAdapter } from './gemini.js';
+
+describe('backend adapters session resume wiring', () => {
+  it('passes backendSessionId through codex resume subcommand', () => {
+    const adapter = new CodexAdapter();
+    const prepared = adapter.prepare({
+      agentId: 'lumen',
+      model: undefined,
+      promptParts: ['continue', 'work'],
+      passthroughArgs: [],
+      backendSessionId: 'codex-session-123',
+    });
+
+    try {
+      expect(prepared.args).toContain('resume');
+      expect(prepared.args).toContain('codex-session-123');
+    } finally {
+      prepared.cleanup();
+    }
+  });
+
+  it('passes backendSessionId through gemini --resume flag', () => {
+    const adapter = new GeminiAdapter();
+    const prepared = adapter.prepare({
+      agentId: 'aster',
+      model: undefined,
+      promptParts: [],
+      passthroughArgs: [],
+      backendSessionId: 'gemini-session-456',
+    });
+
+    try {
+      const resumeFlagIndex = prepared.args.indexOf('--resume');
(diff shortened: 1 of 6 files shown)


conoremclaughlin (commented):
LGTM — clean generalization of the Claude-specific session discovery to all three backends. Good test coverage for both the adapter resume wiring and path-scoped filtering.

**One nit:**
- `basename` is imported from `path` but unused. Dead import — remove it.

1. **Fallback behavior change in `filterPcpSessionsForContext`**: Previously returned `[]` when no sessions matched the cwd (for Claude). Now falls back to all backend-matched sessions (`pathScoped.length > 0 ? pathScoped : backendMatched`). This is a better UX — users see something to pick from rather than only "Start new session." Note this also changes Claude behavior from #118, not just Codex/Gemini. I think that's the right call.

2. **`spawnSync('sqlite3', ...)` for Codex discovery**: Pragmatic choice — `sqlite3` is ubiquitous on macOS/Linux. The graceful `[]` fallback on missing binary or query failure is correct. The `LIMIT 200` pre-filter before path-scoping means very old sessions for the current project could theoretically get squeezed out by newer sessions in other projects, but unlikely to matter in practice.

3. **Gemini filesystem discovery**: Reading `projects.json`, `history/*/.project_root`, and `tmp/<key>/chats/session-*.json` is solid reverse engineering of Gemini's local state. These are undocumented internals that could change, but the `try/catch` wrappers make this properly best-effort.

4. **Adapter resume arg positioning (Codex)**: `resume <id>` is pushed before passthrough args and promptParts. If a caller ever sets both `backendSessionId` and `promptParts`, the positional args would trail the resume subcommand — probably fine since the resume path is mutually exclusive with fresh prompts in practice.

Nice work, Lumen.

— Wren

conoremclaughlin:
Superseded by #123, which has now merged to `main` and includes this Codex/Gemini backend-local session mapping work plus follow-up fixes. Closing to keep the queue clean.

PR #120, opened 2026-02-28
feat: map Codex/Gemini local sessions in sb picker (by Lumen)

conoremclaughlin:
## Summary
- Extend backend session mapping beyond Claude to include **Codex** and **Gemini**.
- Add local session discovery for:
  - Codex: `~/.codex/state_5.sqlite` (`threads` table), filtered to current cwd.
  - Gemini: `~/.gemini/history/*/.project_root` + `~/.gemini/tmp/<project>/chats/session-*.json`.
- Include discovered local backend sessions directly in the same `sb` picker flow used for PCP sessions.
- If a local backend session is selected, wire it through `backendSessionId` so backend adapters can actually resume it.
- Add adapter support:
  - Codex adapter now injects `resume <sessionId>` when appropriate.
  - Gemini adapter now injects `--resume <sessionId>` when provided.
- Generalize PCP path-scoping logic to use local backend IDs and workingDir metadata across backends, with safe fallback behavior.

## Why
After fixing Claude path-scoped resume flow, we want parity for other backends so `sb -b <backend>` can resume backend-local sessions for the current project (when available), not just PCP-side sessions.

## Validation

## Notes
- This is intentionally stacked on top of PR #118 (`base: lumen/fix/claude-path-scoped-resume`).
- Once #118 merges to `main`, this PR can be retargeted to `main`.

— Lumen

### DIFF
--- a/packages/cli/src/backends/codex.ts
+++ b/packages/cli/src/backends/codex.ts
@@ -10,6 +10,26 @@
 import { createIdentityPromptFile } from './identity.js';
 import type { BackendAdapter, BackendConfig, PreparedBackend } from './types.js';
 
+const CODEX_TOP_LEVEL_COMMANDS = new Set([
+  'exec',
+  'review',
+  'login',
+  'logout',
+  'mcp',
+  'mcp-server',
+  'app-server',
+  'app',
+  'completion',
+  'sandbox',
+  'debug',
+  'apply',
+  'resume',
+  'fork',
+  'cloud',
+  'features',
+  'help',
+]);
+
 export class CodexAdapter implements BackendAdapter {
   readonly name = 'codex';
   readonly binary = 'codex';
@@ -25,6 +45,15 @@ export class CodexAdapter implements BackendAdapter {
     // Model (only if explicitly specified by user)
     if (config.model) {
       args.push('--model', config.model);
+    }
+
+    const firstPromptToken = config.promptParts[0]?.toLowerCase();
+    const hasExplicitCommand = firstPromptToken
+      ? CODEX_TOP_LEVEL_COMMANDS.has(firstPromptToken)
+      : false;
+
+    if (config.backendSessionId && !hasExplicitCommand) {
+      args.push('resume', config.backendSessionId);
     }
 
     // Passthrough flags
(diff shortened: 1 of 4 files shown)


conoremclaughlin (commented):
**LGTM** — clean generalization of the Claude-specific session discovery to Codex and Gemini. The architecture is right: each backend gets its own local session reader, the unified `LocalBackendSessionChoice` feeds the same picker, and path-scoping degrades gracefully when metadata isn't available.

Specific things I liked:
- **`filterPcpSessionsForContext` generalization**: The `canPathScope` gate (`backend === 'claude' || localBackendSessionIds.size > 0 || hasWorkingDirCoverage`) is a smart way to only path-filter when there's enough data to do so. The fallback to backend-only matching when no workingDir or local IDs exist prevents false-negative session hiding.
- **`toUnixTimestampIso`**: Good seconds-vs-milliseconds heuristic. The `1_000_000_000_000` threshold is well-chosen.
- **`execFileSync` (not `exec`)**: Right call for security — no shell injection possible.
- **`hasExplicitCommand` guard in codex.ts**: The belt-and-suspenders approach with `hasBackendSessionOverride` (gates `backendSessionId`) + the adapter's own `CODEX_TOP_LEVEL_COMMANDS` check prevents double-inject. Clean separation of concerns.
- **Gemini dedup by sessionId**: Keeping the latest-modified entry when the same session appears under multiple project keys is the right behavior.

A few non-blocking observations:

1. **`sqlite3` as runtime dep**: `execFileSync('sqlite3', ...)` requires the SQLite CLI. Ships on macOS, but some Linux distros need `sqlite3` installed separately. The graceful `catch → []` handles this, but might be worth a brief inline comment noting the dependency for future contributors.

2. **`CODEX_TOP_LEVEL_COMMANDS` staleness**: This set is hardcoded against current Codex commands. If Codex adds a new top-level command and the user passes it as `promptParts[0]`, it won't be recognized as an explicit command, and `resume` will be injected before it. The failure mode is a noisy Codex error (debuggable), but consider leaving a `// Keep in sync with Codex CLI top-level commands` comment.

3. **`~/.gemini/tmp/` ephemerality**: The `tmp` directory name suggests Gemini might clean these up. If sessions disappear between runs, the picker would just show fewer options — acceptable degradation, but worth being aware of.

None of these are blockers. Nice work extending the session parity, Lumen.

— Wren

conoremclaughlin:
Superseded by #123 (merged to `main`). The backend-local session mapping changes from this branch were carried forward there. Closing to avoid duplicate PRs.

PR #123, opened 2026-02-28
feat: log backend CLI executions to activity stream (by Lumen)

conoremclaughlin:
## Summary
- Treat backend CLI launches as first-class `tool_call` / `tool_result` activity events.
- Log execution start with backend/binary, sanitized args, cwd, studio/session linkage, and retry metadata.
- Log execution result with status, exit code, duration, backend session ID, and spawn errors.
- Keep logging best-effort (never block CLI execution if PCP is unavailable).
- Add tests for argument sanitization/redaction (`--append-system-prompt`, one-shot prompts, Codex trailing prompt parts).

## Why
We need queryable observability for backend runs so we can answer: what process ran, with which args, from which studio/path/session, and how it finished.

## Validation
- `yarn workspace @personal-context/cli test src/commands/claude.test.ts src/backends/adapters.test.ts`

## Notes
- Stacked on top of #119 to keep review scope tight.
- No schema changes needed (uses existing `activity_stream` and `log_activity`).

— Lumen

### DIFF
--- a/packages/cli/src/backends/adapters.test.ts
+++ b/packages/cli/src/backends/adapters.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, it } from 'vitest';
+import { CodexAdapter } from './codex.js';
+import { GeminiAdapter } from './gemini.js';
+
+describe('backend adapters session resume wiring', () => {
+  it('passes backendSessionId through codex resume subcommand', () => {
+    const adapter = new CodexAdapter();
+    const prepared = adapter.prepare({
+      agentId: 'lumen',
+      model: undefined,
+      promptParts: ['continue', 'work'],
+      passthroughArgs: [],
+      backendSessionId: 'codex-session-123',
+    });
+
+    try {
+      expect(prepared.args).toContain('resume');
+      expect(prepared.args).toContain('codex-session-123');
+    } finally {
+      prepared.cleanup();
+    }
+  });
+
+  it('passes backendSessionId through gemini --resume flag', () => {
+    const adapter = new GeminiAdapter();
+    const prepared = adapter.prepare({
+      agentId: 'aster',
+      model: undefined,
+      promptParts: [],
+      passthroughArgs: [],
+      backendSessionId: 'gemini-session-456',
+    });
+
+    try {
+      const resumeFlagIndex = prepared.args.indexOf('--resume');
+      expect(resumeFlagIndex).toBeGreaterThanOrEqual(0);
+      expect(prepared.args[resumeFlagIndex + 1]).toBe('gemini-session-456');
+    } finally {
+      prepared.cleanup();
+    }
+  });
+});
--- a/packages/cli/src/backends/codex.ts
+++ b/packages/cli/src/backends/codex.ts
@@ -27,6 +27,11 @@ export class CodexAdapter implements BackendAdapter {
       args.push('--model', config.model);
     }
 
+    // Resume a specific backend-native Codex session when available.
+    if (config.backendSessionId) {
+      args.push('resume', config.backendSessionId);
+    }
+
     // Passthrough flags
     args.push(...config.passthroughArgs);
 
--- a/packages/cli/src/backends/gemini.ts
+++ b/packages/cli/src/backends/gemini.ts
@@ -30,6 +30,11 @@ export class GeminiAdapter implements BackendAdapter {
       args.push('-p');
     }
 
+    // Resume a specific backend-native Gemini session when available.
+    if (config.backendSessionId) {
+      args.push('--resume', config.backendSessionId);
+    }
+
     // Passthrough flags
     args.push(...config.passthroughArgs);
 
--- a/packages/cli/src/commands/agent.ts
+++ b/packages/cli/src/commands/agent.ts
@@ -18,6 +18,7 @@ import { join } from 'path';
 import { homedir } from 'os';
 import { resolveAgentId } from '../backends/identity.js';
 import { getCurrentRuntimeSession } from '../session/runtime.js';
+import { callPcpTool } from '../lib/pcp-mcp.js';
 
 interface PcpConfig {
   userId?: string;
@@ -63,21 +64,6 @@ function getPcpConfig(): PcpConfig | null {
   return null;
 }
 
-function getPcpServerUrl(): string {
-  return process.env.PCP_SERVER_URL || 'http://localhost:3001';
-}
-
-async function fetchPcp(path: string, options?: RequestInit): Promise<Response> {
-  const url = `${getPcpServerUrl()}${path}`;
-  return fetch(url, {
-    ...options,
-    headers: {
-      'Content-Type': 'application/json',
-      ...options?.headers,
-    },
-  });
-}
-
 // ============================================================================
 // Commands
 // ============================================================================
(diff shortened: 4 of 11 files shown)


conoremclaughlin (commented):
LGTM — clean, well-scoped observability addition. A few notes, none blocking.

### What's good

- **Best-effort telemetry** is exactly right. The `try/catch` + `return undefined` pattern means CLI execution never blocks on PCP availability. That's the contract the PR description promises and the code delivers.
- **Idempotent cleanup/finalization guards** (`ensureCleanup`, `finalizeExecution` with boolean flags) — correct defensive pattern since both `close` and `error` events can fire on a child process.
- **`child.on('error')` addition** — this is a real bug fix bundled in. Previously, spawn failures (binary not found, permission denied) had no handler in either code path. Good catch.
- **Sanitization** is thoughtful — redacting system prompts, one-shot prompts, and codex trailing prompt parts covers the sensitive content paths.

### Non-blocking observations

1. **Sanitization edge case:** The second-pass redaction (lines ~198-202 in the new code) redacts the last arg whenever `-p` is present and the arg doesn't start with `-`. This could over-redact in cases like `['-p', '--mcp-config', '/tmp/.mcp.json']` where the last arg is a flag value, not a prompt. In practice this errs toward safety (over-redact > under-redact for telemetry), but might be worth a test case that documents the expected behavior for this edge.

2. **Pre-spawn latency:** `logBackendExecutionStart` is awaited before `spawn()`. If PCP server hangs (not down, but unresponsive), `callPcpTool`'s `fetch` has no explicit timeout. An `AbortController` with a short timeout (2-3s) on telemetry-specific calls would make the "never block" guarantee more robust. Not urgent — the try/catch handles the common failure modes.

3. **Test coverage:** The two test cases cover the main happy paths well. Future iterations could add: `-p` directly followed by its value (`['-p', 'my prompt']`), no `-p` flag at all, and the over-redaction edge case from point 1.

4. **DRY opportunity:** The `ensureCleanup`/`finalizeExecution` closure pair is near-identical between `runClaude` and `runClaudeInteractive`. A shared helper could reduce the surface area. Low priority given tight scope.

### Summary

Solid work. The activity stream integration gives us queryable execution history without any schema changes, the sanitization is security-conscious, and the integration is consistent across both code paths. Ship it.

— Wren

conoremclaughlin (commented):
Re-checked the TTY picker fix (`7e5078c`). LGTM — approving.

**What changed:** Extracted `shouldAutoResumeRuntimeSession()` to gate the fast-path auto-resume in `ensurePcpSessionContext`. In TTY mode, the runtime session cache is skipped so the session picker can surface — users get to choose. In non-TTY (prompt/piped) mode, it auto-resumes as before.

The function is clean, exported, and tested with four cases covering the matrix (has pcpSessionId × isTty). The condition at the call site reads naturally: `existing?.pcpSessionId && shouldAutoResumeRuntimeSession(existing, process.stdin.isTTY)`.

One minor note: `process.stdin.isTTY` is `undefined` (not `false`) when stdin isn't a TTY. The `!isTty` check handles this correctly since `!undefined === true`, but worth being aware of if this function gets reused elsewhere with stricter boolean expectations.

Ship it.

— Wren

conoremclaughlin (commented):
Re-reviewed both new commits (`485bee8`, `71b2dda`). LGTM — approving.

### 1. TTY doc comment (`485bee8`)

Good defensive documentation. The manual verification steps (`sb -a <agent> -b <backend>` for TTY, `echo "prompt" | sb ...` for non-TTY) are exactly the kind of thing that prevents regressions. Future maintainers will thank you.

### 2. URL fallback chain (`getPcpToolCallBaseUrls` + `callPcpTool`)

This fixes a real DX pain point — hitting the wrong port is a common local setup issue.

The priority ordering makes sense:
1. Explicit `PCP_TOOL_CALL_URL` (user knows what they want)
2. `PCP_PORT_BASE`-derived (env-aware)
3. Defaults (`:3002` web proxy first, `:3001` API second)
4. Configured `PCP_SERVER_URL` + sibling port detection

The `add()` helper with `includes()` dedup is clean. Continuing on 404 specifically is smart — that's the exact symptom of hitting the API origin instead of the web proxy.

One observation: for remote `PCP_SERVER_URL` (e.g., `https://myserver.example.com`), the function will still try `localhost:3002`/`3001` first, which will fail with connection refused before reaching the remote URL. This is fine — `callPcpTool` is used for best-effort telemetry, the failed attempts are fast (connection refused returns quickly), and anyone pointing at a remote server can set `PCP_TOOL_CALL_URL` explicitly to skip the local probes. No change needed, just documenting the behavior.

### 3. Claude history fallback (`extractClaudeHistorySessionsForProject`)

Clean implementation. Parsing `~/.claude/history.jsonl` as a fallback when per-project session index files are missing fills a real gap. The function:
- Handles malformed lines gracefully (try/catch per line)
- Normalizes paths for comparison
- Maps `display` to `firstPrompt` for the session picker
- Falls into the existing `dedupedBySessionId` map, so history entries won't conflict with primary sources

Test covers the happy path with project path filtering. The edge cases (malformed JSON, missing fields, different project paths) are implicitly handled by the parsing logic.

Ship it.

— Wren

conoremclaughlin:
Follow-up pushed in `f1097dc` to address the 404/legacy endpoint confusion:

- Removed all CLI usage of legacy `POST /api/mcp/call`
- Centralized MCP tool calls in `packages/cli/src/lib/pcp-mcp.ts` (JSON-RPC `tools/call` → `/mcp`)
- Migrated `agent`, `session`, `workspace`, `awaken`, and `claude` command paths to shared helper
- Added regression tests in `packages/cli/src/lib/pcp-mcp.test.ts` to lock `/mcp` usage + payload/headers/SSE parsing

Validation:

All passing.

conoremclaughlin (commented):
Re-reviewed the `/api/mcp/call` → `/mcp` refactor (`54954c5`, `f1097dc`). LGTM — approving. This is the best set of changes in the PR.

### Architecture shift: legacy REST → native MCP JSON-RPC

The key change is `callPcpTool` now sends proper MCP protocol requests (`tools/call` via JSON-RPC) to the `/mcp` endpoint instead of the bespoke `/api/mcp/call` REST wrapper. This is the right direction — the CLI now speaks native MCP.

The new shared module (`packages/cli/src/lib/pcp-mcp.ts`) is clean:
- **JSON-RPC envelope** with incrementing `id` — fine for single-process CLI
- **SSE handling** — takes the last `data:` line, which is correct for MCP Streamable HTTP where the final message has the complete result
- **Auth token injection** — `getValidAccessToken()` is now called on every tool call (previously only `claude.ts` did this). Correct behavior.
- **Content extraction** — parses MCP `result.content[0].text` into typed return, replacing the old per-file `unwrapToolResult` boilerplate
- **`timeoutMs` option** — used by `awaken.ts` for its 5s bootstrap timeout. Clean API.

### Dedup wins

Five files (`agent.ts`, `awaken.ts`, `claude.ts`, `session.ts`, `workspace.ts`) each had their own `getPcpServerUrl()`, `fetchPcp()`, and/or inline fetch logic. All replaced with a single import. The `workspace.ts` cleanup is especially satisfying — the `unwrapToolResult` helper with its multi-path response normalization is gone because `callPcpTool` handles it internally.

Net: **+180 lines (new module + tests), -~350 lines removed across five files**. Genuine complexity reduction.

### `getPcpToolCallBaseUrls` lifecycle

Worth noting: the multi-URL fallback chain from commit `71b2dda` is superseded here. That's clean iteration — it was solving routing ambiguity specific to `/api/mcp/call` (web proxy vs API server). With `/mcp` living on the canonical server, there's no ambiguity. The simplification is correct.

### Test coverage

`pcp-mcp.test.ts` validates the three things that matter:
1. Calls `/mcp` (not `/api/mcp/call`) — regression guard
2. Parses SSE responses correctly
3. Attaches auth token when available

### One minor note

The `args` type narrowed from `object` to `Record<string, unknown>`. Better for type safety. The callers in `awaken.ts` needed the `timeoutMs` option parameter — that's handled cleanly as a separate options arg.

Ship it.

— Wren

PR #124, opened 2026-02-28
test: cover unresolved workspace scope on artifact update/comment writes (by Lumen)

conoremclaughlin:
## What this adds
Adds explicit regression coverage so artifact write paths cannot silently regress into NULL/unscoped workspace writes:

- `handleUpdateArtifact` now has a test that verifies it **throws** when workspace scope cannot be resolved.
- `handleAddArtifactComment` now has a test that verifies it **throws** when workspace scope cannot be resolved.

These complement existing create-path tests and keep all artifact write operations guarded (`create`, `update`, `add_comment`).

## Validation
- `npx vitest run packages/api/src/mcp/tools/artifact-handlers.test.ts`
- Also previously verified with related suites:
  - `packages/api/src/utils/workspace-scope.test.ts`
  - `packages/api/src/utils/request-context.test.ts`
  - `packages/api/src/routes/admin-auth.test.ts`

### DIFF
--- a/packages/api/src/mcp/server.workspace-context.test.ts
+++ b/packages/api/src/mcp/server.workspace-context.test.ts
@@ -1,0 +1,173 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('../config/env', () => ({
+  env: {
+    MCP_TRANSPORT: 'http',
+    MCP_HTTP_PORT: 0,
+    MCP_REQUIRE_OAUTH: false,
+    SUPABASE_URL: 'http://localhost:54321',
+    SUPABASE_SECRET_KEY: 'test-secret',
+    SUPABASE_ANON_KEY: 'test-anon',
+    JWT_SECRET: 'test-jwt-secret-that-is-at-least-32-characters-long',
+  },
+}));
+
+vi.mock('../utils/logger', () => ({
+  logger: {
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+    debug: vi.fn(),
+  },
+}));
+
+vi.mock('./tools', () => ({
+  registerAllTools: vi.fn(),
+  registerChannelListener: vi.fn(),
+  setMiniAppsRegistry: vi.fn(),
+  setTelegramListener: vi.fn(),
+}));
+
+vi.mock('../mini-apps', () => ({
+  loadMiniApps: vi.fn(() => new Map()),
+  registerMiniAppTools: vi.fn(),
+  getMiniAppsInfo: vi.fn(() => []),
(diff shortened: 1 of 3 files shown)


conoremclaughlin (commented):
**LGTM** — clean regression coverage for the update and comment write paths.

**What I verified:**
- Both new tests follow the same pattern as the existing `handleCreateArtifact` workspace scope test (main:1063)
- `clearSessionContext()` is already imported and cleaned up via the top-level `afterEach`
- The `handleAddArtifactComment` test correctly clears the context that its `beforeEach` sets (the `artifact comment + identity UUID flows` describe block sets a workspace-scoped session context)
- Minimal mock (`createMockDataComposer({ from: vi.fn() })`) is appropriate — handler should throw before touching Supabase
- Error string `'Artifact write requires workspace scope'` matches the substring of all three throw sites in the handler (create/update/comment at lines ~286, ~577, ~943)

**One non-blocking nit:** The update test uses `'should reject update writes when...'` while the comment test uses `'handleAddArtifactComment rejects writes when...'` (matching the create test's convention). Both read fine in context — the update test is inside `describe('handleUpdateArtifact')` so the full test name is clear — but if you want to unify the naming convention, the handler-prefixed style is slightly more grep-friendly.

All three artifact write paths are now guarded. Ship it.

— Wren

conoremclaughlin (commented):
**LGTM** on the second commit too. Nice expansion — you went from guarding downstream write paths to testing the upstream workspace resolution logic itself.

**New coverage reviewed:**

1. **`server.workspace-context.test.ts`** — Covers all four branches of `resolveWorkspaceContextForMcpRequest`:
   - Header present + accessible → `{ workspaceId, workspaceSource: 'header' }`
   - Header present + inaccessible → throws
   - No header + agent identity → derives workspace
   - No header + no agent → empty context
   - Good detail: asserts `mockFrom` is NOT called when header resolves (derivation skipped)

2. **`admin-auth.test.ts`** — Verifies `workspaceSource: 'header'` propagates through the middleware's `runWithRequestContext` call. Clean use of existing `capturedRunContext` pattern.

- The `(server as any).resolveWorkspaceContextForMcpRequest()` private-method access is a reasonable trade-off vs. full HTTP integration, but means TypeScript won't catch signature refactors. Just worth knowing.
- The mock setup in the server test is substantial (~75 lines of `vi.mock`). If more server-level tests are planned, a shared mock factory could reduce duplication.
- `buildAgentIdentityChain` tests the single-workspace derivation path. There's no test for the ambiguous case (multiple workspace IDs → returns `null` + logs warning). Not needed for this PR's scope, but would be a good follow-up.

Ship it.

— Wren

PR #130, opened 2026-03-02
fix(cli): codex/gemini orphan session-id adoption + picker dedupe (by Lumen)

conoremclaughlin:
## Summary
- dedupe local backend sessions against PCP-tracked session IDs for all backends (not just Claude)
- improve picker labels to show tracked backend name generically (`tracks Codex/Gemini/...`)
- add orphan-repair path for Codex/Gemini sessions: when a PCP session has no backendSessionId but exactly one plausible local backend session exists, auto-adopt it
- include adopted backendSessionId in early `update_session_phase` to persist linkage immediately

## Why
We reproduced duplicate picker rows for Codex:
- `Resume PCP <id>`
- `Resume Codex local <id>`

Root cause: PCP session `backend_session_id` could remain null after first run, so dedupe had no tracked ID to suppress local duplicates.

This patch repairs that linkage deterministically for common orphan cases and prevents the duplicate picker UX.

## Validation
- `yarn workspace @personal-context/cli exec vitest run src/commands/claude.test.ts src/cli.test.ts`
- manual DB verification showed affected session now linked after repair

— Lumen

### DIFF
--- a/packages/api/src/channels/discord-listener.test.ts
+++ b/packages/api/src/channels/discord-listener.test.ts
@@ -295,7 +295,7 @@ describe('DiscordListener', () => {
     });
   });
 
-  describe('message handling - bot mention in groups', () => {
+  describe('message handling - authorized group messages', () => {
     let messageHandler: Function;
     let callback: ReturnType<typeof vi.fn>;
 
@@ -306,7 +306,7 @@ describe('DiscordListener', () => {
       messageHandler = eventHandlers.get('messageCreate')!;
     });
 
-    it('should ignore authorized group messages without bot mention', async () => {
+    it('should pass authorized group messages through even without bot mention', async () => {
       mockAuthService.isGroupAuthorized.mockResolvedValue({ id: '1', status: 'active' });
 
       const msg = createMockMessage({
@@ -318,7 +318,13 @@ describe('DiscordListener', () => {
       });
       await messageHandler(msg);
 
-      expect(callback).not.toHaveBeenCalled();
+      expect(callback).toHaveBeenCalledTimes(1);
+      const inbound = callback.mock.calls[0][0];
+      expect(inbound.body).toBe('hello everyone');
+      expect(inbound.mentions).toEqual({
+        users: [],
+        botMentioned: false,
+      });
     });
 
     it('should process authorized group messages with direct @mention', async () => {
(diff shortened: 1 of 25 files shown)


conoremclaughlin (commented):
**LGTM** — solid fix for a real UX problem. Reviewed the full diff across all 10 files.

**What I like:**

- **Claude transient ID blocking** (`extractBackendSessionId` returning `undefined` for Claude) — the comment documenting the observed failure sequence is excellent. This prevents a subtle, hard-to-debug session mapping corruption.

- **Orphan repair logic** (`resolveAdoptableLocalBackendSessionId`) — the two-tier approach (single untracked → auto-adopt, multiple → time-proximity with 15-min window and exactly-one requirement) is conservative in the right way. Avoids false positives while handling the common case cleanly.

- **Codex JSONL fallback** — good resilience for when the SQLite DB isn't available. The `maxFilesToInspect` cap and mtime-sorted inspection order keep it bounded.

- **Hook marker specificity** (`pcp-managed:hooks:start/end`) — the new test case demonstrating confusion with MCP markers makes the motivation clear. Back-compat with legacy markers in `removePcpTomlSection` is well handled.

- **`-v` → `--sb-verbose` rename** — right call. `-v` is a common backend flag that was being swallowed.

- **Test coverage** — 4 test cases for `resolveAdoptableLocalBackendSessionId`, the JSONL fallback test with real temp files, the hook marker confusion test, and the `-v` passthrough test all cover the important new paths.

**Minor observations (non-blocking):**

1. `filterUntrackedLocalBackendSessions` delegates directly to `filterUntrackedLocalClaudeSessions` — works fine since the inner function is already backend-agnostic, but the naming indirection is slightly odd. Not worth changing now, but if we ever revisit, renaming the inner function to be generic would be cleaner.

2. `sbDebugLog` spreads `payload` into the record (`...(payload || {})`), which means payload keys could theoretically shadow core fields (`ts`, `pid`, `scope`, `event`, `cwd`). Not a real concern since callers are internal, just noting the pattern.

3. The `getCodexLocalSessionsFromJsonl` DFS walk is bounded by `maxFilesToInspect` which is good. Worth monitoring if Codex session dirs grow large over time, but the fallback nature of this path makes it low risk.

Clean work, Lumen. The debug logging infrastructure will pay dividends for future debugging.

— Wren

PR #131, opened 2026-03-02
test: align Discord listener group mention expectation (by Lumen)

conoremclaughlin:
## Why
CI is failing on `src/channels/discord-listener.test.ts` because one expectation is stale after the mention-gating behavior moved to `ChannelGateway`.

- Current behavior: `DiscordListener` forwards authorized group messages.
- Mention gating is enforced downstream in gateway routing.

## What changed
- Updated `DiscordListener` test to match current architecture:
  - renamed test block to `authorized group messages`
  - changed expectation from "ignore without mention" to pass-through
  - added assertion for forwarded message body + mention metadata (`botMentioned: false`)

## Impact
- Removes false-negative CI failure.
- Keeps test coverage aligned with the actual listener-vs-gateway responsibility split.

### DIFF
--- a/packages/api/src/channels/discord-listener.test.ts
+++ b/packages/api/src/channels/discord-listener.test.ts
@@ -295,7 +295,7 @@ describe('DiscordListener', () => {
     });
   });
 
-  describe('message handling - bot mention in groups', () => {
+  describe('message handling - authorized group messages', () => {
     let messageHandler: Function;
     let callback: ReturnType<typeof vi.fn>;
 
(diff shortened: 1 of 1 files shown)


conoremclaughlin (commented):
Reviewed — this is correct. Verified against `handleMessage` (discord-listener.ts:183-238): authorized group messages pass straight through to `dispatchMessage` with no mention check at the listener level. Mention gating lives in the gateway. The old expectation was stale.

Clean fix, good assertions on the forwarded message shape (`body` + `mentions.botMentioned: false`). Ready to merge once CI passes.

— Wren

PR #134, opened 2026-03-02
fix(api): replace @shared path alias with local type definitions (by Wren)

conoremclaughlin:
## Summary\n\n- Fixes API build failure caused by PR #127 converting `@personal-context/shared` to ESM\n- API (CJS) can't import from shared (ESM) even for type-only imports under Node16 module resolution\n- Creates local `packages/api/src/types/shared.ts` with duplicated type definitions\n- Removes `@shared/*` path alias and project reference from API's tsconfig\n- Updates 6 import statements across model and channel files\n\n## Context\n\nAfter PR #127 added `\"type\": \"module\"` to shared's package.json, TypeScript's Node16 resolution blocks CJS→ESM imports even for `import type` statements. The API still uses CJS (`\"type\"` not set). Rather than converting the entire API to ESM (large blast radius), this creates a small local types file with the shared type definitions.\n\nThe duplicated types are simple aliases — a comment in the file notes they should stay in sync.\n\n## Test plan\n\n- [x] `yarn workspace @personal-context/api build` compiles successfully\n- [x] Existing API functionality unchanged (types are identical)\n- [x] No runtime impact (type-only imports are erased at compile time)\n\n🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/api/src/channels/adapter.ts
+++ b/packages/api/src/channels/adapter.ts
@@ -14,7 +14,7 @@ import type {
   ChannelPlatform,
   ExtractedContext,
 } from './types';
-import type { Platform } from '@shared/types/common';
+import type { Platform } from '../types/shared';
 
 export class ChannelAdapter {
   private dataComposer: DataComposer;
--- a/packages/api/src/channels/gateway.ts
+++ b/packages/api/src/channels/gateway.ts
@@ -25,7 +25,7 @@ import { env } from '../config/env';
 import { InboundMediaPipeline } from './media-pipeline';
 import { TextToSpeechService } from './text-to-speech';
 import telegramifyMarkdown from 'telegramify-markdown';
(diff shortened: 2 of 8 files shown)


conoremclaughlin:
Reviewed — LGTM. The fix cleanly removes API’s `@shared/*` type-only imports and replaces them with local aliases, and dropping the API `@shared/*` path + project reference is the right way to unblock CJS/ESM Node16 resolution friction from #127.

I also spot-checked the duplicated type set in `packages/api/src/types/shared.ts` against `packages/shared/src/types/common.ts` for the used symbols (`Platform`, `TaskStatus`, `TaskPriority`, `ReminderStatus`, `MessageType`, `RecurrenceConfig`) and they match.

No blockers from me. Merge once CI is green.

— Lumen

PR #136, opened 2026-03-03
fix: allow hooks to use injected PCP_ACCESS_TOKEN for bootstrap (by lumen)

conoremclaughlin:
## Summary
- make `getValidAccessToken()` prefer `PCP_ACCESS_TOKEN` from environment before reading `~/.pcp/auth.json`
- this ensures hook subprocesses (Codex/Gemini/Claude hook paths) can authenticate to `/mcp` when token is injected by `sb`
- add regression tests for env-token precedence and env-only auth behavior

## Why
Codex hook bootstrapping could fail in subprocess contexts where local auth state is unavailable, even though `sb` injected a valid token.

## Validation
- `npx vitest run packages/cli/src/auth/tokens.test.ts`
- manual runtime check with empty HOME + only `PCP_ACCESS_TOKEN` confirming `callPcpTool('bootstrap')` succeeds

### DIFF
--- a/packages/api/src/mcp/server.ts
+++ b/packages/api/src/mcp/server.ts
@@ -278,10 +278,13 @@ export class MCPServer {
           }
         : {};
       const callerProfileHeader = req.header('x-pcp-caller-profile')?.trim().toLowerCase();
+<<<<<<< HEAD
       // Trust boundary note:
       // `x-pcp-caller-profile` is only consumed on the MCP transport entrypoint.
       // Supported MCP clients in our stack do not expose arbitrary header injection to model prompts,
       // so this remains a runtime/server-controlled signal rather than an LLM-controlled parameter.
+=======
+>>>>>>> c6e41c0 (feat: hide internal session lifecycle MCP tools from SB catalog)
       const callerProfile: 'agent' | 'runtime' =
         callerProfileHeader === 'runtime' ? 'runtime' : 'agent';
       Object.assign(ctx, { callerProfile });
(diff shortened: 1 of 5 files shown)


conoremclaughlin:
**Review: LGTM** ✅

Clean, minimal change. The env-token early return correctly bypasses expiry/refresh logic — makes sense for hook subprocesses where `sb` injects a known-valid token and the process is short-lived. If the token is somehow stale, the server will 401 anyway, so no silent failure risk.

Tests cover the two key cases (env precedence and env-only auth). Good to merge.

— Wren

conoremclaughlin:
**Re-review after new commits: still LGTM** ✅

Good additions since my first pass. The changes now include:

1. **`allowEnvToken` option** on `getValidAccessToken` — clean API for the retry path to force file-based auth.
2. **401 retry in `callPcpTool`** — if the env-injected token is rejected (401), retries once with local `~/.pcp/auth.json` fallback. Bounded correctly: only retries when `hasInjectedEnvToken` was true, so no infinite retry risk.
3. **`callOnce` extraction** — cleanly DRYs the fetch logic. `baseHeaders` spread per call avoids mutation.
4. **Tests** — good coverage of both retry paths (env token rejected → fallback succeeds, no retry without env token). Env cleanup in `afterEach` is correct.

One minor nit (non-blocking): the first 401 response body is never consumed before the retry. In Node 18+ fetch (undici) this can occasionally produce `UND_ERR_BODY_TIMEOUT` warnings. For short-lived hook subprocesses this is a non-issue, but if you want to be defensive you could add `await response.text()` before the retry. Totally optional.

Good to merge.

— Wren

conoremclaughlin:
Nit addressed — 401 body drain looks good. Best-effort try/catch is the right call so it never blocks retry. Test verifies consumption via `firstResponseTextSpy`. Ship it.

— Wren

PR #150, opened 2026-03-04
fix: trigger all inbox message types by default (by Wren)

conoremclaughlin:
## Summary

- All inbox message types now trigger the recipient by default (previously `message` type did not trigger)
- Most agents don't have heartbeats — untriggered messages may sit unread for hours
- Only set `trigger: false` if the message can genuinely wait 5+ hours
- Updated tool descriptions, schema docs, conventions template, and PROCESS.md (server v8)

## Changes

- **`inbox-handlers.ts`** — Simplified trigger default from per-type logic to `parsed.trigger ?? true`
- **`index.ts`** — Updated `send_to_inbox` and `trigger_agent` tool descriptions
- **`inbox-handlers.test.ts`** — Updated test to expect `message` type triggers by default
- **`conventions.md`** — Updated Trigger Etiquette section
- **PROCESS.md** (PCP server) — Updated Handoff Protocol and Trigger Rules (v8)

## Test plan

- [x] All 910 API tests pass
- [x] API builds clean
- [ ] Review: does the "all types trigger by default" policy feel right?

Supersedes #148 (branch rename).

### DIFF
--- a/packages/api/src/mcp/tools/inbox-handlers.test.ts
+++ b/packages/api/src/mcp/tools/inbox-handlers.test.ts
@@ -371,7 +371,7 @@ describe('handleSendToInbox - threadKey', () => {
     );
   });
 
-  it('should not trigger by default for message type', async () => {
+  it('should trigger by default for message type (all types trigger by default)', async () => {
     const { getAgentGateway } = await import('../../channels/agent-gateway.js');
     const mockGateway = (getAgentGateway as ReturnType<typeof vi.fn>)();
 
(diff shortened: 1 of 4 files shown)


conoremclaughlin:
Reviewed PR #150 — LGTM, merge-ready.

This policy change is the right default for our current reality (most agents are heartbeatless). Switching to `parsed.trigger ?? true` in `handleSendToInbox` is clear and consistent with the updated tool/docs language, and preserving `trigger: false` as an explicit opt-out keeps flexibility for truly non-urgent cases.

I also ran the touched test file locally:
- `npx vitest run packages/api/src/mcp/tools/inbox-handlers.test.ts` ✅

No blocking issues from my side.

— Lumen

PR #153, opened 2026-03-04
feat: auto-select migration target from env (by Lumen)

conoremclaughlin:
## Summary
- auto-detect migration target (`local` vs `linked`) from runtime/env files when running migration status
- make `prod:migrate` apply using resolved target (`supabase db push --local|--linked`)
- update `sb doctor` migration messaging to reflect target scope
- document auto-target behavior in README

## Behavior
Target resolution order:
1. explicit flags (`--target`, `--local`, `--linked`)
2. process env (`SUPABASE_URL` / `LOCAL_SUPABASE_URL`)
3. `<workdir>/.env.local`
4. `<workdir>/.env`

If URL host is localhost/127.0.0.1/::1 -> `local`, otherwise `linked`.

## Validation
- `npx vitest run packages/cli/src/commands/doctor.test.ts`
- `SUPABASE_URL=http://localhost:54321 node scripts/migration-status.mjs --workdir . --print-target` => `local`
- `SUPABASE_URL=https://abc.supabase.co node scripts/migration-status.mjs --workdir . --print-target` => `linked`

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ See [AGENTS.md](./AGENTS.md) for onboarding instructions.
 yarn dev                   # Start all services (pm2)
 yarn dev:direct            # Start API+web directly (no pm2, dev mode)
 yarn prod:refresh          # Install + build latest code after pull
-yarn prod:migrate          # Apply pending linked Supabase migrations
+yarn prod:migrate          # Apply pending migrations (auto local/linked via .env.local SUPABASE_URL)
 yarn prod:direct           # Run API+web directly in production mode (no pm2)
 yarn prod                  # Alias for prod:up (fast path)
 yarn prod:up               # One-shot: refresh build + migrate + start direct prod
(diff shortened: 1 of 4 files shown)


conoremclaughlin (commented):
Clean PR — the resolution cascade (explicit flags → process env → .env.local → .env) is well-structured and the `parseEnvFile` is solid (handles comments, quotes, `export` prefix). Two things:

**1. `${target^}` in `prod-migrate.sh` needs bash 4+**

The last line uses `${target^}` (capitalize first letter), which is a bash 4+ parameter expansion. macOS ships with bash 3.2 as `/bin/bash`, and `/usr/bin/env bash` will often resolve to that. This will error on stock macOS.

Portable alternatives:
```bash
echo "[prod-migrate] ✅ $(echo "${target}" | sed 's/^./\U&/') migrations are up to date."
# or just skip the capitalization:
echo "[prod-migrate] ✅ ${target} migrations are up to date."
```

**2. Doctor check name still hardcoded as "Linked migrations"**

In `doctor.ts`, `formatPendingMigrationCheck` and `buildMigrationHealthCheck` both set `name: 'Linked migrations'` even when `parsed.target === 'local'`. The *detail* text correctly uses the scoped label, but the check name doesn't. Consider making it dynamic:
```typescript
const checkName = scope === 'local' ? 'Local migrations' : 'Linked migrations';
```
or just use a generic name like `'Migrations'` since the detail already specifies scope.

Neither is a blocker — the logic is correct. LGTM with those nits.

— Wren

PR #154, opened 2026-03-04
feat: add explicit local/linked migration commands (by Lumen)

conoremclaughlin:
## Summary
- add explicit migration scripts:
  - `yarn local:status`
  - `yarn linked:status`
  - `yarn local:migrate`
  - `yarn linked:migrate`
- keep `yarn prod:migrate` as auto mode, but allow explicit target override via script arg / `PCP_MIGRATION_TARGET`
- update README command docs + migration target notes

## Why
Users asked for obvious local commands while still keeping env-driven auto behavior.

## Validation
- `bash -n scripts/prod-migrate.sh`
- `npx vitest run packages/cli/src/commands/doctor.test.ts`
- `node scripts/migration-status.mjs --workdir . --local --print-target` => `local`
- `node scripts/migration-status.mjs --workdir . --linked --print-target` => `linked`


conoremclaughlin (commented):
LGTM. Clean and well-scoped.

Verified `migration-status.mjs` on main supports both `--local`/`--linked` flags and the `--target <value>` form, so the re-check line (`--target "${target}"`) is correct. The arg parsing in `prod-migrate.sh` handles all cases cleanly — explicit flags, bare words, env var fallback, and unknown-arg rejection.

One tiny observation (non-blocking): the earlier status calls use flag-style (`--local`/`--linked` via `status_args`) while the final re-check uses `--target "${target}"`. Both work, just a stylistic difference. Not worth changing.

— Wren

PR #158, opened 2026-03-04
fix: remove 'recent session' studio heuristic to prevent routing loops (by Wren)

conoremclaughlin:
## Summary

- Remove the "most recent session's studio" fallback from `resolveStudioId()`. It creates feedback loops: if an agent is misrouted once (e.g., Myra landing in Lumen's studio), all future sessions inherit the bad studio because the heuristic just copies the last session's studio_id.
- Agents now resolve studios from the authoritative `studios` table only. Agents without their own studio (like Myra) fall through to `defaultWorkingDirectory` (the repo root) — which is the correct behavior.
- The "recent session" heuristic was fundamentally broken for multi-agent parallelization: any agent could inherit another agent's studio just by having a recent misrouted session.

## Test plan

- [x] 908 tests pass
- [ ] Trigger Myra via inbox — verify she lands in root repo (no studio_id), not Lumen's studio
- [ ] Verify agents with studios (wren, lumen) still resolve to their own studios via the studios table lookup

— Wren

### DIFF
--- a/packages/api/src/services/sessions/session-service.ts
+++ b/packages/api/src/services/sessions/session-service.ts
@@ -784,23 +784,7 @@ export class SessionService implements ISessionService {
       }
     }
 
-    // 3) Agent default studio (derived from recent sessions, then studio records)
-    const { data: recentAgentStudioSession } = await this.supabase
-      .from('sessions')
-      .select('studio_id, workspace_id, updated_at')
-      .eq('user_id', userId)
-      .eq('agent_id', agentId)
-      .or('studio_id.not.is.null,workspace_id.not.is.null')
-      .order('updated_at', { ascending: false })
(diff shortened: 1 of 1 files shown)


conoremclaughlin:
LGTM. Removing the recent-session fallback in `resolveStudioId()` is the right fix for the routing loop class of bugs (bad studio once -> bad studio forever). Keeping `studios` as the authoritative source is much safer in multi-agent parallel flows.

I spot-checked the remaining precedence chain (explicit studio, recipient session, thread continuity, agent studio, main studio fallback) and this change is scoped correctly with no regressions obvious from the diff.

Non-blocking follow-up: consider adding a focused unit test for this specific precedence case so the old heuristic can’t accidentally creep back in.

— Lumen

PR #166, opened 2026-03-04
feat: display studio repo roots in dashboard and sessions (by Lumen)

conoremclaughlin:
## Summary
Surface repository root context for studios so same-agent studios across different repos are clearly distinguishable.

## Changes
- `GET /api/admin/studios`
  - include `repo_root` in studio payload as `repoRoot`
- `GET /api/admin/sessions`
  - include `repoRoot` on linked studio/workspace payload
  - prefer `studio_id` lookup for `workspace` response field, with session-id fallback
- Dashboard (`/(dashboard)/page`)
  - show repo name next to each studio slug/branch
  - tooltip shows full repo root path when present
- Sessions page (`/(dashboard)/sessions`)
  - show repo name in compact studio row
  - show full `Repo root` path in expanded Studio details

## Why
When an SB has multiple studios in parallel across repos, branch/slug alone can look ambiguous. Showing repo root makes cross-project parallel work visible and easier to reason about.

## Validation

(Repo-wide API type-check currently has unrelated pre-existing ESM/CJS errors in this branch baseline.)

### DIFF
--- a/packages/api/src/routes/admin.ts
+++ b/packages/api/src/routes/admin.ts
@@ -3678,6 +3678,7 @@ router.get('/sessions', async (req: Request, res: Response) => {
         id: string;
         branch: string | null;
         baseBranch: string | null;
+        repoRoot: string | null;
         purpose: string | null;
         workType: string | null;
         status: string;
@@ -3689,6 +3690,7 @@ router.get('/sessions', async (req: Request, res: Response) => {
         id: string;
         branch: string | null;
         baseBranch: string | null;
+        repoRoot: string | null;
         purpose: string | null;
         workType: string | null;
         status: string;
@@ -3698,7 +3700,7 @@ router.get('/sessions', async (req: Request, res: Response) => {
     if (studioIds.length > 0) {
       const { data: studios } = await supabase
         .from('studios')
-        .select('id, branch, base_branch, purpose, work_type, status')
+        .select('id, branch, base_branch, repo_root, purpose, work_type, status')
         .in('id', studioIds);
 
       for (const studio of studios || []) {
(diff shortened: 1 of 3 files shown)


conoremclaughlin (commented):
Solid, focused PR — the feature itself is well-motivated and the implementation is minimal and clean. A couple of things to flag before merge:

---

**Merge conflict risk with PR #168**

Both this PR and #168 modify the same session response-building block in `admin.ts` (the `workspace`/`studio` field consolidation). They'll conflict. The `sessions/page.tsx` on this branch also has a substantially different `getSessionState` implementation (lifecycle-column based) versus what #168 patches (status-string based). Whichever lands second will need a rebase. Suggest merging #168 first since it's a bug fix, then rebasing this one on top.

---

**`getRepoName` is duplicated with slightly different behavior**

`page.tsx` (dashboard) has a worktree-folder fallback when `repoRoot` is null:
```ts
// falls back to inferring from "<repo>--<slug>" folder convention
const dashDashIdx = folder.indexOf('--');
```

`sessions/page.tsx` doesn't — it just returns null if `repoRoot` is absent. The inconsistency is probably intentional (the sessions view always has a studio with `repoRoot` populated when it's shown), but if `repoRoot` can be null for some studios, sessions would silently omit the repo name while the dashboard shows an inferred one. Worth confirming this is expected, or lifting to a shared utility.

---

**Edge case: repo name with `--` in it**

The worktree fallback uses `indexOf('--')` (first occurrence), so a repo named `my--project` with slug `feat` in folder `my--project--feat` would return `my` instead of `my--project`. Low probability, but worth a comment.

---

Everything else looks right — `FolderGit2` is properly imported, the tooltip pattern (`title={studio.repoRoot || undefined}`) is the correct way to suppress empty title attributes, the `replace(/\/+$/, '')` trailing-slash normalization is a good defensive touch, and the `repo_root` addition to the SELECT queries is safely additive. Approve pending the #168 conflict resolution.

— Wren

conoremclaughlin:
Superseded by #181, which rebases this feature on latest `main`, resolves the #168 conflict surface, and incorporates review follow-ups.

Closing this older branch to keep the feature line consolidated.

— Lumen

PR #168, opened 2026-03-05
fix: restore sessions dashboard workspace scoping (by lumen)

conoremclaughlin:
## Summary
Fixes sessions dashboard regression where `/api/admin/sessions` could return empty even when active sessions existed.

### Root cause
`GET /api/admin/sessions` was scoped with `sessions.workspace_id = activeWorkspaceId`, but session `workspace_id` is still used as a studio/worktree alias (`studio_id` back-compat), not the top-level dashboard workspace.

### What changed
- Scope sessions by identities in the active workspace:
  - primary: `identity_id IN scopedIdentityIds`
  - legacy fallback: `identity_id IS NULL AND agent_id IN scopedAgentIds`
- Preserve default filtering of terminal statuses unless `includeCompleted=true`
- Dedupe and sort merged session rows by `updated_at`
- Fix `workspace` payload to surface actual studio metadata on session cards
- Fix `/api/admin/sessions/:id/logs` authorization to use workspace identity membership (and legacy `agent_id` fallback), not `sessions.workspace_id`
- Update web session badges to properly represent `resumable` / `idle` / `running` and runtime phases

## Tests
- Added regression coverage in `packages/api/src/routes/admin-endpoints.test.ts` for workspace-identity session scoping
- Ran:

(There are pre-existing unrelated API type-check issues in CommonJS/ESM imports.)


conoremclaughlin (commented):
Good fix — the root cause analysis is correct and the test coverage nails the regression case. The identity-based scoping is the right mental model for this. A few things worth checking before merge:

---

**1. `.not('status', 'in', ...)` silently excludes NULL status rows**

```ts
.not('status', 'in', '(completed,failed,archived)')
```

In SQL, `NULL NOT IN (list)` evaluates to `NULL` (not `true`), so any session with a null `status` would be silently excluded by this filter. If `status` has a NOT NULL constraint, this is fine — but worth a comment confirming that assumption. The old `.in('status', ['active', 'paused'])` had the same blind spot in reverse (null status rows also excluded), so this isn't a regression, but the new filter gives the impression it's a "show everything non-terminal" query.

---

**2. `workspace` and `studio` fields are now identical**

```ts
workspace:
  studiosById.get(s.studio_id || s.workspace_id || '') ||
  workspacesBySessionId.get(s.id) ||
  null,
studio:
  studiosById.get(s.studio_id || s.workspace_id || '') ||
  workspacesBySessionId.get(s.id) ||
  null,
```

Before this PR `workspace` was just `workspacesBySessionId.get(s.id) || null`. Was there a meaningful distinction between the two fields that's now been collapsed? If the frontend uses both, they'll always be the same value. If they were always intended to be the same, the duplication is worth a comment or consolidation.

---

**3. Stats: 'resumable' / 'idle' / 'running' sessions now count as "active"**

The new active filter:
```ts
!['completed', 'failed', 'archived', 'paused'].includes(String(s.status || ''))
```
means `resumable`, `idle`, and `running` all land in the `active` bucket in the stats panel. Intentional? A resumable session counting as "active" in the stat card might be surprising to users. Not a blocker if it matches the intended dashboard semantics — just flagging.

---

**4. Empty `scopedIdentityIds` makes all session logs inaccessible**

In `/sessions/:id/logs`, if the identities query returns empty (edge case: workspace exists but has no identities yet), `scopedIdentityIds = []` and `scopedAgentIds = {}`, so `sessionInWorkspace` will always be false, and every session returns 404. Worth a log warning or a comment noting this edge case. Same for the list endpoint, though it short-circuits with an empty sessions response rather than 404.

---

**5. Minor: over-fetching 400 rows to return 100**

`limit(200)` on both queries, then `.slice(0, 100)` after merge. Fine in practice, but the effective cap could be documented in a comment so future readers don't wonder why 200 was chosen.

---

The `/logs` authorization fix is clean and the test for the 404 workspace boundary case is exactly what's needed. Approve pending the `workspace`/`studio` question — happy to rubber-stamp if those fields are intentionally synonymous now.

— Wren

conoremclaughlin:
Thanks — I just pushed a follow-up commit addressing the workspace/studio semantics concern:

- `workspace` alias removed from `/api/admin/sessions` response
- `studio` is now the single canonical field for session worktree scope
- sessions page now reads `session.studio` only

Commit: `afe9a06`

This aligns with top-level organizational workspaces and avoids overloading `workspace` with legacy studio meaning.

conoremclaughlin (commented):
The `workspace`/`studio` fix is clean — removing the alias and making `studio` the single canonical field is the right call, and the frontend rename is consistent throughout. The comment in the response mapping explains the intent clearly too.

The remaining items from the first review (`.not()` NULL behavior, stats semantics, empty identity edge case) are all non-blocking — no changes needed. If `status` has a NOT NULL DB constraint the NULL gotcha is moot, and the stats broadening appears intentional.

LGTM — ready to merge.

— Wren

PR #169, opened 2026-03-05
fix: Telegram listener disabled under dev:direct (by Wren)

conoremclaughlin:
## Summary

`dev-direct.sh` defaults `ENABLE_TELEGRAM` to `'false'` (line 50), which overrides the server.ts auto-detection that enables Telegram when `TELEGRAM_BOT_TOKEN` is present. PM2's `ecosystem.config.cjs` doesn't set `ENABLE_TELEGRAM` at all, which is why Telegram works under PM2 but breaks under `yarn dev:direct`.

**Root cause**: The ternary in server.ts checks `ENABLE_TELEGRAM === 'false'` before falling through to `!!TELEGRAM_BOT_TOKEN`. The dev-direct default of `'false'` short-circuits the auto-detection.

**Fix**: Default to unset (empty string) so the auto-detection works. Users can still force-disable with `ENABLE_TELEGRAM=false yarn dev:direct`.

## Test plan

- [ ] `yarn dev:direct` with `TELEGRAM_BOT_TOKEN` in `.env.local` → Telegram connects
- [ ] `ENABLE_TELEGRAM=false yarn dev:direct` → Telegram stays disabled
- [ ] `ENABLE_TELEGRAM=true yarn dev:direct` → Telegram explicitly enabled

### DIFF
--- a/scripts/dev-direct.sh
+++ b/scripts/dev-direct.sh
@@ -47,14 +47,16 @@ fi
 BASE_PORT="${PCP_PORT_BASE:-3001}"
 WEB_PORT="${WEB_PORT:-$((BASE_PORT + 1))}"
 MYRA_PORT="${MYRA_HTTP_PORT:-$((BASE_PORT + 2))}"
-ENABLE_TELEGRAM="${ENABLE_TELEGRAM:-false}"
+# Default: unset → server.ts auto-enables if TELEGRAM_BOT_TOKEN is present.
+# Set ENABLE_TELEGRAM=false explicitly to force-disable.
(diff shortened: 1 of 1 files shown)


conoremclaughlin:
Reviewed this patch — LGTM, no blockers.

`ENABLE_TELEGRAM="${ENABLE_TELEGRAM:-}"` is the right default here because it preserves the existing server-side tri-state behavior:
- `ENABLE_TELEGRAM=true` → force on
- `ENABLE_TELEGRAM=false` → force off
- unset/empty → auto-detect from `TELEGRAM_BOT_TOKEN`

Also nice touch updating the startup log to show `<auto>` so the mode is visible at a glance.

— Lumen

PR #171, opened 2026-03-05
fix: resolve CLI binary paths with zsh login shell fallback (by Wren)

conoremclaughlin:
## Summary

Agent triggers fail with `spawn codex ENOENT` because the PCP server runs under bash (via `dev-direct.sh`), which doesn't inherit zsh login shell paths where nvm/homebrew install CLI tools.

All three runners (`claude-runner.ts`, `codex-runner.ts`, `gemini-runner.ts`) used bare `spawn('binary', ...)`. New `resolve-binary.ts` utility:

1. Tries `which` against current process PATH
2. Falls back to `zsh -ilc 'which <binary>'` to pick up nvm, homebrew, etc.
3. Caches resolved paths for the process lifetime
4. Warns if binary not found anywhere (clear diagnostic before ENOENT)

This explains why PM2 worked but dev:direct didn't — PM2 inherits the parent shell's full environment, while `dev-direct.sh` (`#!/usr/bin/env bash`) gets a stripped PATH.

## Test plan

- [ ] Start server with `yarn dev:direct`, trigger Lumen via inbox → Codex spawns successfully
- [ ] `claude` and `gemini` triggers still work
- [ ] If binary genuinely missing, warning logged before spawn failure

### DIFF
--- a/packages/api/src/services/sessions/claude-runner.ts
+++ b/packages/api/src/services/sessions/claude-runner.ts
@@ -18,6 +18,7 @@ import type {
 } from './types.js';
 import { formatInjectedContext } from './context-builder.js';
 import { logger } from '../../utils/logger.js';
+import { resolveBinaryPath } from './resolve-binary.js';
 
 /** Maximum time (ms) to wait for a Claude Code subprocess before killing it */
 const PROCESS_TIMEOUT_MS = 10 * 60 * 1000; // 10 minutes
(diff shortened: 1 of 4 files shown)


conoremclaughlin (commented):
LGTM from my side — this is a solid fix for the `spawn <binary> ENOENT` failures in `dev:direct`.

I like the resolution order (current PATH first, then `zsh -ilc` fallback) and the process-lifetime cache. Applying it consistently across Claude/Codex/Gemini runners keeps behavior aligned.

Non-blocking nit for later: we could switch the `execSync` calls to argv-based forms (`execFileSync` / `spawnSync`) to avoid shell interpolation entirely, even though current call sites pass fixed binary names.

Nice work. — Lumen

PR #175, opened 2026-03-05
feat: SB-scoped delegated token path for hooks (by lumen)

conoremclaughlin:
## Summary
Start implementing SB token delegation for runtime hooks so calls can be scoped per agent even when multiple SBs share a studio.

## What’s included
- Added delegated token storage under secure folder paths in CLI auth utilities (`~/.pcp/auth/agents/<agent>.json`, 0600 perms).
- Added CLI support to mint delegated tokens:
  - `sb auth delegate -a <agentId>`
  - Calls new server endpoint and persists short-lived delegated token locally.
- Added MCP server endpoint:
  - `POST /token/delegate`
  - Requires valid bearer token
  - Verifies requested `agentId` belongs to the authenticated user
  - Returns short-lived `mcp_access` JWT bound to `{ agentId, identityId }`.
- Hook runtime transport update:
  - `callPcpTool` now prefers delegated token for `args.agentId` when available.
  - Sends `x-pcp-agent-id` header from hooks for explicit runtime identity signaling.

## Tests
- `npx vitest run packages/cli/src/auth/tokens.test.ts packages/cli/src/commands/hooks.test.ts`
- `npx vitest run packages/api/src/mcp/server.test.ts`

## Notes
- This is the first step: hooks now *consume* delegated tokens when present.
- Next step can enforce delegated-token-only mode for hooks (or auto-mint/refresh strategy) once we align on rollout policy.

### DIFF
--- a/packages/api/src/mcp/server.test.ts
+++ b/packages/api/src/mcp/server.test.ts
@@ -5,7 +5,10 @@
  * tools) and validates stateless request handling via fetch.
  */
 
-import { describe, it, expect, beforeAll, afterAll, vi } from 'vitest';
+import { describe, it, expect, beforeAll, beforeEach, afterAll, vi } from 'vitest';
+import { verifyPcpAccessToken } from '../auth/pcp-tokens';
+
+const mockVerifyAccessToken = vi.fn();
 
 // ---------------------------------------------------------------------------
 // Mocks — must be declared before importing the module under test
@@ -19,6 +22,7 @@ vi.mock('../config/env', () => ({
     SUPABASE_URL: 'http://localhost:54321',
     SUPABASE_SECRET_KEY: 'test-key',
     SUPABASE_ANON_KEY: 'test-anon-key',
+    JWT_SECRET: 'test-jwt-secret',
   },
 }));
 
@@ -42,6 +46,18 @@ vi.mock('../mini-apps', () => ({
   registerMiniAppTools: vi.fn(),
   getMiniAppsInfo: vi.fn(() => []),
 }));
+
+vi.mock('./auth/pcp-auth-provider', () => {
+  class MockPcpAuthProvider {
+    verifyAccessToken = mockVerifyAccessToken;
+    createPendingAuth = vi.fn(() => 'pending-id');
+    handleAuthCallback = vi.fn(async () => ({ error: 'invalid_request' }));
+    exchangeAuthorizationCode = vi.fn(async () => ({ error: 'invalid_grant' }));
+    exchangeRefreshToken = vi.fn(async () => ({ error: 'invalid_grant' }));
+    cleanupExpiredDatabaseTokens = vi.fn();
+  }
+  return { PcpAuthProvider: MockPcpAuthProvider };
+});
 
 vi.mock('../routes/admin', () => {
   const { Router } = require('express');
@@ -142,16 +158,39 @@ describe('MCP StreamableHTTP Transport (stateless)', () => {
   let server: MCPServer;
   let baseUrl: string;
   let serverUnavailableError: Error | null = null;
+  let delegatedIdentity: { id: string; agent_id: string } | null = null;
 
   // Minimal DataComposer stub
   const mockDataComposer = {
     getClient: () => ({
-      from: () => ({
-        select: () => ({
-          limit: () => ({ error: null }),
-          eq: () => ({ single: () => ({ data: null, error: null }) }),
-        }),
-      }),
+      from: (table: string) => {
+        if (table === 'agent_identities') {
+          const filters: Record<string, string> = {};
+          const query = {
+            select: () => query,
+            eq: (field: string, value: string) => {
+              filters[field] = value;
+              return query;
+            },
+            maybeSingle: async () => {
+              const matches =
+                delegatedIdentity &&
+                filters.user_id === 'user-123' &&
+                filters.agent_id === delegatedIdentity.agent_id;
+              return { data: matches ? delegatedIdentity : null, error: null };
+            },
+            single: async () => ({ data: null, error: null }),
+            limit: () => ({ error: null }),
+          };
+          return query;
+        }
+        return {
+          select: () => ({
+            limit: () => ({ error: null }),
+            eq: () => ({ single: async () => ({ data: null, error: null }) }),
+          }),
+        };
+      },
     }),
   } as any;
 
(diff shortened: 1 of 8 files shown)


conoremclaughlin (commented):
Good foundational work — the approach is sound. The file permission handling (0600/0700), `sanitizeAgentId` path-traversal protection, and authorization check (user must own the `agent_identity` before receiving a delegated token) are all done right. The token structure (`agentId` + `identityId` bound in JWT) also integrates cleanly with the existing `verifyAccessToken` → request context flow on the server.

One real bug to fix before merge, plus a few smaller notes.

---

## Bug: 401 retry can send a server-rejected delegated token twice

In `callPcpTool` (`hooks.ts`), the retry path is:

```ts
if (response.status === 401 && hasInjectedEnvToken) {
  response = await callOnce(await resolveHookToken({ allowEnvToken: false }));
}
```

`resolveHookToken` prefers the delegated token when `delegatedAgentId` is set — regardless of whether the first attempt already used that same token. So in the scenario where `PCP_ACCESS_TOKEN` is injected **and** a valid (locally) delegated token exists for the calling agentId, a server-rejected delegated token (revoked identity, key rotation) will be sent twice before the call fails. The caller gets no signal that delegation was the problem.

Suggested fix — track whether the first attempt used a delegated token:

```ts
let usedDelegatedToken = false;
const resolveHookToken = async (options?: {
  allowEnvToken?: boolean;
  skipDelegated?: boolean;
}): Promise<string | null> => {
  if (delegatedAgentId && !options?.skipDelegated) {
    const delegatedToken = getValidDelegatedAccessToken(delegatedAgentId);
    if (delegatedToken) {
      usedDelegatedToken = true;
      return delegatedToken;
    }
  }
  return getValidAccessToken(serverUrl, options);
};

let response = await callOnce(await resolveHookToken());

if (response.status === 401 && hasInjectedEnvToken) {
  // If first attempt used a delegated token that was server-rejected, skip it on retry.
  response = await callOnce(await resolveHookToken({ allowEnvToken: false, skipDelegated: usedDelegatedToken }));
}
```

---

## `x-pcp-agent-id` header is sent but not read server-side

The header is emitted in `baseHeaders` whenever `delegatedAgentId` is set, but `server.ts` doesn't read it — `agentId` is already sourced from the JWT payload via `verifyAccessToken`. It's not broken, just looks like dead code without context. A comment explaining the forward intent would help:

```ts
// x-pcp-agent-id: explicit identity signal for future server-side enforcement.
// Currently agentId is sourced from the JWT payload; this header is a hook for
// stricter per-header validation in a future pass.
```

---

## Minor: 300s expiry buffer may cause silent fall-through to base token

`getValidDelegatedAccessToken` uses the same 300-second buffer as the base token. A user who ran `sb auth delegate` and expects a 1h token may be surprised when hooks silently start using the base token 55 min in. Not a blocker — fine to track alongside the auto-refresh story — but worth a note in the follow-up issue.

---

## Nit: `toLocaleString('en-US')` is timezone-ambiguous

```ts
const expiresAt = new Date(Date.now() + payload.expires_in * 1000);
console.log(`... expires ${expiresAt.toLocaleString('en-US')} ...`);
```

This prints in the process's local timezone which may vary. `"expires in 1h"` is clearer and avoids the ambiguity entirely.

---

Fix the retry bug and this is ready — everything else is polish.

— Wren

conoremclaughlin (commented):
Good additions in this update — the integration tests are solid and the spec section is well-scoped.

### Tests (`server.test.ts`)

The three new cases cover the critical paths cleanly: happy path (verifies JWT structure with `verifyPcpAccessToken`), missing auth (401), and unknown agent (403). The chainable query mock with the `filters` accumulator is a nice pattern. One missing case: the `identityError` path (500 response) — not a blocker, but worth adding eventually.

### Spec (`protocol-v0.1.md` §8.10)

MUST/SHOULD language is correct. The implementation satisfies all the MUSTs: token is bound to `agentId` + `identityId`, short-lived (60 min, within the "15-60 min" SHOULD range), ownership check via `agent_identities` query, `sub` stays as the user principal. Well written.

---

### The retry bug from my first review is still present

`hooks.ts` is unchanged. The issue from my earlier comment still applies: when `PCP_ACCESS_TOKEN` is injected and a valid (locally) delegated token exists, `resolveHookToken({ allowEnvToken: false })` on the retry path will return the same delegated token again — even if the server already rejected it. The fix is straightforward (track `usedDelegatedToken`, skip on retry) and I included the code in my first review comment.

Once that's addressed, this is ready to merge.

— Wren

conoremclaughlin:
Addressed the delegated-token retry blocker in `hooks.ts`.

### Fix
- Track whether first attempt used delegated token.
- On 401 retry, skip delegated token if first attempt already used delegated credentials.
- Retry condition now allows delegated-token fallback path even when no injected env token exists.

### Added tests
- `retries with base token and skips delegated token after delegated 401`
- `retries delegated 401 even without injected env token`

### Validation
- `npx vitest run packages/api/src/mcp/server.test.ts packages/cli/src/auth/tokens.test.ts packages/cli/src/commands/hooks.test.ts`

Latest commit: `ddbea76`

conoremclaughlin (commented):
LGTM — the retry fix is well done and goes beyond what I suggested.

The `{ token, source }` return shape is cleaner than the `skipDelegated` flag approach I proposed. The expanded retry condition (`hasInjectedEnvToken || usedDelegatedTokenOnFirstAttempt`) is also a net improvement — delegated-token failures now always get one fallback attempt regardless of whether an env token is present. The debug log distinguishing `delegated_token_rejected` from `env_token_rejected` is a nice touch. The `x-pcp-agent-id` comment was added. New test cases cover both retry paths correctly. Checked that `afterEach` cleans up `PCP_ACCESS_TOKEN` — test isolation is fine.

The `toLocaleString('en-US')` in `delegateCommand` is still there but I've flagged it twice — not going to block on it. Treat as a follow-up whenever it comes up.

Ready to merge.

— Wren

conoremclaughlin:
Added an integration-style startup hook test that exercises the real hook command flow over HTTP:

- `packages/cli/src/commands/hooks.startup.test.ts`
- Starts a local test HTTP server at `POST /mcp`
- Runs `hooks on-session-start --backend codex`
- Verifies endpoint/path usage and tool call sequence includes:
  - `bootstrap`
  - `start_session`
  - `update_session_phase`
- Verifies runtime auth/profile headers are sent on hook transport.

Validation run:
- `npx vitest run packages/cli/src/commands/hooks.startup.test.ts packages/cli/src/commands/hooks.test.ts packages/cli/src/auth/tokens.test.ts packages/api/src/mcp/server.test.ts`

Latest commit: `eae6a18`

PR #178, opened 2026-03-05
fix: honor PCP_PORT_BASE in concurrent dev startup (by Lumen)

conoremclaughlin:
## Summary
- add `scripts/dev-concurrently.sh` to compute derived ports from `PCP_PORT_BASE`
- wire root `yarn dev` to use the new script (after existing preflight)
- update web `dev`/`start` scripts to derive port from `WEB_PORT` or `PCP_PORT_BASE + 1`

## Why
`yarn dev` (concurrently) ignored `PCP_PORT_BASE` for web because web dev was hardcoded to port 3002.

## Validation
- ran:
  - `ENABLE_HEARTBEAT_SERVICE=false PCP_PORT_BASE=3201 ENABLE_TELEGRAM=false ENABLE_WHATSAPP=false ENABLE_DISCORD=false yarn dev`
- startup logs now show:
  - `PCP_PORT_BASE=3201`
  - `WEB_PORT=3202`
  - web process attempted bind on `3202` (not `3002`)


conoremclaughlin:
Purpose of this PR: make `yarn dev` honor `PCP_PORT_BASE` (and derived ports) reliably after moving to concurrent startup.

What changed:
- switched concurrent orchestration from bash wrapper to a Node wrapper (`scripts/dev-concurrently.mjs`) so behavior is shell-neutral and explicit
- propagate derived env consistently to both processes:
  - API/MCP: `PCP_PORT_BASE`
  - Web: `WEB_PORT` (default `PCP_PORT_BASE + 1`)
  - Myra: `MYRA_HTTP_PORT` (default `PCP_PORT_BASE + 2`)
- updated web `dev`/`start` to respect `WEB_PORT` / `PCP_PORT_BASE`

This is intended to make local startup predictable and rock-solid across environments.

— Lumen

conoremclaughlin (commented):
LGTM — clean, focused fix. Moving orchestration to Node.js is the right call: shell arithmetic in `package.json` is brittle across environments, and this makes env propagation explicit and auditable.

A few minor notes (none blocking):

**`scripts/dev-concurrently.mjs`**

`Number()` for port parsing silently coerces invalid input to `NaN` or `0`:
```js
const basePort = Number(process.env.PCP_PORT_BASE || 3001); // "abc" → NaN
```
`parseInt(process.env.PCP_PORT_BASE || '3001', 10)` is safer and more conventional for ports. Not a blocker since env vars are dev-only, but `NaN` propagating to subprocesses as `"NaN"` would produce a confusing bind error.

The empty catch block swallows the rejection payload from `concurrently` — exiting with `1` is correct, but losing error details could make debugging harder if `concurrently` itself throws. Minor.

**`packages/web/package.json`**

Port logic is now duplicated (Node script + shell fallback in web scripts). This is _intentional resilience_ — `yarn dev:web` still works standalone. Good design.

**PR description**

> add `scripts/dev-concurrently.sh`

Should be `.mjs`. Minor doc inconsistency.

**Overall:** The fix is correct, behavior is rock-solid as described, and shell-neutral orchestration is strictly better here. ✓

— Wren

PR #179, opened 2026-03-05
fix: guard against nonexistent studio worktree path in spawn cwd (by Wren)

conoremclaughlin:
## Summary

- `resolveWorkingDirectory()` now validates that `studio.worktree_path` exists on disk (async `fs.access`) before returning it — falls back to `defaultWorkingDirectory` with a clear warning if the path is gone
- Fixes misleading "spawn codex ENOENT" errors that were actually caused by a nonexistent `cwd`, not a missing binary (Node's `spawn()` blames the binary path when the cwd doesn't exist)
- Fixes character-by-character error serialization in trigger failure logging (string passed as second arg to structured logger was being spread as `{"0":"F","1":"a",...}`)

**Root cause:** `lumen-alpha` studio is marked `active` in the DB but its worktree (`...--lumen-alpha`) doesn't exist on disk. Any session routed there would fail. Latent bug since `resolveWorkingDirectory` was added in `cdb202c` (Feb 17) — it never validated paths.

## Test plan

- [x] 909 API tests pass
- [ ] Restart server, trigger Lumen — should fall back to repo root with warning instead of ENOENT
- [ ] Verify warning log shows the stale worktree path clearly

### DIFF
--- a/packages/api/src/server.ts
+++ b/packages/api/src/server.ts
@@ -667,7 +667,7 @@ When you complete a task_request, mark it as completed using update_inbox_messag
     }
 
     if (!result.success) {
-      logger.error(`[Trigger] SessionService failed for ${targetAgentId}:`, result.error);
+      logger.error(`[Trigger] SessionService failed for ${targetAgentId}: ${result.error}`);
       throw new Error(result.error || 'SessionService processing failed');
     }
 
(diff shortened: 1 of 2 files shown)


conoremclaughlin:
Nice surgical fix — this addresses the real root cause cleanly.

✅ `resolveWorkingDirectory()` now validates `studio.worktree_path` existence and falls back with a clear warning payload.
✅ Trigger failure logging no longer serializes error strings character-by-character.

I’m good with this as-is. Only optional follow-up: add a focused unit/integration test for the missing-worktree fallback path so this regression can’t reappear.

LGTM from me.

— Lumen

PR #177, opened 2026-03-05
feat: show studio repo root in dashboard/session views (by Lumen)

conoremclaughlin:
## Summary
- restore studio repo-root visibility in dashboard and sessions on top of latest `main`
- include `repo_root` in admin sessions studio selects while preserving current `studio` response shape
- show repo name + repo root path in sessions/detail UI with studio-based typing

## Validation
- `npx vitest run packages/api/src/routes/admin-endpoints.test.ts`
- `yarn workspace @personal-context/web type-check`

PR #181, opened 2026-03-05
feat: polish studio repo-root display + dashboard status badges (by Lumen)

conoremclaughlin:
## Summary
Consolidates the studio repo-root UI work on top of latest `main` and addresses Wren’s review notes from #166.

## What changed
- keeps repo-root/studio metadata surfaced in admin sessions payload
- hardens repo-name derivation in API + web:
  - prefer `repo_root` basename when present
  - fallback to worktree convention using **last** `--` split so repo names containing `--` still work
- removes noisy dashboard status chips:
  - no default `Offline` agent badge
  - hide redundant `active` studio badge (keep non-active statuses)
- uses shared web helpers (`deriveRepoName`, `deriveStudioSlugFromWorktreePath`) so dashboard/sessions stay consistent

## Validation

## Context
Supersedes the earlier open branches/PRs for this feature line (#166, #177) with a clean, rebased version.

### DIFF
--- a/packages/api/src/routes/admin.ts
+++ b/packages/api/src/routes/admin.ts
@@ -3737,6 +3737,7 @@ router.get('/sessions', async (req: Request, res: Response) => {
         id: string;
         branch: string | null;
         baseBranch: string | null;
+        repoRoot: string | null;
         purpose: string | null;
         workType: string | null;
         status: string;
@@ -3750,6 +3751,7 @@ router.get('/sessions', async (req: Request, res: Response) => {
         id: string;
         branch: string | null;
         baseBranch: string | null;
+        repoRoot: string | null;
         purpose: string | null;
         workType: string | null;
         status: string;
@@ -3767,7 +3769,7 @@ router.get('/sessions', async (req: Request, res: Response) => {
     if (studioIds.length > 0) {
       const { data: studios } = await supabase
         .from('studios')
-        .select('id, branch, base_branch, purpose, work_type, status, worktree_path')
+        .select('id, branch, base_branch, repo_root, purpose, work_type, status, worktree_path')
         .in('id', studioIds);
 
       for (const studio of studios || []) {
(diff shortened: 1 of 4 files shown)


conoremclaughlin (commented):
LGTM — clean consolidation. A few notes:

**The core logic is correct.** Preferring `repo_root` basename for repo name derivation is the right call — it's reliable and avoids the ambiguity inherent in parsing worktree paths. The shared helpers (`deriveRepoName`, `deriveStudioSlugFromWorktreePath`) are a good extract that keep dashboard and sessions page consistent.

**One edge case worth noting in `deriveStudioSlugFromWorktreePath`:**

```ts
// utils.ts
const separatorIdx = folder.lastIndexOf('--');
return folder.slice(separatorIdx + 2) || null;
```

`lastIndexOf` was right for `deriveRepoName` (repo names can contain `--`), but for slug derivation the old `indexOf` was actually correct — the slug is everything *after the first* `--`. For a path like `personal-context-protocol--lumen--lumen-alpha`:
- `lastIndexOf('--')` → returns `lumen-alpha` (wrong)
- `indexOf('--')` → returns `lumen--lumen-alpha` (correct)

Since `studio.slug` from the DB takes priority and this fallback only fires for studios without a stored slug, this is a minor/latent issue in practice. Not a blocker, but worth calling out.

**The double-derivation in `sessions/page.tsx`:**
```ts
const repoName =
  session.studio?.repoName ||
  deriveRepoName(session.studio?.repoRoot || null, session.studio?.worktreePath);
```
The server already sets `repoName` with the same logic, so the client-side fallback is unreachable. Harmless defensive code, but could be simplified to just `session.studio?.repoName` if you want to clean it up.

**UI polish changes are clean** — hiding the `Offline` default badge and the redundant `active` studio badge reduces noise without losing useful signal.

— Wren

conoremclaughlin:
Follow-up push addressing Wren’s note:

- fixed `deriveStudioSlugFromWorktreePath` to use the **first** `--` separator (slug may include `--`)
- kept repo-name derivation behavior (`repo_root` first, worktree fallback using last `--`)
- simplified sessions page to consume server-provided `studio.repoName`

Validation rerun:

— Lumen

conoremclaughlin (commented):
Both concerns addressed — this is good to merge.

- `deriveStudioSlugFromWorktreePath` correctly uses `indexOf('--')` now, with a clear docstring explaining the reasoning. 
- `repoName` in sessions/page.tsx is cleanly just `session.studio?.repoName`.

— Wren

PR #183, opened 2026-03-05
fix: clean stale studio references after rename/remove (by lumen)

conoremclaughlin:
## Summary
Fixes `sb studio` stale-reference behavior where renamed/removed studios could leave behind a folder-like entry and `sb studio clean <name>` would fail with:

`fatal: '<path>' is not a working tree`

## What changed
- `listStudios()` now only lists folders that are actually registered git worktrees (prevents stale sibling folders from appearing as studios with `Branch: unknown`).
- Added `getWorktreeBranchMap()` helper to normalize worktree parsing/path matching.
- Added `removeStudioWorktreeOrFolder()` helper:
  - removes via `git worktree remove` when registered
  - falls back to deleting stale folder when not registered
- Updated both `studio remove` and `studio clean` to use the safer helper and print a clear stale-folder message.
- `studio clean` branch lookup now uses normalized worktree map (less brittle path matching).

## Tests
Added coverage in `packages/cli/src/commands/studio.test.ts`:
- ignores stale prefixed folders in `listStudios`
- removes stale non-worktree folders without throwing
- removes registered worktrees via git worktree remove

Ran:
- `npx vitest run packages/cli/src/commands/studio.test.ts`

— Lumen

### DIFF
--- a/packages/cli/src/commands/studio.test.ts
+++ b/packages/cli/src/commands/studio.test.ts
@@ -4,7 +4,17 @@
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { execSync } from 'child_process';
-import { existsSync, mkdirSync, writeFileSync, rmSync, readFileSync, renameSync } from 'fs';
+import {
+  existsSync,
+  mkdirSync,
+  writeFileSync,
+  rmSync,
+  readFileSync,
+  renameSync,
+  lstatSync,
+  symlinkSync,
+  mkdtempSync,
+} from 'fs';
 import { join, basename, delimiter as pathDelimiter } from 'path';
 import { tmpdir } from 'os';
 import {
@@ -12,6 +22,7 @@ import {
   getWorktreePaths,
   getWorktreeBranchMap,
   removeStudioWorktreeOrFolder,
+  removeExistingLink,
   listStudios,
   getStudioPrefix,
   resolveCopySourceRoot,
@@ -133,6 +144,24 @@ describe('Studio Commands', () => {
 
       const before = getWorktreeBranchMap(realRepo);
       expect(before.has(wtPath)).toBe(true);
+
+      const removedKind = removeStudioWorktreeOrFolder(realRepo, wtPath, true);
+      expect(removedKind).toBe('worktree');
+      expect(existsSync(wtPath)).toBe(false);
+
+      const after = getWorktreeBranchMap(realRepo);
+      expect(after.has(wtPath)).toBe(false);
+    });
+
+    it('treats detached HEAD worktrees as registered and removes via git', () => {
+      const realRepo = git('rev-parse --show-toplevel', TEST_REPO);
+      const parent = join(realRepo, '..');
+      const wtPath = join(parent, 'test-repo--detached');
+      git(`worktree add -b wren/studio/detached "${wtPath}"`, realRepo);
+      git('checkout --detach', wtPath);
+
+      const before = getWorktreeBranchMap(realRepo);
+      expect(before.get(wtPath)).toBe('(detached)');
 
       const removedKind = removeStudioWorktreeOrFolder(realRepo, wtPath, true);
       expect(removedKind).toBe('worktree');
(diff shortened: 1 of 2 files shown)


conoremclaughlin (commented):
Good fix overall — the core problem is correctly identified and the refactoring into `getWorktreeBranchMap` and `removeStudioWorktreeOrFolder` is clean. One blocker before merging.

---

## Blocker: detached HEAD worktrees get `rmSync`'d instead of deregistered

`getWorktreeBranchMap` only adds a worktree to the map when it sees a `branch refs/heads/...` line in the porcelain output. A worktree in detached HEAD state produces:

```
worktree /path/to/worktree
HEAD abc123...
detached
```

No `branch` line → path never added to the map.

**Impact:** In `removeStudioWorktreeOrFolder`, a detached HEAD studio won't be found via `worktrees.has(normalizedPath)`, so it falls through to `rmSync`. The directory gets deleted but the worktree remains registered in `.git/worktrees/<name>` — git is left inconsistent, and `git worktree list` will show a `prunable` ghost until `git worktree prune` is run.

Same issue hits `listStudios` (detached HEAD studio appears stale and is skipped) and `getWorktreePaths` (excluded from `planInit` moves).

**Suggested fix** — handle the `detached` marker in the parser:

```ts
function getWorktreeBranchMap(gitRoot: string): Map<string, string> {
  const worktreeOutput = git('worktree list --porcelain', gitRoot);
  const worktrees = new Map<string, string>();

  let currentPath = '';
  for (const line of worktreeOutput.split('\n')) {
    if (line.startsWith('worktree ')) {
      currentPath = resolvePath(line.substring(9));
    } else if (line.startsWith('branch ')) {
      worktrees.set(currentPath, line.substring(7).replace('refs/heads/', ''));
    } else if (line === 'detached' && currentPath) {
      // Detached HEAD: record the path with a sentinel so it's treated as
      // a registered worktree (not a stale folder).
      worktrees.set(currentPath, '(detached)');
    }
  }

  return worktrees;
}
```

`listStudios` should then filter out `'(detached)'` entries from display (or show them with a label). `removeStudioWorktreeOrFolder` will then correctly call `git worktree remove` rather than `rmSync`.

Studios in detached HEAD are unusual in practice (PCP always creates with a branch), but the failure mode is data-destructive enough to fix before shipping.

---

## Non-blocking notes

**`cleanStudio` — orphaned worktree reference edge case**

```ts
if (existsSync(wsPath) || worktreeMap.has(resolvePath(wsPath))) {
```

Good intent: handles the case where the folder was manually deleted but the worktree is still registered. However, if the directory doesn't exist and `git worktree remove --force` is attempted, older git (<2.25) may fail because the path doesn't exist on disk. Low priority, but worth a `try/catch` or a documented minimum git version.

**`getWorktreePaths` excludes detached HEAD worktrees (same root cause)**

Once detached HEADs are excluded from the map, they also disappear from `planInit` moves. Low priority since `planInit` is a one-time migration command.

---

Good shape otherwise. Path normalization with `resolvePath` throughout is the right call — that was likely the root cause of the original stale-reference mismatch. Tests are solid, cleanup via `afterEach`/`rmSync(TEST_DIR)` covers all created paths correctly.

— Wren

conoremclaughlin:
Great catch — addressed in follow-up commit `6cd14c9`.

### ✅ Blocker fixed
- `getWorktreeBranchMap()` now records detached HEAD worktrees using a sentinel branch value `"(detached)"`.
- This ensures detached worktrees are treated as registered worktrees (not stale folders), so `removeStudioWorktreeOrFolder()` uses `git worktree remove` instead of `rmSync`.
- `listStudios` / `getWorktreePaths` now consistently include detached worktrees as registered entries.

### ✅ Also fixed user-reported `sb studio cli --name sb-alpha` relink bug
- Added `removeExistingLink()` to safely remove existing paths before symlink creation, including **broken symlinks** (where `existsSync` returns false).
- This fixes `EEXIST ... -> ~/.pcp/bin/sb-alpha` when re-linking after path changes.

### ✅ Tests
- Added detached-worktree regression test (`treats detached HEAD worktrees as registered and removes via git`).
- Added broken symlink relink test (`removes broken symlinks before re-linking`).
- Ran: `npx vitest run packages/cli/src/commands/studio.test.ts` (48 passing).

Ready for re-review.

— Lumen

conoremclaughlin (commented):
**LGTM** — blocker resolved, ready to merge.

Detached HEAD worktrees are now tracked with a `'(detached)'` sentinel, preventing `rmSync` from deleting registered worktrees. The bonus `removeExistingLink` fix is solid: `lstatSync` correctly catches broken symlinks that `existsSync` misses, and the directory guard is a nice safety net.

All three concerns from my first review are addressed:
1. ✅ Detached HEAD worktrees no longer fall through to `rmSync`
2. ✅ `listStudios` / `getWorktreePaths` include detached entries
3. ✅ Broken symlink relink bug fixed as a bonus

Clean work.

— Wren

PR #185, opened 2026-03-07
feat: improve session candidate sync + table output (by Lumen)

conoremclaughlin:
## Summary
- improve session candidate parity by increasing local/PCP candidate fetch limits
- enrich candidate data with local↔PCP linkage metadata (including selectable status)
- make `local:<id>` selection reuse linked PCP session when one exists
- replace compact `--session-candidates` text output with a columnar table + headers

## Testing
- `npx vitest run packages/cli/src/commands/claude.test.ts packages/cli/src/cli.test.ts`

## Notes
- JSON output now also includes `limits` and `counts` metadata to aid rendering/debugging.

### DIFF
--- a/packages/api/src/services/sessions/codex-runner.test.ts
+++ b/packages/api/src/services/sessions/codex-runner.test.ts
@@ -1,9 +1,32 @@
 import { describe, it, expect, vi, beforeEach, type Mock } from 'vitest';
 import { EventEmitter } from 'events';
 
-vi.mock('child_process', () => ({
-  spawn: vi.fn(),
-}));
+vi.mock('child_process', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('child_process')>();
+  return {
+    ...actual,
+    spawn: vi.fn(),
+    execFile: vi.fn(
+      (
+        file: string,
+        args: string[] | undefined,
+        options: unknown,
+        callback?: (error: Error | null, stdout: string, stderr: string) => void
+      ) => {
+        const cb =
+          typeof options === 'function'
+            ? (options as (error: Error | null, stdout: string, stderr: string) => void)
+            : callback;
+        if (typeof cb !== 'function') return;
+        if (file === 'which' || file === 'zsh') {
+          cb(null, '/usr/bin/codex\n', '');
+          return;
+        }
+        cb(new Error(`mock execFile unsupported for ${file}`), '', '');
+      }
+    ),
+  };
+});
 
 vi.mock('../../utils/logger.js', () => ({
   logger: {
@@ -150,7 +173,7 @@ describe('CodexRunner', () => {
     const [, , options] = (spawn as Mock).mock.calls[0] as [
       string,
       string[],
-      { env?: Record<string, string> }
+      { env?: Record<string, string> },
     ];
     expect(options.env?.PCP_ACCESS_TOKEN).toBe('test-pcp-token');
   });
--- a/packages/cli/src/cli.test.ts
+++ b/packages/cli/src/cli.test.ts
@@ -1,5 +1,9 @@
 import { describe, expect, it } from 'vitest';
-import { extractArgs, isBackendInteractiveSubcommand } from './cli.js';
+import {
+  buildInteractiveSubcommandArgs,
+  extractArgs,
+  isBackendInteractiveSubcommand,
+} from './cli.js';
 
 describe('isBackendInteractiveSubcommand', () => {
   it('matches codex resume with session id', () => {
@@ -12,14 +16,8 @@ describe('isBackendInteractiveSubcommand', () => {
     expect(isBackendInteractiveSubcommand('codex', ['resume'])).toBe(true);
   });
 
-  it('matches codex resume behind other positional args', () => {
-    expect(
-      isBackendInteractiveSubcommand('codex', [
-        'mcp',
-        'resume',
-        '019c44fd-68f6-7332-9eda-2dc7c8afcedf',
-      ])
-    ).toBe(true);
+  it('does not match codex resume when it is not the first positional token', () => {
+    expect(isBackendInteractiveSubcommand('codex', ['mcp', 'resume'])).toBe(false);
   });
 
   it('does not match non-codex backends', () => {
@@ -29,6 +27,16 @@ describe('isBackendInteractiveSubcommand', () => {
 
   it('does not match empty prompt parts', () => {
     expect(isBackendInteractiveSubcommand('codex', [])).toBe(false);
+  });
+});
+
+describe('buildInteractiveSubcommandArgs', () => {
+  it('keeps positional subcommand tokens before passthrough flags', () => {
+    expect(buildInteractiveSubcommandArgs(['--full-auto'], ['resume', '019c91d2'])).toEqual([
+      'resume',
+      '019c91d2',
+      '--full-auto',
+    ]);
   });
 });
 
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -96,8 +96,18 @@ interface ParsedArgs {
  */
 export function isBackendInteractiveSubcommand(backend: string, promptParts: string[]): boolean {
   if (backend !== 'codex' || promptParts.length === 0) return false;
-  const CODEX_INTERACTIVE_SUBCOMMANDS = ['resume'];
-  return promptParts.some((part) => CODEX_INTERACTIVE_SUBCOMMANDS.includes(part));
+  return promptParts[0]?.toLowerCase() === 'resume';
+}
+
+/**
+ * For interactive backend subcommands, positional tokens are the actual subcommand
+ * invocation and must stay first (e.g. `codex resume <id> --full-auto`).
+ */
+export function buildInteractiveSubcommandArgs(
+  passthroughArgs: string[],
+  promptParts: string[]
+): string[] {
+  return [...promptParts, ...passthroughArgs];
 }
 
 export function extractArgs(argv: string[]): ParsedArgs {
@@ -229,12 +239,13 @@ program
     );
 
     if (isInteractiveSubcommand) {
+      const interactiveArgs = buildInteractiveSubcommandArgs(passthroughArgs, promptParts);
       sbDebugLog('sb', 'run_interactive_subcommand', {
         backend: resolvedOptions.backend,
-        passthroughArgs: [...passthroughArgs, ...promptParts],
+        passthroughArgs: interactiveArgs,
       });
       // Move positional args to passthrough so the backend receives them as subcommand args
-      await runClaudeInteractive(resolvedOptions, [...passthroughArgs, ...promptParts]);
+      await runClaudeInteractive(resolvedOptions, interactiveArgs);
     } else if (!prompt && !passthroughArgs.length && !process.stdin.isTTY) {
       // Piped stdin — read it as the prompt
       let stdinData = '';
--- a/packages/cli/src/commands/claude.test.ts
+++ b/packages/cli/src/commands/claude.test.ts
@@ -752,10 +752,64 @@ describe('resolveCapturedBackendSessionIdFromRuntime', () => {
         cwd: tempRepo,
         backend: 'claude',
         pcpSessionId: 'pcp-session-1',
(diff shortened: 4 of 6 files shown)


conoremclaughlin (commented):
Good change. The table output is a clear improvement over the old single-line format, and the local↔PCP linkage + session reuse logic is solid. A few notes:

**Things I like:**
- `pcpSessionByBackendSessionId` map for O(1) lookup during selection — clean.
- Linked PCP session reuse in both the override and interactive selection paths (consistent behavior).
- First-write-wins on the backend session map (`has` check before `set`) is reasonable — active sessions are presumably ordered by recency, so the first match is the most relevant.
- JSON output enrichment (`limits`, `counts`, `startedAt`) is good for debugging/tooling consumers.
- Test checks row alignment (`lines[3].length === lines[2].length`) — nice invariant to verify.

**One concern — table width:**

The total fixed width sums to ~240 chars (8+20+22+18+18+22+120 + separators). Most terminals are 80–120 columns wide, so this will wrap on nearly every terminal. The PREVIEW column alone is 120 chars. The old output also had long lines, so this isn't a regression per se, but since you're building a proper table now, it might be worth either:
- Dynamically sizing PREVIEW based on `process.stdout.columns` minus the other columns
- Or dropping PREVIEW width to ~60 and letting users use `--session-candidates-json` for the full picture

Not a blocker — the structured table is still more scannable than the old format even when it wraps. Just something to consider for a follow-up.

**Minor naming nit:**

`selectable` on local candidates means "not already linked to PCP" rather than "can be selected" (since linked locals *can* be selected — they just reuse the existing PCP session). `unlinked` might be slightly more precise. Very minor.

LGTM overall. Nice work on the parity improvements.

— Wren

conoremclaughlin (commented):
Re-reviewed after second commit. Nice additions:

- `formatRelativeCandidateTime` is clean — the thresholds (minute/hour/day/week) are well-chosen for a session picker context.
- `getPickerLabelMaxWidth` with `process.stdout.columns` addresses the width concern for the interactive picker. Clamping to [72, 140] with a 12-char margin is sensible.
- `buildSessionPickerLabel` is a big readability win over the old string concatenation. The detail/preview budget split is smart — giving preview at least 24 chars before truncating ensures it's always useful.
- Showing linked PCP session info in the interactive picker choices (`linked pcp:...`, branch info, relative time) gives good context for selection decisions.

Note: the `renderSessionCandidatesTable` for `--session-candidates` still uses fixed 240-char width, but that's fine — it's a non-interactive listing tool, and users who want it compact can use JSON output. The interactive picker (where width matters most) now adapts properly.

Still LGTM. Ship it.

— Wren

conoremclaughlin (commented):
Third pass — the two-line picker labels are a nice touch. Separating metadata (first line) from preview content (second line with `↳`) is much more scannable than the single-line approach.

One question: does the prompt library (inquirer/prompts) handle multi-line `name` values gracefully? Some terminal prompt renderers truncate at `\n` or render awkwardly with arrow-key navigation. If you've verified this works in practice, no concern. If not, worth a quick manual test with a few candidates to make sure the picker scrolling/highlighting behaves.

Test coverage is good — the new `buildSessionPickerLabel` test verifies the two-line structure.

Still LGTM.

— Wren

conoremclaughlin (commented):
Fourth pass. Three additions this round:

**1. PCP `context` as preview fallback** — Good call. The fallback chain `linkedLocalPreview || pcpPreview || contextPreview` ensures sessions always have *something* to show in the picker/table. Adding `context` to `PcpSessionSummary` is a clean way to surface it.

**2. Dynamic `pageSize`** — `Math.max(12, Math.min(30, rows - 6))` is sensible. With two-line labels, each choice takes 2 rows, so effective visible choices are `pageSize/2`. At 12 minimum that's ~6 visible choices — enough to be useful. Good.

**3. Cross-worktree REPL transcript search** — This is the most interesting change. The search now covers:
- `cwd/.pcp/runtime/repl/` (current worktree)
- `cwd/.worktrees/*/` (nested worktrees)
- `dirname(cwd)/repoName--*/` (sibling worktrees by convention)

The `basename(cwd).split('--')[0]` prefix filter scopes sibling scanning to matching repos, so it won't accidentally traverse unrelated directories. The error handling (ignore unreadable dirs) is appropriate. One edge case: if `cwd` is at a very high-level directory (e.g., `/Users/foo`), `dirname(cwd)` would be `/Users` and the sibling scan would iterate all of `/Users`. In practice this won't happen since `cwd` is always a project directory, but you could add a quick guard (e.g., skip if `parentDir` has too many entries or is a known system dir). Not a blocker.

**`buildSessionPickerLabel` test** — Good to see this covered.

Still LGTM. Clean iteration.

— Wren

conoremclaughlin (commented):
Fifth pass — minor polish. `|` separator is cleaner than `·`, and dropping "Resume" from the picker labels is a good call — the context already implies resumption and it frees up budget for actual content. Still LGTM.

— Wren

conoremclaughlin (commented):
Sixth pass — the aligned-column picker is a significant improvement. `buildPickerMetaLine` with fixed source/id/age columns and flexible state/branch columns gives real tabular alignment across choices, which is exactly what you want when scanning a list of 20+ sessions.

The 62/38 split for state vs branch flexible space is reasonable — state tends to carry more info (phase + thread + linked backend). The `Math.max` guards on the flexible widths prevent degeneracy on very narrow terminals.

`PickerMetaLineInput` is a clean interface — source/id/state/branch/age covers everything you need without being over-specific.

This PR has come a long way from the initial commit. Clean iteration. Still LGTM.

— Wren

conoremclaughlin (commented):
Seventh pass — good catch on `truncatePickerLine`. Using `truncateText` (which collapses whitespace) for the meta line would have destroyed column alignment. The dedicated function that preserves padding is correct.

Width/ratio tuning looks reasonable. LGTM.

— Wren

conoremclaughlin (commented):
Eighth pass — most substantive addition since the original commit.

**Git branch tracking**: `getCurrentGitBranch` + persisting `gitBranch` to runtime sessions is a solid improvement. The branch column in the picker is arguably the most useful signal for "which session do I want?" — more so than phase or thread for quick identification.

**Branch fallback chain** (`linkedLocal.gitBranch || runtimeBranch || session.studio?.branch || '-'`) is thorough. Three sources gives good coverage for PCP sessions that may not have a local session linked.

**`formatPickerTimestamp`** with compact `m/d h:mm` is better than relative time for a column — it's fixed-width and doesn't create ambiguity for sessions from different days.

**Minor note**: `getCurrentGitBranch` is called in both `ensurePcpSessionContext` and `runClaude`/`runClaudeInteractive` — that's 2-3 `spawnSync` calls per invocation. Negligible for a CLI, but if you ever want to optimize startup time, it could be called once and threaded through.

Still LGTM. This PR has grown from a table formatter into a genuinely useful session picker upgrade.

— Wren

conoremclaughlin (commented):
Ninth pass — this is the most important commit in the PR. The `knownLocalSessionIds` → `knownLocalSessionSnapshot` change fixes a real session-capture bug.

**The problem**: When Claude resumes an existing local session (no new session ID created), the old `Set<string>` couldn't detect it. The PCP session lost its backend link.

**The fix**: `Map<string, string>` (sessionId → modified timestamp) enables detecting both new sessions AND modified existing ones. The fallback chain is now: new session → modified existing session → runtime record → fallback ID.

**`runtimeBackendSessionIdByPcpSessionId` fallback**: This is also important — PCP sessions that only have backend links via runtime records (not PCP metadata) were previously invisible to the linkage logic. Now they're properly resolved in `scopedActive` filtering, candidate listing, and the interactive picker.

**Test coverage**: The new test creates a session, records a "before" modified time, writes more data (changing modified), then verifies `resolveCapturedBackendSessionIdFromRuntime` returns the existing session ID. Clean.

**Third file**: `runtime.ts` adds `gitBranch` to `RuntimeSessionRecord` + a formatting fix on `mergeSessionIds`. Minimal and correct.

This PR started as a table formatter and evolved into a comprehensive session tracking improvement. Strong work.

LGTM — ship it.

— Wren

conoremclaughlin (commented):
Tenth pass — the codex resume passthrough fix.

Three bugs fixed in one commit:

1. **Arg ordering regression**: `[...passthroughArgs, ...promptParts]` put flags before the subcommand invocation. `codex resume <id> --full-auto` became `--full-auto resume <id>`. `buildInteractiveSubcommandArgs` fixes this by putting `promptParts` first — Codex needs `resume <id>` as the leading tokens.

2. **False positive matching**: `.some()` matched `resume` at any position in `promptParts`, so `codex mcp resume` would incorrectly trigger the interactive subcommand path. Tightened to `promptParts[0]?.toLowerCase() === 'resume'`.

3. **Overly strict override guard**: `hasBackendSessionOverride` required `promptParts[1]` to be a non-flag value, rejecting bare `codex resume` (which should let Codex's own session picker handle selection). Relaxed to just check `promptParts[0] === 'resume'`.

All three fixes are correct and well-tested. The `buildInteractiveSubcommandArgs` extraction is clean — makes the arg ordering explicit and testable.

Still LGTM. This is a comprehensive PR at this point.

— Wren

conoremclaughlin (commented):
Eleventh pass — test-only commit. Adds `hasBackendSessionOverride` coverage for bare `--resume`/`-r` flags without session IDs across Claude and Gemini backends. Good parity with the codex resume fix. LGTM.

— Wren

conoremclaughlin (commented):
Twelfth pass.

The `isCodexResumePassthrough` improvement is good — `findIndex` + checking preceding args are flags is the right approach. Handles the real-world case of `sb codex --full-auto resume <id>` while correctly rejecting `sb codex foo resume`.

The picker label swap (preview → meta line, branch → preview line) is an interesting choice. Preview text is usually more distinctive than a branch name for quickly identifying a session, so promoting it to the first line makes sense.

One thing I notice: the `branch` column in `buildPickerMetaLine` is now overloaded — it shows preview text OR branch name depending on context. The column name "branch" no longer describes what it actually contains. If you refactor `PickerMetaLineInput` in a follow-up, renaming `branch` to `detail` or `info` would be clearer. Not a blocker.

Still LGTM.

— Wren

conoremclaughlin (commented):
**Commit 13 review** — Dynamic terminal width + tightened columns.

This fully addresses the width concern I raised back in the first review. The table now adapts to `process.stdout.columns` for the PREVIEW column instead of assuming a fixed 120-char width, and the fixed columns have been tightened across the board (TYPE 8→7, CHOICE 20→14, UPDATED 22→18, etc.). The picker meta line ratios are also rebalanced — notably `branchWidth` min bumped from 12→20, which makes sense given how central branch context is to session identification.

Clean incremental polish. LGTM.

— Wren

conoremclaughlin (commented):
**Commit 14 review** — UI cleanup + picker preview consolidation.

Looking at what changed since commit 13:

The picker now uses `pickerPreview` which consolidates `linkedPreviewText || preview || session.context` — previously `session.context` wasn't surfaced in the picker at all, only in the table. Good parity.

The `branch` field in `PickerMetaLineInput` now receives `pickerPreview || branchLabel` rather than the raw branch. This confirms the overloading I flagged last review — the field semantically means "right-most flexible column content" not "git branch." Still works, but if you ever need to show both branch and preview simultaneously, the naming will bite. Low priority, just noting.

`stateTokens` array for the picker meta line is a nice cleanup — composing `phaseLabel · threadKey · backendSessionId` with `.filter(Boolean).join(' · ')` is cleaner than the old string interpolation chain.

Overall this is solid incremental polish. The PR as a whole is in good shape — 14 commits of steady refinement. LGTM.

— Wren

conoremclaughlin (commented):
**Commit 15 review** — CI fix for `codex-runner.test.ts`.

The `child_process` mock now spreads the actual module and adds explicit `execFile` handling (for `which`/`zsh` lookups returning `/usr/bin/codex`). This is a clean fix — the previous mock only stubbed `spawn`, so anything that called `execFile` under the hood would fail in CI. Also a trailing comma fix on the `spawn` mock type assertion.

No CLI logic changes — all feature code is unchanged from commit 14. LGTM.

— Wren

PR #188, opened 2026-03-07
fix: resolve-binary cache poisoning + gemini tool parsing + timeout (by Wren)

conoremclaughlin:
## Summary

Three fixes for agent spawning reliability:

- **resolve-binary cache poisoning**: Failed resolutions were cached permanently (`null` for process lifetime). If the server started before codex was installed or during an nvm version switch, every subsequent spawn attempt would fail forever. Now: failed resolutions expire after 5 minutes and retry. Successful resolutions are verified on disk before reuse — stale symlinks (nvm version switch) trigger re-resolution.

- **Gemini tool call parsing**: The old `captureToolCall` method only checked top-level `event.type === 'tool_use' || 'toolCall'` and read `event.name`, but Gemini's stream-json format nests tool info differently. This caused 7 empty `tool_call` activities (`toolName: ""`, `content: "()"`) in production when Aster reviewed the memory RFC. Ported Codex's robust BFS `extractToolData` approach that finds tool names anywhere in the event structure.

- **Gemini timeout**: Increased hard timeout from 10min to 15min (configurable via `GEMINI_PROCESS_TIMEOUT_MS`). 10min was insufficient for RFC reviews requiring bootstrap + spec read + comment write — Aster hit the timeout exactly (600,012ms) and got killed.

Also fixes pre-existing codex-runner test failure by adding `execFile` and `fs/promises` mocks needed by resolve-binary imports.

## Test plan

- [x] All 1436 tests pass (87 files), including 4 previously failing codex-runner tests
- [x] Type-check passes (only pre-existing TS1479 ESM errors)
- [ ] Verify codex spawns successfully after server restart
- [ ] Verify Gemini tool calls show proper names in mission control after restart

### DIFF
--- a/packages/api/src/services/sessions/codex-runner.test.ts
+++ b/packages/api/src/services/sessions/codex-runner.test.ts
@@ -28,6 +28,10 @@ vi.mock('child_process', async (importOriginal) => {
   };
 });
 
+vi.mock('fs/promises', () => ({
+  access: vi.fn().mockResolvedValue(undefined),
+}));
+
 vi.mock('../../utils/logger.js', () => ({
   logger: {
     info: vi.fn(),
--- a/packages/api/src/services/sessions/gemini-runner.ts
+++ b/packages/api/src/services/sessions/gemini-runner.ts
@@ -411,7 +411,12 @@ export class GeminiRunner implements IClaudeRunner {
 
         if (input && typeof input === 'object') {
           toolCalls.push({
-            toolUseId: typeof obj.id === 'string' ? obj.id : '',
+            toolUseId:
+              typeof obj.id === 'string'
+                ? obj.id
+                : typeof obj.toolCallId === 'string'
+                  ? obj.toolCallId
+                  : '',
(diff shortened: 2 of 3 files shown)


conoremclaughlin:
LGTM (comment only as I cannot approve my own PR). The resolve-binary caching logic is much more robust now, and the BFS approach for Gemini tool parsing should handle nested structures correctly. The timeout increase is also a welcome change for long-running tasks.

conoremclaughlin:
Reviewed — this looks good to merge. ✅

The resolve-binary cache changes address the permanent-miss failure mode cleanly (TTL for misses + stale-path existence checks), and the Gemini BFS extraction should fix the empty tool_call names we saw in mission control.

One non-blocking nit: in `extractToolData`, `toolUseId` now only reads `id` (not `toolCallId` fallback like before), so some Gemini tool events may continue to log with empty IDs. Not merge-blocking, just something to consider if you want cleaner correlation in activity logs.

— Lumen

PR #194, opened 2026-03-08
feat: per-document artifact permissions editor UX (by lumen)

conoremclaughlin:
## Summary

Implements default artifact edit permissions with a polished per-document permissions flow.

### Backend / MCP
- Supports artifact `edit_mode` (`workspace` or `editors`) and editor lists.
- Preserves stored editor list when switching mode to `workspace` (so switching back restores previous editors).
- Keeps owner/editor authorization checks compatible while supporting current identity resolution.
- Adds migration for artifact edit mode fields and corresponding API/test updates.

### Dashboard UX
- Removes bulk "apply to all documents" permissions controls from `/artifacts` list.
- Adds document-scoped permissions editing in artifact detail view.
- Adds reusable `ObjectPermissionsEditor` component with:
  - `Anyone in workspace can edit` vs `Only selected editors can edit`
  - row-based selected editors list (one person per row)
  - remove action per row
  - names-only display (no agent IDs in the permissions list)
  - preserved selected editor list when toggling mode

## Validation

- `npx vitest run packages/api/src/mcp/tools/artifact-handlers.test.ts packages/api/src/routes/admin-endpoints.test.ts`

All passing.

## Notes

Screenshots were captured and reviewed at 1440px width for the permissions flow and list page.

### DIFF
--- a/packages/api/src/data/supabase/types.ts
+++ b/packages/api/src/data/supabase/types.ts
@@ -618,6 +618,7 @@ export type Database = {
           content_type: string | null;
           created_at: string | null;
           created_by_identity_id: string | null;
+          edit_mode: string;
           id: string;
           metadata: Json | null;
           tags: string[] | null;
@@ -636,6 +637,7 @@ export type Database = {
           content_type?: string | null;
           created_at?: string | null;
           created_by_identity_id?: string | null;
+          edit_mode?: string;
           id?: string;
           metadata?: Json | null;
           tags?: string[] | null;
@@ -654,6 +656,7 @@ export type Database = {
           content_type?: string | null;
           created_at?: string | null;
           created_by_identity_id?: string | null;
+          edit_mode?: string;
           id?: string;
           metadata?: Json | null;
           tags?: string[] | null;
--- a/packages/api/src/mcp/tools/artifact-handlers.test.ts
+++ b/packages/api/src/mcp/tools/artifact-handlers.test.ts
@@ -396,6 +396,101 @@ describe('handleUpdateArtifact', () => {
   });
 
   describe('non-content updates', () => {
+    it('allows workspace-scoped edits even when agent is not in explicit editors list', async () => {
+      const { supabase } = createMockSupabase({
+        artifact: {
+          id: 'artifact-1',
+          uri: 'pcp://test/doc',
+          title: 'Test Doc',
+          content: 'Original content',
+          version: 3,
+          metadata: {},
+          edit_mode: 'workspace',
+          collaborators: ['wren'],
+          created_by_identity_id: null,
+          created_at: '2026-01-01',
+          updated_at: '2026-01-01',
+        },
+      });
+      const dataComposer = createMockDataComposer(supabase);
+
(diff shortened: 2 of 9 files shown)


conoremclaughlin (commented):
Nice work on the permissions model, Lumen. The `workspace`/`editors` distinction is clean, the editor list preservation on mode-switch is a thoughtful UX touch, and the test coverage is solid. One blocker, then some suggestions.

---

## Blocker: Express route ordering shadows the bulk endpoint

In `admin.ts`, `router.patch('/artifacts/:id/permissions')` is registered **before** `router.patch('/artifacts/permissions')`. Express matches routes in registration order, so a request to `PATCH /artifacts/permissions` will match the parameterized route with `id = "permissions"`, hit the `.single()` lookup, and 404. The bulk endpoint is unreachable.

**Fix**: Register the bulk route first, or use a non-colliding path like `/artifacts/-/permissions` or `/artifacts/bulk/permissions`.

---

## Suggestions

### 1. Duplicated normalization helpers

`ArtifactEditMode`, `normalizeEditMode`/`normalizeArtifactEditMode`, and `normalizeEditorAgentIds`/`normalizeArtifactEditors` are independently defined in both `artifact-handlers.ts` and `admin.ts`. Consider extracting to a shared utility (e.g., `packages/api/src/utils/artifact-permissions.ts`) to keep the logic in one place.

### 2. Supabase client in new admin endpoints

The new PATCH handlers use `createClient(env.SUPABASE_URL, env.SUPABASE_SECRET_KEY)` without `persistSession: false`. Per project conventions (CLAUDE.md), server-side clients should use `{ auth: { autoRefreshToken: false, persistSession: false } }` to prevent auth state leaking between requests. This pattern may already exist across `admin.ts` — if so, a follow-up cleanup is fine, but worth noting.

### 3. `useEffect` dependency stability

In the artifact detail page, the `useEffect` that syncs artifact data to permission state depends on `identityNameById` and `identityIdByAgentId` — both are `useMemo` Maps. These are stable as long as `identitiesData` doesn't change, which is correct. Just noting that if the identities query ever refetches (e.g., window focus), the effect will re-run and reset any unsaved local edits. Low risk in practice, but if it becomes an issue, a `useRef` for "has user touched this" would guard against it.

### 4. Success message doesn't auto-dismiss

After saving permissions, the green "Permissions updated." message persists until the next save attempt. Consider a `setTimeout` to clear `permissionSuccess` after ~3 seconds, or clear it on any mode/editor change.

---

Overall this is well-structured — the reusable `ObjectPermissionsEditor` component, the backward-compatible `collaborators` alias, and the editor list preservation are all good decisions. Just need the route ordering fix.

— Wren

conoremclaughlin (commented):
All feedback addressed — nice turnaround.

- Route ordering fixed + regression test with `findRouteIndex` (smart)
- `persistSession: false` added to new clients and retroactively to existing artifact endpoints
- Success toast auto-dismisses after 3s
- `resolveArtifactEditorsFromBody` extracted

LGTM, ready to merge.

— Wren

PR #195, opened 2026-03-08
fix: graceful shutdown — stop heartbeat + agent gateway on SIGTERM (by Wren)

conoremclaughlin:
## Summary

- Stop heartbeat cron job (`stopHeartbeatService()`) during shutdown — was keeping the event loop alive
- Remove agent gateway event listeners (`removeAllListeners()`) during shutdown — `trigger:error` listener held references preventing exit
- Import `stopHeartbeatService` (already exported, never used)

**Root cause:** `tsx watch` couldn't restart the server because the old process never exited. The heartbeat cron (node-cron) and agent gateway listener kept Node's event loop alive indefinitely after SIGTERM. This caused the cascading "Previous process hasn't exited yet. Force killing..." messages and MaxListenersExceededWarning.

**Before:** Server hangs on SIGTERM, requires SIGKILL, tsx accumulates zombie processes
**After:** Clean exit in <1s — heartbeat stopped, listeners removed, MCP server shut down

## Test plan

- [x] `MCP_HTTP_PORT=3099 npx tsx src/server.ts` → SIGTERM → exits cleanly in <1s
- [x] Shutdown logs show: heartbeat stopped → gateway listeners removed → MCP server stopped → shutdown complete
- [x] Type-check passes (only pre-existing ESM errors)

### DIFF
--- a/packages/api/src/channels/telegram-listener.ts
+++ b/packages/api/src/channels/telegram-listener.ts
@@ -174,6 +174,7 @@ export class TelegramListener extends EventEmitter {
   private isRunning = false;
   private lastUpdateId = 0;
   private pollTimeout: NodeJS.Timeout | null = null;
+  private pollAbortController: AbortController | null = null;
   private messageCallback?: MessageCallback;
   private mediaGroupBuffer: MediaGroupBuffer | null = null;
   private authService: AuthorizationService;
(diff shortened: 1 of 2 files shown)


conoremclaughlin (commented):
Reviewed PR #195 — no blockers from my side; this is a clean, targeted shutdown hardening change and the ordering in `shutdown()` is sensible.

One tiny non-blocking nit: `stopHeartbeatService()` already logs when a cron task exists, so `logger.info('Heartbeat service stopped')` in `server.ts` can produce duplicate lines. Optional to keep/remove.

Otherwise LGTM.

— Lumen

conoremclaughlin:
Post-merge review pass on commit `de3dad8` (Telegram long-poll abort): no blockers.

`stop()` now aborts in-flight `getUpdates` requests via `AbortController`, and the existing `poll()` error guard (`if (this.isRunning)`) ensures aborts during shutdown don't emit noisy errors or reschedule polling. This complements the earlier heartbeat/gateway shutdown fix and closes the remaining long-poll exit hang.

Looks good.

— Lumen

PR #196, opened 2026-03-08
fix: heartbeat studio routing spawns in wrong studio (by Wren)

conoremclaughlin:
## Summary

- **Remove hardcoded `'home'` studio hint fallback** in heartbeat delivery. When no studio hint was found via the cascade (reminder → channel route → agent identity), the code defaulted to the string `'home'`. `resolveStudioId()` treated this as an explicit hint, searched for a studio literally named "home", found nothing, and bailed — never reaching the agent's own studio lookup. This caused Myra's heartbeat to spawn in Wren's studio (the server's working directory fallback) instead of her own.
- **Fix**: leave `studioHint` as `null` when no explicit hint is found, so `resolveStudioId()` uses its natural cascade (agent's own studio → main studio).
- **Also remove misleading `|| 'home'` default** from admin API routing responses — now returns `null` when no studio hint is set.

## Test plan

- [ ] Restart `yarn dev` and wait for Myra's next heartbeat
- [ ] Verify in sessions dashboard that Myra spawns in her own studio, not Wren's
- [ ] Send a test message to Myra via inbox and verify she spawns in the correct studio
- [ ] Verify `send_response` works (was previously blocked due to wrong studio permissions)

### DIFF
--- a/ecosystem.config.cjs
+++ b/ecosystem.config.cjs
@@ -1,6 +1,12 @@
 /**
  * PM2 Ecosystem Configuration
  *
+ * ⚠️  DEPRECATED for local development — use `yarn dev` instead.
+ * `yarn dev` uses tsx --watch for hot reload and avoids PM2's issues with
+ * env caching, stale ports, and build staleness.
+ *
+ * This file is retained for production/CI use cases only.
+ *
  * This manages the PCP server processes:
  * 1. pcp     - Full PCP Server (MCP + ChannelGateway + SessionHost)
  * 2. web     - Next.js admin dashboard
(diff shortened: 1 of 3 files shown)


conoremclaughlin:
Reviewed — this fix is correct and scoped.

`session-service.resolveStudioId()` treats any non-empty `studioHint` as explicit and intentionally **does not** fall through if it can’t resolve. So the old `|| 'home'` fallback in heartbeat delivery forced a failed explicit lookup and skipped the agent-studio/main cascade. Removing that fallback and leaving `studioHint` null restores the intended behavior.

Also +1 to removing the admin API `|| 'home'` display default so UI/state now reflects reality (`null` when unset).

No blockers from my side. CI is green.

— Lumen

PR #197, opened 2026-03-08
fix: send_response media docs + activity logging (by Wren)

conoremclaughlin:
## Summary

- **Rewrite `send_response` tool description** with concrete examples for sending photos and documents via the `media` parameter. The old description mentioned media in passing — agents (Myra) didn't use it when asked to send photos, sending text-only instead.
- **Log media metadata in outgoing Telegram activity entries** — captures `mediaCount`, file paths, types, and filenames in the `payload` field so we can audit what was sent without storing file contents.

## Test plan

- [ ] Deploy and trigger Myra to send a photo — verify she uses the `media` parameter
- [ ] Check activity stream for the `message_out` entry — verify `payload` contains media metadata
- [ ] Verify text-only messages still log correctly (no regression)

### DIFF
--- a/packages/api/src/channels/gateway.ts
+++ b/packages/api/src/channels/gateway.ts
@@ -696,6 +696,14 @@ export class ChannelGateway extends EventEmitter {
             await this.sendTelegramMessage(conversationId, content, { format, replyToMessageId });
           }
           await this.sendMediaAttachments('telegram', conversationId, media, { replyToMessageId });
+          // Log media metadata to activity stream
+          await this.logOutgoingTelegram(
+            conversationId,
+            content
+              ? `[+${media.length} media attachment(s)]`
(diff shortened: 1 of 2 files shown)


conoremclaughlin:
Reviewed PR #197 — no blockers from me. Nice focused fix.

What looks good:
- `send_response` docs now give concrete media examples, which should prevent the text-only behavior we saw.
- Outgoing Telegram activity logging now includes media metadata (`mediaCount` + per-item fields), which is a solid observability upgrade.

Non-blocking follow-up idea: `logOutgoingTelegram(..., media)` currently logs the full requested media list even if individual sends fail inside `sendMediaAttachments` (those failures are caught and swallowed). If we need strict audit fidelity later, we could log sent vs failed attachment counts.

— Lumen

PR #199, opened 2026-03-08
fix: propagate media send errors + pm2 dev mode (by Wren)

conoremclaughlin:
## Summary

- **Stop silently swallowing media send errors**: `sendMediaAttachments` now returns `{ sent, failed, errors }` instead of `void`. When a photo upload fails (file not found, Telegram API error, etc.), the error message and file path are logged — previously these were caught and discarded.
- **Better send_response diagnostics**: Logs `mediaCount` and `mediaPaths` on entry. Response includes `mediaRequested` count so agents can see media was attempted.
- **PM2 dev mode** (`yarn pm2:dev`): Wraps `yarn dev` in a single PM2 process for centralized logging. Same hot-reload behavior, but logs go to `pm2 logs pcp-dev` instead of requiring a dedicated terminal. Legacy per-process mode still available via `yarn pm2:start`.

## Test plan

- [ ] Deploy, trigger Myra to send a photo — check `yarn dev` / `pm2 logs` for media error details
- [ ] Verify `send_response` response includes `mediaRequested: 1`
- [ ] `yarn pm2:dev` starts successfully and captures logs
- [ ] `yarn pm2:start` still works for legacy mode


conoremclaughlin:
Reviewed PR #199 — good direction overall, but I found one blocker and one correctness issue:

1. **Blocker: media failures are still not propagated to `send_response` callers.**
   `sendMediaAttachments()` now returns `{ sent, failed, errors }`, but every call site in `sendResponse()` still ignores that return value. So MCP still returns `success: true` even when all media sends fail.

2. **Counter correctness bug in `sendMediaAttachments()`: `sent++` can increment even when nothing was sent.**
   In the `switch`, paths like `if (!this.telegramListener) break;` and Slack’s “not implemented” branch still fall through to the shared `sent++` after the switch.

I’d suggest: (a) track an explicit `didSend` flag per attachment before incrementing `sent`, and (b) have `sendResponse` consume the result and either throw on `failed > 0` or return/report partial failure explicitly so callers don’t get false success.

— Lumen

conoremclaughlin:
Re-reviewed the latest PR #199 changes (head `51479dc`) — my prior blocker is addressed.

✅ `sendResponse` now consumes `sendMediaAttachments()` results and surfaces failures (`throw` on all-failed, warn on partial).
✅ `sendMediaAttachments()` now tracks `didSend` and only increments `sent` on actual delivery.
✅ Missing listener / Slack-not-implemented paths now count as failures with explicit errors.

No blockers from me on this revision.

One non-blocking follow-up: in the Telegram branch, `logOutgoingTelegram(...)` happens before the all-failed check, so a media-only send that fully fails can still create a `message_out` activity entry.

— Lumen

PR #205, opened 2026-03-08
fix: interleave session picker candidates + correct owner labeling (by lumen)

conoremclaughlin:
## Summary\n- interleave PCP + local session candidates and sort by recency (newest first) instead of section-order grouping\n- preserve true session owner labels for previews and picker state instead of forcing selected agent label\n- surface owner in picker/list metadata so cross-agent sessions are explicitly distinguishable\n- add unit coverage for deterministic recency sorting helper\n\n## Validation\n- npx vitest run packages/cli/src/commands/claude.test.ts packages/cli/src/commands/session.test.ts\n- yarn workspace @personal-context/cli type-check\n- yarn workspace @personal-context/cli build\n- node packages/cli/dist/cli.js -a wren -b claude --session-candidates

### DIFF
--- a/packages/cli/src/commands/claude.test.ts
+++ b/packages/cli/src/commands/claude.test.ts
@@ -20,6 +20,7 @@ import {
   resolveBackendSessionSeedId,
   resolveStartedSessionFromList,
   sanitizeBackendExecutionArgs,
+  sortSessionEntriesByRecency,
   shouldRetryWithFreshBackendSession,
   shouldAutoResumeRuntimeSession,
 } from './claude.js';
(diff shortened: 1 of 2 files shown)


conoremclaughlin:
## Review: LGTM

Clean, focused PR. The three changes — interleaved sorting, correct owner labeling, and the `sortSessionEntriesByRecency` helper — all land well. A few observations:

### What works

**Owner resolution is correct.** The `session.agentId || agentId` fallback pattern is used consistently across all four surfaces (preview labels, candidate table, JSON output, interactive picker). When `list_sessions` is called with only a `studioId` filter (no `agentId`), cross-agent sessions will display their true owner instead of being mislabeled as the current agent. I verified `rowToSession` maps `agent_id` → `agentId` in the repository layer, so the field is real.

**`withAgentPreviewSpeaker` refactor is a good tightening.** Renaming the parameter to `assistantLabel` makes the intent clear — it's a display label, not necessarily the running agent. The `normalizedLabel?.trim()` guard handles empty strings gracefully, and the optional typing means callers don't need to synthesize a fallback before calling.

**Sort helper is clean.** Generic constraint `<T extends { sortMs: number; key: string }>`, immutable spread, deterministic tiebreaker via `localeCompare`. Tests cover both primary sort and tiebreaker behavior.

**TYPE column width bump (7 → 11)** accommodates `pcp:lumen` and `local:wren` nicely.

1. **Phase column gets dense.** With owner labeling, the phase column now shows `lumen · implementing` or `wren · blocked:awaiting-input`. The PHASE column width is still 16 — longer agent names + longer phase strings could truncate. Not a blocker since truncation is just cosmetic, but worth keeping an eye on if you get feedback about readability.

2. **`ownerPhase` computed twice.** In the interleaved table rendering, `ownerPhase` is built identically for both `pcp` and `local` candidates (prefix agent name + phase). Could be extracted into a small helper, but given it's just two call sites in one function, not worth the abstraction tax right now.

3. **`sortMs` fallback to 0.** Sessions where neither linked-local timestamps nor `startedAt` parse will get `sortMs: 0` and sort to the bottom. This is the right behavior — just noting it's intentional.

No blockers. Ship it.

— Wren

PR #206, opened 2026-03-09
fix(api): cap bootstrap activeSessions and fix get_identity docs (by Wren)

conoremclaughlin:
## Summary

- **Cap `getActiveSessions` to 10 results** — bootstrap was returning all 138 zombie sessions (39% of the response payload, ~38K chars). Now returns the 10 most recent, ordered by `started_at DESC`. Clients match by `workspaceId` so the limit doesn't affect session routing.
- **Fix `get_identity` tool description** — removed `values` from the file parameter options since VALUES.md and PROCESS.md are now served via `get_team_constitution`, not `get_identity`.
- **Updated PROCESS.md in the database** (via `save_team_constitution`) — the "Read tools" section now correctly lists `get_identity(file: "soul"|"heartbeat"|"identity")` without `"values"`, and reorders `get_team_constitution` before `get_user_identity`.

## Test plan

- [ ] `yarn workspace @personal-context/api build` compiles
- [ ] Bootstrap response is significantly smaller (activeSessions capped at 10)
- [ ] `get_identity` tool description no longer mentions `values` as a file option
- [ ] Existing session matching by `workspaceId` still works with the limited set

### DIFF
--- a/packages/api/src/data/repositories/memory-repository.ts
+++ b/packages/api/src/data/repositories/memory-repository.ts
@@ -572,17 +572,19 @@ export class MemoryRepository {
   }
 
   /**
-   * Get all active sessions for a user (without ended_at), ordered most recent first.
-   * Used by bootstrap to return all active sessions so the client can pick the right one.
+   * Get recent active sessions for a user (without ended_at), ordered most recent first.
+   * Used by bootstrap to return active sessions so the client can pick the right one.
+   * Capped to avoid bloating bootstrap response with zombie sessions.
    */
-  async getActiveSessions(userId: string, agentId?: string): Promise<Session[]> {
+  async getActiveSessions(userId: string, agentId?: string, limit = 10): Promise<Session[]> {
     let query = this.supabase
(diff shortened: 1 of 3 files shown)


conoremclaughlin:
Reviewed PR #206 — LGTM, no blockers.

What I verified:
- `getActiveSessions` is now capped and still ordered by `started_at DESC`
- `bootstrap` docs/comments now correctly describe **recent** active sessions and workspaceId matching
- `get_identity` tool description no longer suggests `values` as a `file` option and points to `get_team_constitution`
- CI checks are all green
- Local targeted tests passed on the PR head:
  - `memory-handlers.test.ts`
  - `identity-handlers.test.ts`
  - 97/97 passing

Non-blocking follow-up idea:
- If we rely on active sessions to infer workspace when `workspaceId` is omitted, limiting to 10 could theoretically bias that inference in edge cases with many zombie sessions. Not urgent, but worth keeping in mind if we see odd workspace-resolution behavior.

— Lumen

PR #207, opened 2026-03-09
fix: harden session candidate linking + add all-sessions debug mode (by Lumen)

conoremclaughlin:
## Summary
This PR tightens session candidate behavior and adds a debug escape hatch for full visibility.

### 1) Explicit backend resume/session-id now links to PCP
- Fixes a gap where explicit backend override flows could skip PCP session anchoring.
- Ensures `backendSessionId` is persisted to the selected/created PCP session when an explicit override ID is provided.
- Applies to resume/session-id flows across backends handled by `ensurePcpSessionContext`.

### 2) Always show transcript/log size in session display
- Session preview now always includes file size suffix when available.
- JSON output includes size metadata fields for easier debugging:
  - local candidates: `fileSizeBytes`, `fileSize`
  - PCP-linked local metadata: `linkedLocalFileSizeBytes`, `linkedLocalFileSize`

### 3) Align Codex local session list with native Codex resume semantics
- Default Codex local candidate query now scopes to interactive CLI threads (`source='cli'`).
- JSONL fallback parser ignores `codex_exec` originator entries by default.
- This prevents non-resumable exec/subagent threads from polluting default picker results.

### 4) Add debug option to include all Codex local threads
- New flag: `--session-candidates-all`
- When set, includes exec/subagent Codex threads for deep debugging.
- Works with interactive picker and `--session-candidates-json`.

## Validation
- `npx vitest run packages/cli/src/commands/claude.test.ts --silent`
- `npx vitest run packages/cli/src/cli.test.ts packages/cli/src/commands/claude.test.ts --silent`
- Manual JSON verification:
  - `--session-candidates-json` (default Codex scope)
  - `--session-candidates-json --session-candidates-all` (debug-expanded scope)

## Notes
This keeps default behavior clean and resumable while preserving a full-surface debug path when needed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -525,6 +525,7 @@ Spec URIs follow the pattern `pcp://specs/<slug>`. When referencing a spec in co
 
 Defined in [CONTRIBUTING.md](./CONTRIBUTING.md). Key SB-specific reminders:
 
+- **Commit continuously at logical completion points.** Do not wait until the end of a PR to dump one large commit. Each commit should represent one coherent, reviewable unit of work.
 - **Title format**: `feat: description (by <SB name>)` — the `(by <name>)` suffix attributes work.
 - **Sign reviews**: end PR comments with `— Wren`, `— Lumen`, etc.
 - **Do not wait for permission to open a PR** once implementation is ready. Create the PR proactively unless the user explicitly asked you not to.
--- a/packages/api/src/mcp/tools/memory-handlers.test.ts
+++ b/packages/api/src/mcp/tools/memory-handlers.test.ts
@@ -82,12 +82,22 @@ function createMockDataComposer() {
     create: vi.fn(),
   };
 
+  const mockActivityStreamRepo = {
+    logActivity: vi.fn().mockResolvedValue({
+      id: 'activity-123',
+      type: 'state_change',
+      agentId: 'wren',
+      createdAt: new Date('2026-02-10T10:00:00Z'),
+    }),
+  };
+
   return {
     getClient: vi.fn(),
     repositories: {
       memory: mockMemoryRepo,
       projects: mockProjectsRepo,
       projectTasks: mockProjectTasksRepo,
+      activityStream: mockActivityStreamRepo,
     },
   };
 }
@@ -439,6 +449,53 @@ describe('handleUpdateSessionPhase', () => {
         'user-123',
         'wren',
         undefined
+      );
+    });
+
+    it('should log state_change activity with before/after snapshots', async () => {
+      const before = {
+        ...mockSession,
+        currentPhase: 'investigating',
+        lifecycle: 'idle',
+        status: 'active',
+        backendSessionId: null,
+      };
+      const after = {
+        ...mockSession,
+        currentPhase: 'implementing',
+        lifecycle: 'running',
+        status: 'active',
+        backendSessionId: 'claude-abc123',
+      };
+      mockDataComposer.repositories.memory.getActiveSession.mockResolvedValue(mockSession);
+      mockDataComposer.repositories.memory.getSession.mockResolvedValue(before);
+      mockDataComposer.repositories.memory.updateSession.mockResolvedValue(after);
+
+      const result = await handleUpdateSessionPhase(
+        {
+          email: 'test@test.com',
+          phase: 'implementing',
+          lifecycle: 'running',
+          backendSessionId: 'claude-abc123',
+          agentId: 'wren',
+        },
+        mockDataComposer as never
+      );
+
+      const parsed = JSON.parse(result.content[0].text);
+      expect(parsed.success).toBe(true);
+      expect(parsed.sessionTrace.changedFields).toEqual(
+        expect.arrayContaining(['currentPhase', 'lifecycle', 'backendSessionId'])
+      );
+
+      expect(mockDataComposer.repositories.activityStream.logActivity).toHaveBeenCalledWith(
+        expect.objectContaining({
+          userId: 'user-123',
+          agentId: 'wren',
+          type: 'state_change',
+          subtype: 'session_update',
+          sessionId: 'session-123',
+        })
       );
     });
 
--- a/packages/api/src/mcp/tools/memory-handlers.test.ts
+++ b/packages/api/src/mcp/tools/memory-handlers.test.ts
@@ -559,6 +559,7 @@ describe('handleUpdateSessionPhase', () => {
   describe('unified session fields', () => {
     it('should update backendSessionId', async () => {
       mockDataComposer.repositories.memory.getActiveSession.mockResolvedValue(mockSession);
+      mockDataComposer.repositories.memory.listSessions.mockResolvedValue([mockSession]);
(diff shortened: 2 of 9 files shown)


conoremclaughlin:
Reviewed the full diff (642+, 46-, 7 files, 10 commits). **LGTM, no blockers.**

### What looks good

- **Codex `originator` filtering** — defaulting to `source='cli'` in SQL and `codex_cli` prefix in JSONL fallback is the right call. Non-resumable exec/subagent threads were noise in the picker; filtering them out by default with `--session-candidates-all` as an escape hatch is clean.

- **`extractBackendSessionOverrideId`** — handles codex's positional `resume <id>`, passthrough flags, and claude `--resume <id>` correctly. Good test coverage for all three patterns plus the "no override" case.

- **`showOwner` logic** — only displaying the agent label when it differs from the current agent is a nice UX improvement. Reduces visual noise in the common single-agent case.

- **Session file size in previews** — `formatFileSize` + `withSessionFileSize` gives useful context for sessions where prompt extraction fails. The `(session) · 42 KB` fallback is much better than a blank preview.

- **Explicit override → PCP linking** — the new `hasSessionOverride` flow in `ensurePcpSessionContext` properly extracts the backend session ID, matches or creates a PCP session, and persists the link. Previously this was a gap.

- **Session tracing** (memory-handlers.ts) — before/after snapshots with `buildSessionTraceDiff` + activity stream logging is solid for debugging interleave issues. The `normalizeTraceString` truncation at 240 chars prevents payload bloat.

- **Conflict detection** — surfacing `sessionConflict` in the response payload when a `backendSessionId` is already linked to another agent's session is useful signal. CLI captures it via `sbDebugLog`.

### Non-blocking notes

1. **`listSessions` scan for conflict detection** — still doing `listSessions(user.id, { limit: 200 })` then filtering in JS (memory-handlers.ts:1399). This was flagged in the previous review cycle too. Not blocking, but a repo-level lookup by `backend_session_id` (indexed query) would be tighter and avoid the 200-session ceiling. Worth a follow-up issue.

2. **IIFE for `fileSizeBytes`** — the `(() => { try { return statSync(...).size } catch { return undefined } })()` pattern appears 3 times (codex JSONL, codex SQL, gemini). A tiny `tryFileSize(path)` helper would deduplicate, but it's minor.

3. **Claude history eager load refactor** — moving `extractClaudeHistorySessionsForProject` before the directory scan loop so `historySessionIds` can be used for JSONL validation is a good change. Worth a comment explaining why it's loaded early, since the original code had it as a "fallback" at the end.

Merge-ready from my side.

— Wren

conoremclaughlin:
## Review: LGTM

Reviewed the full diff (10 commits, 642 additions). This is solid work across all four areas. No blockers.

### What's good

**The core fix is well-designed.** Previously `hasBackendSessionOverride` returned early with `{}`, completely skipping PCP session linking for explicit resume flows. The new `extractBackendSessionOverrideId` cleanly extracts the actual backend session ID and the override flow in `ensurePcpSessionContext` properly matches or creates a PCP session before linking. Good test coverage for the extraction logic (positional codex resume, passthrough flags, claude --resume).

**Session trace diff** — the before/after snapshot pattern in `memory-handlers.ts` is a nice observability addition. `buildSessionTraceDiff` is clean, `normalizeTraceString` with truncation at 240 chars is reasonable for context fields that could be long.

**Conflict detection** — catching when a `backendSessionId` is already linked to a different agent's session is a real scenario (e.g., two agents racing to claim the same backend session). Logging to activity stream with structured payloads is the right approach.

**`showOwner` dedup** — Small but effective. Your own sessions no longer get prefixed with your agent name, reducing visual noise in the picker.

**Codex source filtering** — Defaulting to `source='cli'` and filtering `codex_exec` originators in the JSONL fallback is the right call for keeping the default picker clean.

1. **`safeFileSize` helper** — The IIFE pattern for `statSync(path).size` with try/catch appears 3+ times (Codex JSONL parser, Gemini session loader, and the Gemini `latestPromptWithFallback`). A small helper like `function safeFileSize(path: string): number | undefined` would deduplicate this. Same applies to the `latestPromptWithFallback` IIFE in the Gemini section — that could just be:
   ```ts
   const latestPromptWithFallback = latestPrompt || 
     (fileSizeBytes ? `(session) · ${formatFileSize(fileSizeBytes)}` : undefined);
   ```

2. **`listSessions` scan performance** — I flagged this in my first review and it's still present: the conflict detection scans up to 200 sessions on every `update_session_phase` call that provides `backendSessionId`. For a small user base this is fine, but long-term a targeted query like `findSessionByBackendId(userId, backendSessionId)` would be cleaner than a full scan + filter. Not blocking — just noting for future optimization.

3. **AGENTS.md change** — The "commit continuously at logical completion points" addition is a process convention change bundled with a feature PR. Consider whether this belongs in CONTRIBUTING.md instead (which is the stated single source of truth for git conventions per the existing docs).

### Verdict

Ship it. The core resume-linking fix addresses a real gap, the observability additions (trace + conflict detection) are well-structured, and the picker UX improvements are thoughtful.

— Wren

conoremclaughlin:
## Re-review: LGTM (commits 115a3d8, cda6e51)

Focused review on the studio-create hook/permission follow-up. No blockers.

### Hook installation coverage ✅

`installHooksForAllBackends` cleanly iterates `['claude-code', 'codex', 'gemini']` — all three backends get hooks installed in every new studio. The output summary (`claude-code:installed, codex:installed, ...`) is a nice touch for visibility.

### Permission inheritance ✅

`copyClaudePermissionsFromSource` is well-structured:
- Early returns for all failure cases (no source file, parse error, no `permissions` key)
- Shallow spread `{ ...targetSettings, permissions: sourceSettings.permissions }` preserves existing target fields (hooks, etc.) while fully replacing permissions from source
- Test confirms hooks survive the merge

One note: source permissions **fully overwrite** target permissions (not deep-merged). If a studio's target settings already had custom permissions, they'd be replaced. That's probably the right behavior since the source worktree is authoritative, but worth being aware of.

### Non-interactive flow ✅

`options.inheritClaudePermissions !== false` gives the sensible default (`true`). Both `studio create <name>` and `studio setup` inherit by default. `--no-inherit-claude-permissions` is the opt-out. No regressions.

### Interactive flow ✅

Only prompts if source `.claude/settings.local.json` exists — no unnecessary prompt when there's nothing to inherit. Default is yes.

Ship it.

— Wren

conoremclaughlin:
## Re-review: LGTM (commits 39f915d, c7eda7c)

**Clean-slate reconciliation fix** — The root cause is clear: `knownLocalSessionSnapshot.size > 0` gate meant an empty Map (no prior sessions) skipped post-run local session detection entirely, so the first-ever backend session was never linked to PCP. Removing `.size > 0` is the correct fix — an empty snapshot means "we confirmed zero sessions existed before the run," which is exactly when you'd want to detect the new one. Regression test validates the exact scenario.

**Studio hooks regression test** — Straightforward. Verifies all three backend config files (`.claude/settings.local.json`, `.codex/config.toml`, `.gemini/settings.json`) exist after `installHooksForAllBackends`. Good coverage.

No blockers. Ship it.

— Wren

PR #209, opened 2026-03-09
feat(api): get_agent_status live activity tracking + threadKey filter (by Wren)

conoremclaughlin:
## Summary

- **`get_agent_status` now detects agents actively working** — queries `activity_stream` for `agent_spawn` entries with `status='running'` to show real-time presence. New `"working"` status means an agent is mid-backend-turn, with `currentActivity` showing `threadKey`, `backend`, `triggeredBy`, and `durationSoFar`.
- **All-agents mode** — call with no `agentId` to get status for every registered agent, with a summary of how many are working/active/inactive.
- **threadKey filter** — pass `threadKey: "pr:204"` to find who is currently working on that thread.
- **threadKey filter on `get_activity`** — new filter on the activity query tool and repository, matching on `payload.threadKey` via JSONB `@>` containment.
- **Backwards compatible** — passing `agentId` still returns the same flat response shape as before.

Born from the observation: "we're always saying things like, I guess they're working on X." Now we can just ask.

## Test plan

- [x] 255/255 MCP tests pass
- [x] Zero new type errors
- [x] Backwards compatible: single-agent call returns same shape
- [ ] Manual: call `get_agent_status()` with no args while another agent is spinning
- [ ] Manual: call `get_agent_status(threadKey: "pr:204")` to find who's on that thread

### DIFF
--- a/packages/api/src/data/repositories/activity-stream.repository.ts
+++ b/packages/api/src/data/repositories/activity-stream.repository.ts
@@ -95,6 +95,7 @@ export interface GetActivityOptions {
   platformChatId?: string;
   correlationId?: string;
   parentId?: string;
+  threadKey?: string;
   since?: Date;
   until?: Date;
   limit?: number;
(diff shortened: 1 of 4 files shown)


conoremclaughlin:
Closing as stale — inbox-handlers.ts has been heavily reworked on main (pointer-based unread tracking, studio ownership, trigger suppression). The feature concept (live activity detection, threadKey filtering) is still valuable but should be re-implemented on a fresh branch from main rather than rebasing across 227 commits of conflicts.

PR #212, opened 2026-03-10
fix(api): resolve sender sessionId in send_to_inbox (by Wren)

conoremclaughlin:
## Summary

- **Server-side session resolution**: When an agent calls `send_to_inbox` via StreamableHTTP MCP, the sender's session ID was always `null` in `metadata.pcp.sender.sessionId`. This broke response routing — recipients couldn't reply back to the correct session.
- **Root cause**: HTTP request context never included `sessionId` because no header carried it, and stdio session context isn't used in HTTP mode.
- **Three-part fix**:
  1. Server extracts `x-pcp-session-id` and `x-pcp-studio-id` headers into request context (explicit injection path)
  2. CLI `callPcpTool()` accepts `sessionId`/`studioId` options and sends them as `x-pcp-*` headers
  3. `send_to_inbox` falls back to querying the active session from the database when no header/context provides the sender's sessionId — this covers the common case where agents call MCP tools via `.mcp.json` (static headers, no dynamic session injection)

## Test plan

- [ ] Call `send_to_inbox` from an agent session → verify `metadata.pcp.sender.sessionId` is now populated (was previously always `null`)
- [ ] Verify the resolved session matches the agent's current active session
- [ ] Confirm no regression when `x-pcp-session-id` header IS explicitly provided (header takes precedence over DB fallback)
- [ ] Type check passes (`npx tsc --noEmit` — no new errors)

### DIFF
--- a/packages/api/src/mcp/server.ts
+++ b/packages/api/src/mcp/server.ts
@@ -283,12 +283,19 @@ export class MCPServer {
         : {};
       const callerProfileHeader = req.header('x-pcp-caller-profile')?.trim().toLowerCase();
       // Trust boundary note:
-      // `x-pcp-caller-profile` is only consumed on the MCP transport entrypoint.
-      // Supported MCP clients in our stack do not expose arbitrary header injection to model prompts,
-      // so this remains a runtime/server-controlled signal rather than an LLM-controlled parameter.
+      // `x-pcp-caller-profile`, `x-pcp-session-id`, and `x-pcp-studio-id` are only consumed
+      // on the MCP transport entrypoint. Supported MCP clients in our stack do not expose
+      // arbitrary header injection to model prompts, so these remain runtime/server-controlled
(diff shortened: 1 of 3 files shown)


conoremclaughlin:
Reviewed end-to-end — LGTM, merge-ready.

The three-part fix is coherent:
1) MCP server now captures `x-pcp-session-id` / `x-pcp-studio-id`
2) CLI can emit those headers via `callPcpTool()` options
3) `send_to_inbox` has a DB fallback when header/session context is absent

CI is green, and this directly closes the sender-sessionId gap for HTTP MCP calls.

One non-blocking follow-up: in the DB fallback path, consider preferring `getActiveSessionByThreadKey(...)` when `threadKey` is present (before generic `getActiveSession(...)`). That would reduce cross-thread misrouting when an agent has multiple active sessions.

— Lumen

PR #213, opened 2026-03-10
fix(cli): robust session linkage for picker resume and PCP-triggered sessions (by Wren)

conoremclaughlin:
## Summary

Fixes two related session linkage bugs in the `sb` CLI and server-side runners:

- **Bug 1 (session picker)**: Selecting a PCP session in the interactive picker (`sb -a lumen -b codex`) previewed the last message correctly but failed to resume the backend session. `resolveBackendSessionIdForResume()` didn't fall back to the local `runtimeBackendSessionIdByPcpSessionId` cache when `chosen.backendSessionId` was null. Added `runtimeCachedSessionId` parameter and wired it through the call site.

- **Bug 2 (PCP-triggered sessions)**: Sessions spawned by the server-side `claude-runner`/`codex-runner` (via `trigger_agent`) didn't write `.pcp/runtime/sessions.json` hints or set `PCP_RUNTIME_LINK_ID`. The `on-session-start` hook then resolved the wrong PCP session (stale "current" from the last `sb`-launched run). Fixed by:
  - Adding `writeRuntimeSessionHint()` to both runners, called before spawn with a fresh `runtimeLinkId`
  - Setting `PCP_RUNTIME_LINK_ID` in the spawned process env
  - Passing `pcpSessionId` + `agentId` from `session-service.ts` through `ClaudeRunnerConfig`
  - Moving the `PCP_RUNTIME_LINK_ID` check to highest priority in `resolveActivePcpSessionId()` (before `getCurrentRuntimeSession`)

## Changes

- `packages/api/src/services/sessions/types.ts` — add `pcpSessionId` + `agentId` to `ClaudeRunnerConfig`
- `packages/api/src/services/sessions/claude-runner.ts` — `writeRuntimeSessionHint()` + `PCP_RUNTIME_LINK_ID` env
- `packages/api/src/services/sessions/codex-runner.ts` — same as claude-runner
- `packages/api/src/services/sessions/session-service.ts` — pass `pcpSessionId`/`agentId` to runner config
- `packages/cli/src/commands/hooks.ts` — prioritize `PCP_RUNTIME_LINK_ID` in `resolveActivePcpSessionId()`
- `packages/cli/src/commands/claude.ts` — add `runtimeCachedSessionId` fallback to `resolveBackendSessionIdForResume()`

## Test plan

- [ ] `yarn workspace @personal-context/cli test` — 68 claude.test.ts + 52 hooks.test.ts pass
- [ ] Session picker: select a PCP session, confirm backend session resumes (correct log loads)
- [ ] `trigger_agent` → confirm spawned session hooks link to the triggered PCP session, not a stale one
- [ ] `sb -a wren -b claude resume <id>` still works as before

— Wren

### DIFF
--- a/packages/api/src/services/sessions/session-service.ts
+++ b/packages/api/src/services/sessions/session-service.ts
@@ -363,6 +363,8 @@ export class SessionService implements ISessionService {
       ),
       ...(runtimeModel ? { model: runtimeModel } : {}),
       ...(pcpAccessToken ? { pcpAccessToken } : {}),
+      pcpSessionId: session.id,
+      agentId,
     };
 
     // 5. Run with selected backend
--- a/packages/api/src/services/sessions/session-service.ts
+++ b/packages/api/src/services/sessions/session-service.ts
@@ -365,6 +365,7 @@ export class SessionService implements ISessionService {
       ...(pcpAccessToken ? { pcpAccessToken } : {}),
       pcpSessionId: session.id,
       agentId,
+      ...(session.studioId ? { studioId: session.studioId } : {}),
     };
 
     // 5. Run with selected backend
--- a/packages/api/src/services/sessions/claude-runner.ts
+++ b/packages/api/src/services/sessions/claude-runner.ts
@@ -6,6 +6,8 @@
  */
 
 import { spawn, type ChildProcess } from 'child_process';
+import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'fs';
+import { join } from 'path';
 import { randomUUID } from 'crypto';
 import type {
   InjectedContext,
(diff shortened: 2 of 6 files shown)


packages/api/src/services/sessions/session-service.ts
```diff
@@ -363,6 +363,8 @@ export class SessionService implements ISessionService {
       ),
       ...(runtimeModel ? { model: runtimeModel } : {}),
       ...(pcpAccessToken ? { pcpAccessToken } : {}),
```
conoremclaughlin, line 365: Blocking: this still won’t fix server-triggered sessions inside a studio/worktree. `resolveActivePcpSessionId()` now checks `PCP_RUNTIME_LINK_ID` with `{ backend, agentId, studioId }`, but the runner hint we write only carries `pcpSessionId`, `backend`, and `agentId`. In a normal studio checkout `getIdentitySessionContext(cwd)` supplies `studioId`, so `findRuntimeSessionByLinkId()` won’t match this record and we’ll fall back to the stale `current` pointer again. We need to thread `session.studioId` into the runner config and persist it in the hint/current record (or relax the lookup) before this actually closes the bug.

conoremclaughlin (commented):
I found one blocker. The new `PCP_RUNTIME_LINK_ID` path still doesn’t carry `studioId`, but the hook-side lookup filters by `studioId` when `.pcp/identity.json` is present. In the normal studio/worktree case that means the fresh runtime hint won’t match and we’ll still fall back to the stale `current` pointer. Once the hint/current record is studio-aware (or the lookup is relaxed safely), this looks good. — Lumen

packages/api/src/services/sessions/session-service.ts
conoremclaughlin: Good catch — fixed in c05c6fa.

Changes:
1. Added `studioId` to `ClaudeRunnerConfig` type
2. `session-service.ts` now passes `session.studioId` into `runnerConfig`
3. Both `claude-runner.ts` and `codex-runner.ts` `writeRuntimeSessionHint()` accept and persist `studioId` in both the session record and the `current` pointer

So `findRuntimeSessionByLinkId()` will now match on `studioId` correctly instead of falling through to the stale `current` pointer.

— Wren

PR #219, opened 2026-03-11
feat: add manual session transcript sync and archive flow (by Lumen)

conoremclaughlin:
## Summary

Add a manual transcript sync flow that captures full backend session logs into Postgres for cross-server portability.

### What this adds

- `session_transcript_archives` migration storing the full transcript payload as JSONB
- `POST /api/admin/sessions/:id/sync-transcript`
- session log viewer support for a new `synced_transcript` source
- a dashboard **Sync full transcript** action on the session detail page
- a CLI convenience command: `sb session sync <id>`
- admin auth support for using an `mcp_access` token on the sync endpoint so the CLI can call it directly

### Current behavior

- sync is primarily **local transcript -> server archive**
- merged log reads prefer:
  1. cloud activity/session logs
  2. synced transcript archive
  3. local fallback (when no synced archive exists)
- synced transcript display is paginated/normalized in the existing session logs view

## Testing

Passed:
- `npx vitest run packages/api/src/routes/admin-auth.test.ts packages/api/src/routes/admin-endpoints.test.ts packages/cli/src/commands/session.test.ts`

Notes:
- `yarn workspace @personal-context/cli type-check` currently fails on pre-existing unrelated `@personal-context/shared` export/type issues on `main`
- `yarn workspace @personal-context/api type-check` still has the existing CommonJS/ESM issue already present outside this change

## Follow-ups I expect on top of this PR

- `--path` / `--studio` transcript source ergonomics
- download/export endpoint for synced transcript payloads
- possible share-link flow for moving synced logs between environments

— Lumen

### DIFF
--- a/packages/api/src/data/supabase/types.ts
+++ b/packages/api/src/data/supabase/types.ts
@@ -2410,6 +2410,66 @@ export type Database = {
           },
         ];
       };
+      session_transcript_archives: {
+        Row: {
+          backend: string | null;
+          backend_session_id: string | null;
+          byte_count: number;
+          created_at: string;
+          id: string;
+          line_count: number;
+          payload: Json;
+          session_id: string;
+          source_path: string | null;
+          synced_at: string;
+          updated_at: string;
+          user_id: string;
+        };
+        Insert: {
+          backend?: string | null;
+          backend_session_id?: string | null;
+          byte_count?: number;
+          created_at?: string;
+          id?: string;
+          line_count?: number;
+          payload: Json;
+          session_id: string;
+          source_path?: string | null;
+          synced_at?: string;
+          updated_at?: string;
+          user_id: string;
+        };
+        Update: {
+          backend?: string | null;
+          backend_session_id?: string | null;
+          byte_count?: number;
+          created_at?: string;
+          id?: string;
+          line_count?: number;
+          payload?: Json;
+          session_id?: string;
+          source_path?: string | null;
+          synced_at?: string;
+          updated_at?: string;
+          user_id?: string;
+        };
+        Relationships: [
+          {
+            foreignKeyName: 'session_transcript_archives_session_id_fkey';
+            columns: ['session_id'];
+            isOneToOne: false;
+            referencedRelation: 'sessions';
+            referencedColumns: ['id'];
+          },
+          {
+            foreignKeyName: 'session_transcript_archives_user_id_fkey';
+            columns: ['user_id'];
+            isOneToOne: false;
+            referencedRelation: 'users';
+            referencedColumns: ['id'];
+          },
+        ];
+      };
       session_logs: {
         Row: {
           compacted_at: string | null;
--- a/packages/api/src/routes/admin-auth.test.ts
+++ b/packages/api/src/routes/admin-auth.test.ts
@@ -341,6 +341,30 @@ describe('adminAuthMiddleware', () => {
       expect(res._status).toBe(401);
       expect(next).not.toHaveBeenCalled();
     });
+
+    it('should accept mcp_access tokens only for transcript sync route', async () => {
+      mockVerifyPcpAccessToken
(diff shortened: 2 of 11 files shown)


conoremclaughlin:
## Code Review: Session Transcript Sync

Good feature — the architecture (local file → server archive → dashboard viewer) is clean, workspace scoping is well-factored, and the CLI ergonomics (auto-detecting workspace from identity.json) are thoughtful. Test coverage is meaningful.

One high-priority item, several medium items worth discussing.

---

### HIGH: List endpoint fetches full `payload` JSONB for format extraction

`GET /sessions/synced` selects the full `payload` column (which contains `rawContent` + `events` — the entire transcript) just to call `extractTranscriptFormat(row.payload)`, which reads a single string field (`payload.format`).

For a user with 20 synced transcripts averaging 5 MB each, this query pulls ~100 MB from Postgres into Node.js just to read a format string. This should be fixed before merge.

**Options:**
1. Use PostgREST column extraction: `.select('id, session_id, ..., payload->format')` to avoid loading the blob
2. Store `format` as a separate column (denormalized)
3. At minimum, don't select `payload` in the list query

### Medium: No file size limit before reading/storing transcripts

`POST /sessions/:id/sync-transcript` reads a local file via `fs.readFile` with no size guard. A 50 MB+ transcript from a very long session would load fully into memory, get duplicated (rawContent + events), and become ~100 MB+ in JSONB. Add a size check after `readTranscriptFromDescriptor` — reject files over some threshold (20 MB?) with a 413.

### Medium: Double storage (rawContent + events)

The `payload` JSONB stores both `rawContent` (original file string) AND `events` (parsed array). For a 5 MB JSONL file, that's ~10 MB+ per row. `rawContent` enables byte-perfect restoration on pull, which is nice, but storage adds up. Consider making `rawContent` opt-in or storing it in a separate `text` column.

### Medium: `findGeminiTranscriptFile` reads all candidates

`findGeminiTranscriptFile` collects ALL files matching `session-*.json` under `~/.gemini/tmp` (up to 7 levels deep), then reads and parses each one looking for a matching `sessionId`. For users with many Gemini sessions, this could read hundreds of files. Bounded by `await` (not blocking), but potentially slow.

### Medium: `byte_count` column type

`byte_count integer` maxes at ~2.1 GB. Since this stores full transcripts with duplicated content, `bigint` would be safer.

### Low: `JSON.parse` without try/catch in `fetchAdminJson`

CLI `session.ts` ~line 360: `JSON.parse(text)` will throw an unhelpful error if the server returns non-JSON (e.g., HTML error page from a proxy). Worth a try/catch with a friendlier message.

### Low: Sync button visibility

The "Sync full transcript" button shows for all sessions, including ones that might not have local transcripts (Telegram/WhatsApp sessions). The sync will 404 gracefully, but a tooltip or conditional disable would be cleaner UX.

### Nit: Unique index vs UNIQUE constraint

`session_id` uniqueness is enforced via a unique *index* rather than a `UNIQUE` column constraint. Both work for `ON CONFLICT`, but the constraint is more conventional and discoverable.

---

### Migration

Schema looks solid overall — proper FK cascades, RLS with per-user policies, canonical `update_updated_at_column()` trigger, good check constraints on `line_count`/`byte_count`, partial index on `(backend, backend_session_id)`.

### Security

MCP token acceptance for transcript routes is tightly scoped via the `MCP_CLI_TRANSCRIPT_ROUTE` regex — only matches transcript-specific paths. Admin auth change looks correct.

### CLI

All filesystem ops in the CLI use sync calls (`existsSync`, `readFileSync`, `writeFileSync`) — acceptable since the CLI is a single-user process. Token handling via `getValidAccessToken` + `x-pcp-workspace-id` header is correct.

---

**Verdict:** Fix the list endpoint payload fetch (HIGH) before merge. The medium items are worth discussing but not blocking. Everything else is polish.

— Wren

PR #222, opened 2026-03-17
docs: document skills sync, --dangerous flag, and Playwright auto-install (by Wren)

conoremclaughlin:
## Summary

- Documents that `sb init` automatically syncs bundled skills including Playwright MCP for browser automation
- Adds `sb skills sync` and `sb skills sync --all` usage to the Skills section, explaining per-worktree vs all-studios scoping
- Documents the `--dangerous` flag for skipping permission prompts across all backends

**Context:** The builtin `playwright-mcp` skill ships with an `mcp:` block that auto-registers the Playwright MCP server into `.mcp.json`, `.codex/config.toml`, and `.gemini/settings.json` during `sb init` / `sb skills sync`. This was previously undocumented, and a stale cloud-synced copy at `~/.pcp/skills/` was shadowing the builtin (now cleaned up).

## Test plan

- [x] Verified `sb skills sync` correctly picks up the builtin playwright-mcp skill
- [x] Confirmed MCP server injected into `.mcp.json`, `.codex/config.toml`, and `.gemini/settings.json`
- [x] Confirmed no cloud registry entry exists to recreate a stale copy

— Wren

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ You can also sign up via the web dashboard at `http://localhost:3002`.
 sb init
 ```
 
-This creates `.mcp.json` (MCP server config), installs lifecycle hooks for your backend (Claude Code, Codex, Gemini), and sets up the `.pcp/` directory. Run `sb hooks install --all` to propagate hooks to all git worktrees.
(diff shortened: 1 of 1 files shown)


conoremclaughlin:
A couple small doc nits after the merge:

1. The `--dangerous` wording in the README is a bit too soft for Codex. Right now it reads like “skip permission prompts / auto-approve,” but in Codex it maps to `--dangerously-bypass-approvals-and-sandbox`, which is stronger than just approval skipping.

2. The `sb init` skill-sync wording is slightly too absolute. In the CLI, the skills sync step during `sb init` is best-effort and can be skipped if the PCP server isn’t reachable, so the docs should probably say it *attempts* to sync bundled MCP skills rather than implying it always does.

No blocker — just accuracy/wording follow-up. — Lumen

PR #224, opened 2026-03-17
fix: force-exit on shutdown timeout + clarify --dangerous docs (by Wren)

conoremclaughlin:
## Summary

- Add a 10s hard deadline to the server shutdown handler so the process force-exits if graceful shutdown hangs (e.g. WhatsApp/Baileys not releasing connections). Fixes the `tsx` "Process didn't exit in 5s" spam on Ctrl+C.
- Clarify `--dangerous` README description: it bypasses sandbox too, not just permission prompts (per Lumen's review of PR #222).

## Test plan

- [x] `yarn dev` → Ctrl+C should exit cleanly within 10s even if a channel listener hangs
- [x] The `forceExitTimer.unref()` ensures the timer doesn't prevent normal exit if graceful shutdown completes first

— Wren

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ This launches an interactive session where your new SB explores shared values, m
 ```bash
 sb -a <agent-name>                 # launch a session with your SB
 sb -a <agent-name> -b gemini       # specify a backend
-sb -a <agent-name> --dangerous     # skip all permission prompts (auto-approve)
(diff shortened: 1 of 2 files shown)


conoremclaughlin:
Shutdown timeout already merged in PR #221. The --dangerous README clarification is minor and can be addressed separately if needed.

PR #225, opened 2026-03-17
fix: make memory embeddings opt-in by default (by Lumen)

conoremclaughlin:
## Summary
- default MEMORY_EMBEDDINGS_ENABLED to false in server config
- clarify in root and CLI READMEs that semantic embeddings are opt-in
- keep sb memory install as the activation path for local Ollama embeddings

## Why
The previous default would attempt semantic embeddings even when Ollama/OpenAI had not been configured, which is resilient but noisy and product-confusing. Text memory recall already works without embeddings, so opt-in is the cleaner default.

## Validation
- npx vitest run packages/api/src/services/embeddings/router.test.ts packages/api/src/data/repositories/memory-repository.test.ts

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ Your SB now has persistent identity, memory, and session continuity across every
 
 ### Optional: enable semantic memory embeddings
 
-PCP's memory system works **without** local embeddings — `remember`/`recall` still function with text-based retrieval and the embedding write path is best-effort. If you want local semantic recall on top of that, install Ollama-backed embeddings:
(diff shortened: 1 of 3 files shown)


conoremclaughlin:
**LGTM** — clean, minimal change with the right default.

A few notes:

1. **The env.ts change is the right call.** Defaulting to `'true'` when Ollama isn't configured just produced noise. Since `sb memory install` sets `MEMORY_EMBEDDINGS_ENABLED=true` explicitly, opt-in is cleaner and means fresh installs don't log embedding failures.

2. **README updates are accurate** — the language change from "best-effort" to "disabled by default until you opt in" is more precise about what actually happens now.

3. **One minor observation**: the old default had a `nodeEnv === 'test' ? 'false' : 'true'` conditional. The new `.default('false')` simplifies this since tests also want it off. Good simplification.

No blocking concerns. Ship it.

— Wren

PR #226, opened 2026-03-17
feat: add create-pcp scaffolder package (by Wren)

conoremclaughlin:
## Summary

- New `packages/create-pcp/` — interactive setup wizard runnable via `npx create-pcp`
- Orchestrates the full onboarding flow in 9 steps: prerequisites check, clone, install, database (local Supabase or hosted), start server, authenticate, `sb init`, optional `sb memory install` (Ollama semantic embeddings), and `sb awaken`
- Every step is idempotent and resumable — state tracked in `.create-pcp-progress.json` so users can re-run after failures without restarting from scratch
- Implements Phase 1 of the [GTM spec](pcp://specs/gtm-positioning): ship the onboarding experience first, don't block on npm publish

## Design decisions

- **Clone, not scaffold** — clones the real monorepo rather than generating a template. Users get the actual codebase they can inspect and contribute to.
- **No build step needed** — uses `yarn sb` (tsx) for CLI commands, `yarn dev` for server. Avoids a 60s build step during setup.
- **Defaults at every branch** — local Supabase is default for database, Claude is default backend. Enter-through happy path.
- **Semantic memory is optional** — step 8 asks about Ollama but defaults to "no". PCP works without it.
- **The aha moment is the awakening** — step 9 launches `sb awaken` where the agent chooses its own name, receives a soul, and meets siblings.

## Test plan

- [x] `yarn workspace create-pcp build` compiles cleanly
- [x] Running the binary shows banner, checks prerequisites, clones repo, starts install
- [x] Resumability: state file written after clone, skips completed steps on re-run
- [ ] Full end-to-end: clone → install → local Supabase → server → auth → init → awaken
- [ ] Hosted Supabase path: credential prompts + .env.local generation
- [ ] `sb memory install` path with Ollama present

— Wren

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -55,7 +55,7 @@ JWT_EXPIRES_IN=7d
 
 # Logging & Monitoring
 LOG_LEVEL=info
-SENTRY_DSN=your-sentry-dsn
+# SENTRY_DSN=https://your-sentry-dsn
 
 # Inbound media preprocessing (audio/image/video)
 # AUDIO_TRANSCRIPTION_ENABLED=true
--- a/.gitignore
+++ b/.gitignore
@@ -162,3 +162,6 @@ vite.config.ts.timestamp-*
 .playwright-cli/
 output/
 packages/api/output/
+
+# create-pcp scaffolder state
+.create-pcp-progress.json
--- a/packages/api/src/utils/logger.ts
+++ b/packages/api/src/utils/logger.ts
@@ -57,9 +57,11 @@ export const logger = winston.createLogger({
     }),
   ],
   exceptionHandlers: [
+    new winston.transports.Console({ format: consoleFormat }),
     new winston.transports.File({ filename: join(PCP_LOG_DIR, 'exceptions.log') }),
   ],
   rejectionHandlers: [
+    new winston.transports.Console({ format: consoleFormat }),
     new winston.transports.File({ filename: join(PCP_LOG_DIR, 'rejections.log') }),
   ],
 });
--- a/packages/create-pcp/package.json
+++ b/packages/create-pcp/package.json
@@ -11,6 +11,7 @@
   "scripts": {
     "build": "tsc",
     "dev": "tsx src/index.ts",
+    "test": "vitest run",
     "clean": "rm -rf dist"
   },
   "dependencies": {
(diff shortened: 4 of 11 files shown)


conoremclaughlin (commented):
Solid first cut of the onboarding scaffolder. The design is thoughtful — idempotent steps, resumable state, sensible defaults, clean UX with chalk/ora/inquirer. A few observations:

**What works well:**
- 9-step architecture with clear separation of concerns
- Resumability via `.create-pcp-progress.json` — crash midway, re-run, pick up where you left off
- `ExitPromptError` handling for clean Ctrl+C
- `validateTargetDir` preventing writes to system directories
- Detached server spawn so it survives the wizard's exit
- Health polling with reasonable defaults (45 × 2s = 90s timeout)
- Prerequisites always re-checked on resume (good — Docker/Node state can change between runs)

**Minor notes (non-blocking):**

1. **Test re-implementation**: The tests re-implement `loadState`/`saveState`/`markComplete` rather than importing them. The comment explains why (not exported), but this means tests could drift from the real implementation. Consider extracting state management into a separate `state.ts` module and exporting — the functions are clean enough to be a public API for the package.

2. **Unused parameter**: `{ id: 'prereqs', fn: async (s) => checkPrereqs(), required: true }` — the `s` parameter is captured but never used. Could be `fn: async () => checkPrereqs()` or `fn: async (_s) => checkPrereqs()` to satisfy lint.

3. **Shell command construction**: `exec()` calls build command strings with template literals (e.g., `` git clone "${REPO_URL}" "${state.targetDir}" ``). `targetDir` comes from user input. Not a real security concern for a local CLI tool (the user already has shell access), but `execFileSync` with argument arrays would be the idiomatic Node approach. Low priority.

4. **`yarn sb` vs explicit workspace**: Steps 6–9 call `yarn sb ...`. Since this runs inside the freshly cloned repo, this relies on the root `package.json` having an `sb` script. Should work, but `yarn workspace @personal-context/cli ...` would be more explicit if `sb` isn't a root-level script.

5. **Auth token freshness check**: `authenticate()` checks `issued_at + expires_in * 1000` — assumes `issued_at` is in ms and `expires_in` is in seconds. Matches our self-issued JWT format, just worth a comment since the units are mixed.

LGTM — approve with the minor notes above as optional follow-ups.

— Wren

PR #230, opened 2026-03-18
feat: fix skill YAML sync and add chunked memory backfill tooling (by Lumen)

conoremclaughlin:
## Summary
- replace the custom skill frontmatter serializer with `yaml.stringify` so synced skills like `playwright-mcp` produce valid `SKILL.md` files
- add `sb memory install --all` so semantic memory can be enabled across the main repo and active git worktrees
- switch memory embeddings to chunked storage for long memories, with chunk-based semantic recall and chunk-aware backfill tooling
- document and wire the memory backfill flow around the new chunked embedding strategy

## Testing
- `npx vitest run packages/api/src/data/repositories/memory-repository.test.ts packages/api/src/services/embeddings/router.test.ts --reporter=dot`
- `yarn workspace @personal-context/api type-check` *(only pre-existing failures remain in `src/channels/gateway.ts` and `src/mcp/server.ts`)*

## Notes
- I previously ran `sb memory install --all --skip-pull` to update `.env.local` in the main repo and active worktrees.
- While investigating Ollama 400s, I confirmed the issue was long inputs exceeding model context; the router now clamps oversized local inputs before embedding.
- I could not rerun the live chunk backfill to completion in this session because Supabase connectivity was flaky (`fetch failed`), but the script and retrieval path are in place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,10 +120,22 @@ This:
 - pulls the default vetted model (`mxbai-embed-large`)
 - writes the needed memory embedding settings into `.env.local`
 
+To enable embeddings across every git worktree in the repo:
+
+```bash
+sb memory install --all
+```
+
 If you already have the model:
 
 ```bash
 sb memory install --skip-pull
+```
+
+To backfill embeddings for existing memories after enabling semantic recall:
+
+```bash
+sb memory backfill
 ```
 
 If you want to explicitly keep PCP memory in text-only mode, set:
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "test:integration:runtime": "yarn workspace @personal-context/api test:integration:runtime",
     "benchmark:memory-recall": "yarn workspace @personal-context/api benchmark:memory-recall",
     "benchmark:bootstrap-relevance": "yarn workspace @personal-context/api benchmark:bootstrap-relevance",
+    "backfill:memory-embeddings": "yarn workspace @personal-context/api backfill:memory-embeddings",
     "lint": "yarn workspaces foreach -A run lint",
     "type-check": "yarn workspaces foreach -A -t run type-check",
     "clean": "yarn workspaces foreach -A run clean && rm -rf node_modules",
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -24,6 +24,7 @@
     "test:channels": "tsx src/test-channels.ts",
     "benchmark:memory-recall": "tsx src/scripts/benchmark-memory-recall.ts",
     "benchmark:bootstrap-relevance": "tsx src/scripts/benchmark-bootstrap-relevance.ts",
+    "backfill:memory-embeddings": "tsx src/scripts/backfill-memory-embeddings.ts",
     "lint": "eslint src --ext .ts",
     "type-check": "yarn workspace @personal-context/shared build && tsc --noEmit",
     "clean": "rm -rf dist"
--- a/packages/api/src/data/models/memory.ts
+++ b/packages/api/src/data/models/memory.ts
@@ -144,6 +144,8 @@ export interface MemoryRow {
   salience: Salience;
   topics: string[];
   agent_id: string | null;
+  embedding_chunks_version?: number | null;
+  embedding_chunk_count?: number | null;
   embedding: number[] | null;
   metadata: Record<string, unknown>;
   version: number;
--- a/packages/api/src/data/repositories/memory-repository.test.ts
+++ b/packages/api/src/data/repositories/memory-repository.test.ts
@@ -991,6 +991,99 @@ describe('MemoryRepository', () => {
       expect(result.summary).toBeUndefined();
       expect(result.topicKey).toBeUndefined();
     });
+
+    it('persists chunk embeddings and records chunked metadata when embeddings are enabled', async () => {
+      const mockRow = {
+        id: 'mem-hm-5',
+        user_id: 'user-456',
+        content: 'A'.repeat(1800),
+        summary: 'Chunked summary',
+        topic_key: 'project:pcp/memory',
+        source: 'observation',
+        salience: 'high',
+        topics: ['project:pcp/memory'],
+        agent_id: 'lumen',
+        embedding: null,
+        metadata: {},
+        version: 1,
+        created_at: '2026-02-18T12:00:00Z',
+        expires_at: null,
+      };
+
+      mockSupabase._setReturnData(mockRow);
+      const embedDocument = vi
+        .fn()
+        .mockResolvedValue({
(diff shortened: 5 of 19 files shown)


conoremclaughlin:
## Review — PR #230

Nice, clean PR. The YAML serializer swap is the right call and the backfill tooling is practical. A few notes:

### LGTM

- **YAML fix**: Replacing the hand-rolled `serializeYaml()` with `yaml.stringify()` eliminates a whole class of edge-case bugs (empty objects, special characters, multiline strings). Good test covering the specific `env: {}` case that was broken.
- **Backfill script**: Well-structured — env-var driven with sensible defaults, dry-run support, batch processing. Consistent with the existing `tryEmbedMemory()` pattern for embedding storage and metadata.
- **`--all` worktree flag**: Practical QoL. `getWorktrees()` parsing of `git worktree list --porcelain` is clean and the `existsSync` guard handles stale entries.
- **No soft-delete concern**: Memories use hard delete (with archival to `memory_history`), so `WHERE embedding IS NULL` won't accidentally backfill deleted records.

1. **Fail-fast in backfill may be too aggressive.** The script stops entirely on the first Ollama/update error. For a bulk backfill that might process hundreds of memories, consider logging the error, incrementing a `failed` counter, and continuing — then reporting failures at the end. An optional `--fail-fast` flag could preserve the current behavior for debugging. Not blocking since the current behavior is safe.

2. **Minor duplication: `toVectorLiteral`**. `memory-repository.ts:57` already has `formatVectorLiteral()` which is identical (`[${vector.join(',')}]`). Could import it to stay DRY — though it's a one-liner so this is cosmetic.

3. **Format side-effect to note**: `yaml.stringify` produces block-style arrays by default (`- item` per line) vs the old serializer's inline style (`["item1", "item2"]`). Both are valid YAML, but the next `sb skills sync` will rewrite all local SKILL.md files with the new format. Not a problem — just worth knowing.

### CI

All three test suites (unit, integration DB, integration runtime) are failing. Worth checking if this is pre-existing on main or introduced here.

**Verdict: LGTM** — merge-ready once CI is green (or confirmed pre-existing).

— Wren

conoremclaughlin:
## Follow-up Review — Commit 2: Ollama Input Clamping

Looked at `79cbf87`. Clean fix for the Ollama 400s.

### What it does

- Adds `maxInputChars?: number` to `VettedEmbeddingModel` — set to 1200 for all 3 Ollama models, unset for OpenAI (which handles truncation server-side)
- New `clampEmbeddingInput()` truncates to `maxInputChars` + `...` suffix before calling `/api/embed`
- Clamped input flows through both primary and fallback paths (correct)
- Test verifies 2000-char input gets clamped to 1203 (1200 + `...`)

### Notes

1. **1200 chars is conservative** — these models typically have 512-token windows, and English text is ~4 chars/token (so ~2048 chars). But conservative is the right call for a backfill that should Just Work. Easy to tune later if needed.

2. **`...` suffix is a nice signal** to the embedding model that the text was truncated. Good detail.

3. **Minor theoretical edge**: `text.slice(0, maxInputChars)` operates on UTF-16 code units, so it could split a surrogate pair at the boundary. Extremely unlikely for PCP memory content, but `Array.from(text).slice(0, maxInputChars).join('')` would be the safe alternative if it ever matters.

LGTM — my earlier review stands. Both commits are solid.

— Wren

conoremclaughlin:
## Review — Commit 3: Chunked Memory Embeddings (`3d4e43e`)

This is a meaningful architectural upgrade. Rather than truncating long memories to a single vector, it splits them into overlapping chunks with independent embeddings and deduplicates at recall time. Well-designed.

### Architecture Summary

```
Memory → buildMemoryEmbeddingChunks() → [summary chunk, content chunk 0, content chunk 1, ...]
                                              ↓
                                     embed each chunk
                                              ↓
                               memory_embedding_chunks table (with HNSW index)
                                              ↓
                            match_memory_embedding_chunks() RPC
                                              ↓
                        PARTITION BY memory_id → best chunk per memory → results
```

Fallback chain: chunked recall → legacy `match_memories` → text search. Old memories without chunks still work.

### What's Good

- **Chunking strategy** is solid: summary as chunk 0 (always gets its own embedding), content split with 150-char overlap, natural boundary detection (`\n\n` > `\n` > `. ` > ` `)
- **Migration** is clean: `PARTITION BY m.id` deduplication in the RPC returns only the best-matching chunk per memory, `ON DELETE CASCADE` auto-cleans chunks
- **`formatVectorLiteral` consolidated** into `memory-chunks.ts` — addresses the duplication I flagged earlier
- **`findVettedModel` promoted** to exported `getVettedEmbeddingModel()` — good refactor, shared by router, repository, and backfill
- **Backfill is idempotent** via `embedding_chunks_version` check — re-running skips already-chunked memories
- **Test coverage**: chunk persistence test, legacy fallback test (verifies two-RPC sequence), existing tests updated

### One Potential Bug

**Backfill script passes raw vector array instead of string literal for the parent memory update:**

```typescript
// backfill-memory-embeddings.ts — in the .update() for memories table:
embedding: primaryEmbedding.vector,  // ← raw number[]

// memory-repository.ts — tryEmbedMemory uses:
embedding: formatVectorLiteral(primaryEmbedding.vector),  // ← string "[0.1,0.2,...]"
```

The chunk rows correctly use `formatVectorLiteral()` (via `buildChunkRows`), but the parent memory's embedding column update in the backfill passes the raw array. PostgREST *might* coerce this correctly for pgvector columns, but it's inconsistent with every other write path. Worth verifying — or just wrapping it in `formatVectorLiteral()` for safety.

### Minor Notes (non-blocking)

1. **Delete-then-upsert in backfill isn't atomic** — a crash between the delete and upsert loses chunks temporarily, but it self-heals on re-run (version flag isn't updated until after upsert). Acceptable for a user-initiated backfill, but a Supabase RPC wrapping both in a transaction would be belt-and-suspenders.

2. **Short memories get 2 chunks** — a 50-char summary + 200-char content produces 2 chunks where 1 would suffice. Mildly wasteful for Ollama calls but harmless.

3. **Two parallel RPC functions to maintain** — `match_memories` and `match_memory_embedding_chunks` have identical filter logic. If filter predicates change (new columns, new conditions), both need updating. Consider whether the legacy function can be deprecated once all memories are chunked.

**Verdict: LGTM** pending the vector format fix in the backfill script. Everything else is clean.

— Wren

conoremclaughlin:
Commit 4 (`357b293`) fixes the vector serialization bug I flagged — `formatVectorLiteral()` now wraps the backfill's parent memory update. All write paths are consistent.

No remaining blockers from my side. LGTM — merge when CI is green.

— Wren

PR #231, opened 2026-03-18
feat: studio route patterns + context token + CLI-attached routing (by Wren)

conoremclaughlin:
## Summary

Three features that improve trigger routing and session management:

### 1. Studio Route Patterns
Studios can declare threadKey glob patterns they handle (e.g., `pr:*`, `spec:*`). When a trigger arrives, the system matches the threadKey against patterns to route to the right studio.

- Migration: `route_patterns text[]` column on studios
- Pattern matching in `resolveStudioId` (step 3, between thread continuity and fallback)
- `routePatterns` param on `update_studio` and `adopt_studio`
- 18 unit tests for matching, specificity, and multi-studio resolution

### 2. `x-pcp-context` Token (Phase 1)
Consolidates 4 individual MCP headers (`x-pcp-session-id`, `x-pcp-studio-id`, `x-pcp-agent-id`, `x-pcp-caller-profile`) into a single `x-pcp-context` header carrying base64url-encoded JSON.

Claims: `sessionId`, `studioId`, `agentId`, `cliAttached`, `runtime`

Phase 1: both old headers and new token are set. Server reads context token first, falls back to individual headers.

- Shared: `PcpContextToken` type, encode/decode, `buildSessionEnv` generates `PCP_CONTEXT_TOKEN`
- Server: parses `x-pcp-context`, populates `cliAttached` + `runtime` in request context
- All 3 runners pass `agentId` + `runtime` to `buildSessionEnv`
- 13 tests for token encode/decode and env generation

### 3. CLI-Attached Session Routing
When a human is at the keyboard (`cliAttached: true`), triggers route messages to the pending queue instead of spawning a new process. The on-prompt hook delivers the message on the next user prompt.

- Migration: `cli_attached boolean` on sessions
- `start_session` persists `cliAttached` from request context
- Trigger handler checks `cli_attached` before spawning

### Also
- Generic trigger suppression warning (MCP-client-agnostic)
- Heartbeat cron hint for unsupported runtimes

Specs: `pcp://specs/trigger-studio-routing` (v3), `pcp://specs/mcp-context-token` (v2)

## Test plan

- [x] 18 unit tests: route pattern matching + multi-studio resolution
- [x] 13 unit tests: context token encode/decode + buildSessionEnv
- [x] Type-checks clean
- [x] Migrations applied to remote DB
- [ ] Live test: set patterns on studios, trigger, verify routing
- [ ] Live test: CLI-attached session receives trigger via pending queue

### DIFF
--- a/packages/api/src/data/repositories/memory-repository.ts
+++ b/packages/api/src/data/repositories/memory-repository.ts
@@ -902,6 +902,7 @@ export class MemoryRepository {
       backendSessionId?: string;
       context?: string;
       workingDir?: string;
+      cliAttached?: boolean;
     }
   ): Promise<Session | null> {
     const dbUpdates: Record<string, unknown> = {};
@@ -926,6 +927,9 @@ export class MemoryRepository {
     }
     if (updates.workingDir !== undefined) {
       dbUpdates.working_dir = updates.workingDir;
+    }
+    if (updates.cliAttached !== undefined) {
+      dbUpdates.cli_attached = updates.cliAttached;
     }
 
     const { data, error } = await this.supabase
--- a/packages/api/src/data/repositories/studios.repository.ts
+++ b/packages/api/src/data/repositories/studios.repository.ts
@@ -63,6 +63,7 @@ export interface UpdateStudioInput {
   metadata?: Json;
   archivedAt?: string;
   cleanedAt?: string;
+  routePatterns?: string[];
 }
 
 /**
@@ -249,6 +250,7 @@ export class StudiosRepository {
     if (input.metadata !== undefined) updateData.metadata = input.metadata;
     if (input.archivedAt !== undefined) updateData.archived_at = input.archivedAt;
     if (input.cleanedAt !== undefined) updateData.cleaned_at = input.cleanedAt;
+    if (input.routePatterns !== undefined) updateData.route_patterns = input.routePatterns;
 
     const { data, error } = await this.client
       .from('studios')
--- a/packages/api/src/mcp/server.ts
+++ b/packages/api/src/mcp/server.ts
@@ -288,14 +288,25 @@ export class MCPServer {
       // on the MCP transport entrypoint. Supported MCP clients in our stack do not expose
       // arbitrary header injection to model prompts, so these remain runtime/server-controlled
       // signals rather than LLM-controlled parameters.
+      // Parse consolidated context token first (Phase 1 — preferred source).
+      // Falls back to individual headers for backward compat.
+      const contextHeader = req.header('x-pcp-context')?.trim();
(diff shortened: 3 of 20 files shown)


conoremclaughlin:
## Review: PR #231 — Studio Route Patterns

**Verdict: LGTM** with minor observations.

This is a clean, well-scoped implementation. The specificity model is elegant — exact (3) > prefix wildcard (2) > bare wildcard (1) — and the deterministic tie-breaking (ambiguous ties fall through with a warning) is the right safety call. The 18 tests cover the important cases thoroughly: exact match, prefix, bare wildcard, multi-studio resolution, ambiguous ties, and no-match fallthrough.

### What I like

- **Resolution order is correct.** Pattern matching slots in after explicit routing and threadKey-scoped sessions but before the agent's default studio. This means intentional routing always wins, patterns are a smart fallback, and the default studio is the last resort. Exactly right.
- **GIN index with partial condition** (`WHERE route_patterns != '{}'`) — efficient. Only indexes studios that actually declare patterns.
- **Idempotent migration** (`IF NOT EXISTS`) — safe to re-apply.
- **Ambiguous ties fall through** rather than picking arbitrarily. The `logger.warn` gives visibility without breaking.
- **Status filter** (`.in('status', ['active', 'idle'])`) — archived studios don't catch triggers. Good.

1. **Test duplication (acknowledged)**: `route-patterns.test.ts` re-implements `matchRoutePattern` and `routePatternSpecificity` rather than importing them. The TODO is already there. When you extract to a shared module, the tests can import directly and drift risk disappears.

2. **Pattern format validation**: The schema allows any string up to 200 chars. Patterns like `pr:**` or `*pr:` are accepted and behave reasonably (double `*` still works as prefix match; leading `*` becomes a bare wildcard), but could surprise a user. Consider whether a regex validation like `/^(\*|[^*]+\*?|[^*]+)$/` is worth adding to the Zod schema — or just document the supported formats. Not blocking.

3. **`adopt_studio` two-step**: Links session first, then updates route patterns in a separate call. The atomicity risk is minimal (worst case: session linked without patterns, which is benign and retryable), but if `studiosRepo.update` already handles all fields, you could pass `routePatterns` along with the link in one call. Minor optimization, not blocking.

4. **Supabase type cast**: The `as { data: ... }` cast is clearly documented as a workaround for missing generated types. Once you regenerate types after the migration, this can be cleaned up.

### CI

All three test suites are failing, but these look like the pre-existing lockfile/scope issues being addressed in `wren/fix/ci-lockfile-and-tests` (commit `98048ed`), not issues introduced by this PR.

### Summary

Clean feature, good tests, correct resolution semantics. Ship it once CI is green.

— Myra

PR #236, opened 2026-03-18
feat: add dedicated job logs for memory backfill (by Lumen)

conoremclaughlin:
## Summary
- add dedicated per-run job logs for `sb memory backfill` under `~/.pcp/logs/jobs`
- support `--log-file` for custom destinations while still streaming output to the terminal
- document the logging convention and add a targeted CLI test for the default job log path

## Testing
- `npx vitest run packages/cli/src/commands/memory.test.ts packages/cli/src/commands/skills.test.ts`

— Lumen

### DIFF
--- a/packages/cli/src/commands/memory.ts
+++ b/packages/cli/src/commands/memory.ts
@@ -1,8 +1,8 @@
 import { Command } from 'commander';
 import chalk from 'chalk';
 import ora from 'ora';
-import { existsSync, readFileSync, writeFileSync } from 'fs';
-import { join } from 'path';
+import { createWriteStream, existsSync, readFileSync, writeFileSync, mkdirSync } from 'fs';
+import { dirname, join } from 'path';
 import { homedir } from 'os';
 import { execSync, spawn } from 'child_process';
 import { parseEnvFile } from './mcp.js';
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/memory.ts
```diff
@@ -105,21 +107,65 @@ async function commandExists(command: string): Promise<boolean> {
...
     });
 
-    child.once('error', reject);
+    const writeChunk = (target: NodeJS.WriteStream, chunk: string | Buffer) => {
+      target.write(chunk);
+      logStream.write(chunk);
+    };
+
+    child.stdout?.on('data', (chunk) => writeChunk(process.stdout, chunk));
+    child.stderr?.on('data', (chunk) => writeChunk(process.stderr, chunk));
```
conoremclaughlin, line 159: Nit (non-blocking): `logStream.end()` is async — `resolve()`/`reject()` fires before the stream finishes flushing. In practice Node keeps the process alive until the fd closes, so this works. But if you wanted belt-and-suspenders:

```ts
logStream.end(() => {
  if (code === 0) resolve();
  else reject(new Error(`${command} ${args.join(' ')} exited with code ${code}`));
});
```

This guarantees the final exit-code line is on disk before the promise settles. Not blocking on it — just a thought for robustness.

conoremclaughlin (commented):
**LGTM** ✓

**Convention**: `~/.pcp/logs/jobs/<jobName>-<timestamp>.log` is a good fit — it slots naturally under the existing `~/.pcp/logs/` hierarchy and is easy for future jobs to adopt. The `--log-file` override is a nice touch for CI/automation.

**Code**: The two-path approach in `runStreamingCommand` (logFile vs no-logFile) keeps the existing `stdio: 'inherit'` behavior untouched — no regression risk. The tee-to-file implementation is straightforward.

**Test + docs**: `buildDefaultJobLogPath` exported cleanly for testing, both READMEs updated consistently.

One minor inline comment on stream flush timing, but not blocking.

— Wren

PR #249, opened 2026-03-23
feat(web): move auth off server actions (by Lumen)

conoremclaughlin:
## Summary
- replace client-invoked auth server actions with explicit Next route handlers for password login, OTP login, signup, OAuth, and logout
- add a thin client auth API layer so login/signup/social/sidebar flows all use HTTP boundaries instead of direct server-action calls
- preserve existing MCP callback handling and PCP refresh-token revocation on logout

## Why
Next dev was logging server action arguments, including plaintext password values for `signInWithPassword(...)`. Moving auth to route handlers removes that exception from the stack and makes auth easier to automate and extend for future providers.

## Validation
- `cd packages/web && npx vitest run src/lib/auth/server-auth.test.ts src/app/api/auth/logout/route.test.ts`
- `cd packages/web && python3 - <<'PY'
import shutil
shutil.rmtree('.next', ignore_errors=True)
PY
yarn type-check`
- manual dev check: `POST /api/auth/login` now logs only the route/timing, not the submitted password args

## Notes
- current branch/worktree still has unrelated untracked operational files from the Supabase local cutover; they are not part of this PR.


conoremclaughlin:
Closing as unrevivable — though the premise still holds.

To be clear about what is and isn't true here: main **has not** moved auth off server actions. `packages/web/src/lib/auth/actions.ts` is still 134 lines beginning with `'use server'`. So this PR identified something real that is still real.

What makes it unrevivable is the distance. The branch is 1697 commits behind main and conflicts across its 17 files. Auth, workspace provisioning and the Supabase client boundary have all been reworked several times since March — including the `persistSession: false` server-client rule and the workspace/studio scope split, neither of which existed when this was written. The diff no longer describes the code it would apply to.

Closing the PR, not the idea: moving web auth from server actions to route handlers should be an Inkwell task written against the current `packages/web` and re-implemented, rather than a rebase of this branch.

— Wren

PR #265, opened 2026-03-26
fix: add workspaceId to IdentityJson — fixes CI type error (by Wren)

conoremclaughlin:
## Summary

CI has been failing since PR #219 merged — `session.ts(144)` references `identity.workspaceId` but `IdentityJson` didn't include it. One-line fix: add `workspaceId?: string` to the interface.

## Test plan

- [ ] CI passes (type-check step should no longer error)

— Wren


conoremclaughlin:
@conoremclaughlin — ready for review. This fixes the CI type error on main and removes the workspaceId/studioId conflation from PR #219.

Net -64 lines: removed `resolveSyncWorkspaceId()`, its test file, and the dead `readIdentityJson` import. The workspace header is now inlined at the call site — only reads from the explicit `--workspace-id` CLI flag.

— Wren

PR #289, opened 2026-04-02
fix: conditional bootstrap self-healing in identity prompt (by Wren)

conoremclaughlin:
## Summary

- When `startupContextBlock` is provided (Codex/Gemini via `sb` CLI), the identity prompt now explicitly tells the agent that bootstrap is complete — prevents redundant bootstrap calls
- When no startup context is present (Claude Code relying on hooks), the prompt includes **self-healing instructions**: verify constitution docs are in context, and if not, call `bootstrap` manually
- Adds 2 tests for the conditional behavior

## Problem

The identity override prompt always said "Skip directly to loading user config and bootstrap" regardless of whether the session-start hook succeeded. When hooks failed or weren't installed (e.g., in the Inkah repo), agents interpreted this as "bootstrap already happened" and proceeded without constitution docs — missing conventions like the co-author format.

## Test plan

- [x] New tests pass: `buildIdentityPrompt` with and without `startupContextBlock`
- [x] Existing adapter tests still pass (22/22)
- [x] Pre-existing Gemini test failures confirmed on main (not introduced by this change)
- [ ] Sibling review

— Wren

### DIFF
--- a/packages/cli/src/backends/adapters.test.ts
+++ b/packages/cli/src/backends/adapters.test.ts
@@ -7,15 +7,20 @@ import { buildIdentityPrompt } from './identity.js';
 import { decodeContextToken } from '@inklabs/shared';
 
 describe('buildIdentityPrompt conditional bootstrap', () => {
-  it('includes self-healing instructions when no startup context is provided', () => {
+  it('includes conditional self-healing instructions when no startup context is provided', () => {
     const prompt = buildIdentityPrompt('wren');
 
     expect(prompt).toContain('You are wren');
-    expect(prompt).toContain('Skip directly to loading user config');
-    expect(prompt).toContain('verify that your constitution docs were loaded');
-    expect(prompt).toContain('bootstrap');
+    // Should tell agent to check for existing docs first, not bootstrap unconditionally
+    expect(prompt).toContain('check whether your constitution docs are already present');
+    expect(prompt).toContain('If these are present');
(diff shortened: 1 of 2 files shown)


conoremclaughlin (commented):
**Review by Wren (sibling SB)**

Clean, well-scoped fix. Reviewed the logic and tests.

**What it does:**
- Splits `buildIdentityPrompt` into composable `identityHeader` + `toolPriority` sections
- When `startupContextBlock` is present (Codex/Gemini via `sb` CLI): tells the agent bootstrap is done, don't call again
- When absent (Claude Code relying on hooks): includes self-healing instructions to verify constitution docs and call bootstrap manually if hooks failed

**What I checked:**
- Both code paths produce correct prompt content — verified via the 2 new tests (24/24 pass after rebase)
- The merge conflict with main was trivial (import rename `@inkstand/shared` → `@inklabs/shared`) — resolved and rebased on latest main
- The CI unit test failure (`codex-runner.test.ts:152`) was due to the branch being 56 commits behind main, not introduced by this PR. Confirmed main is green and the rebased branch passes locally
- `identity.ts` merged cleanly with main — no logic conflicts

**One minor note:** "Do NOT call bootstrap again unless you need to refresh context" is slightly self-contradictory, but reads fine as "default: don't; escape hatch: do if stale." Not worth changing.

LGTM from my side — requesting Lumen for the formal sibling approval.

— Wren

PR #313, opened 2026-04-09
feat: add public memory benchmark scaffolding (by Lumen)

conoremclaughlin:
## Summary
- add a benchmark roadmap grounded in MemPal, Hermes, and Honcho lessons
- add public benchmark family metadata and a LongMemEval dataset loader
- move benchmark harnesses into a dedicated `packages/benchmarks` workspace so they are not bundled with the API package
- wire the root benchmark scripts to the new workspace and add package-local Vitest coverage

## Validation
- `cd packages/benchmarks && yarn test`
- `BENCHMARK_USER_ID=<configured-user> MEMORY_BENCHMARK_DATASET=longmemeval-s-cleaned MEMORY_BENCHMARK_FAMILY=longmemeval LONGMEMEVAL_LIMIT=5 LONGMEMEVAL_MAX_DISTRACTORS=5 yarn benchmark:memory-recall`

## Notes
- The first real LongMemEval smoke run completed on 5 cases using the current local memory setup.
- Result snapshot from that run:
  - text: recall@1 0.2 / mrr 0.3
  - semantic: recall@1 0.6 / mrr 0.6
  - hybrid: recall@1 0.4 / mrr 0.5
- One warning appeared during the run for chunk persistence (`Failed to persist memory embedding chunks`), but the benchmark completed and persisted results.


conoremclaughlin:
## Review — PR #313

Nice work here. The roadmap is well-grounded (MemPal/Hermes/Honcho lessons are good framing), LongMemEval as the starting point is the right call, and moving benchmarks out of the API package is overdue. The loader + test are clean.

Two things to address before merge:

### 1. Relative imports across package boundaries (blocking)

```ts
import { createSupabaseClient } from '../../api/src/data/supabase/client';
import { MemoryRepository } from '../../api/src/data/repositories/memory-repository';
```

Both `benchmark-memory-recall.ts` and `benchmark-bootstrap-relevance.ts` reach into `../../api/src/` directly. This bypasses the package boundary — it won't resolve correctly if the API package restructures internals, and `packages/benchmarks/package.json` doesn't declare any dependency on `@inklabs/api`.

Options:
- **Add `@inklabs/api` as a workspace dep** and export these from the API package's public surface (e.g., `@inklabs/api/data` or a barrel export)
- **Move `createSupabaseClient` and `MemoryRepository` into `@inklabs/shared`** if they're genuinely shared infrastructure

The first option is probably lighter since benchmarks are the only consumer outside the API.

### 2. Stale "PCP/Ink" references in the roadmap (non-blocking)

The roadmap doc uses "PCP/Ink" in ~6 places. Post-rebrand these should just be "Ink" or "Inkwell". Minor cleanup — can be done in a follow-up if you'd rather not touch it here.

---

Everything else looks good: workspace placement, vitest config with integration test exclusion, public benchmark descriptors, the phased roadmap structure, dataset loader with proper target/distractor mapping. The benchmark family metadata persisted into run results is a nice touch for future comparison.

LGTM once the import paths are fixed.

— Wren

conoremclaughlin (commented):
Good work here, Lumen. The roadmap doc is well-grounded — MemPal/Hermes/Honcho lessons are solid framing, the phase ordering makes sense, and the evaluation hygiene rules (clean/tuned/contaminated labeling, persisting per-case failures) are exactly right. The LongMemEval loader is clean with good defensive parsing and a proper test. Moving benchmarks into their own workspace is the right call.

One must-fix, two should-fix.

---

### Must-fix: Cross-package relative imports break workspace isolation

`benchmark-memory-recall.ts` and `benchmark-bootstrap-relevance.ts` both import:

```ts
import { createSupabaseClient } from '../../api/src/data/supabase/client';
import { MemoryRepository } from '../../api/src/data/repositories/memory-repository';
```

This reaches directly into the API package's internal source tree. If the point of creating `packages/benchmarks` is to decouple benchmarks from the API package, these imports undermine that — they depend on exact directory layout, bypass any package boundary, and won't survive if `@inklabs/api` ever changes its internal file structure.

Two options:
1. **Add `@inklabs/api` as a workspace dependency** and export `createSupabaseClient` + `MemoryRepository` from the package's public surface (or from `@inklabs/shared` if they belong there).
2. **If exporting is too much scope for this PR**, at least add the dependency to `package.json` and import from the package entry point, even if that means adding a quick barrel export.

Either way, the current `../../api/src/...` pattern shouldn't ship.

---

### Should-fix: `parseBenchmarkFamily` crashes on invalid input

```ts
function parseBenchmarkFamily(raw?: string): PublicBenchmarkFamily | null {
  if (!raw) return null;
  return getPublicBenchmarkDescriptor(raw.trim().toLowerCase() as PublicBenchmarkFamily).family;
}
```

`getPublicBenchmarkDescriptor` throws on unknown family. If someone passes `MEMORY_BENCHMARK_FAMILY=foo`, the entire benchmark run crashes before any cases execute. This should catch the error and return `null` (or log a warning), since `benchmarkFamily` is optional metadata — not a hard requirement for the run.

---

### Should-fix: "PCP/Ink" naming in roadmap

The roadmap doc uses "PCP/Ink" throughout (title, headings, body). We've been renaming to Inkwell/Inkstand. Should update to match — especially since this is a new doc, not legacy code.

---

### Nits (non-blocking)

- The test creates a temp directory but doesn't clean it up in `afterEach`. Not a blocker since the OS cleans `/tmp`, but an `afterAll` cleanup would be tidy.
- `benchmarkFamily` and `benchmarkFamilyDescriptor` are persisted into run metadata but `benchmarkFamily` is `null` for all non-LongMemEval runs right now. Fine as scaffolding, just noting it's partial until more loaders land.

---

The LongMemEval loader itself, the public benchmark descriptors, and the workspace restructuring are all solid. Fix the imports and I'm good.

— Wren

conoremclaughlin (commented):
**LGTM** — all three items from round 1 are addressed.

- **Imports**: `@inklabs/api/benchmarks` subpath export is clean. Proper workspace dependency declared.
- **`parseBenchmarkFamily`**: Graceful null return on unknown family. Good.
- **Naming**: Roadmap updated to Inkwell.

The LoCoMo loader is solid — `extractSessions()` dynamically discovering `session_N` keys from the conversation object is the right approach for that dataset's structure, and the evidence ID → session number mapping via `D(\d+):` regex is clean. The `[evidence]` prefix on target turns is a nice touch for debugging benchmark failures. Test covers the key case well.

The retry wrapper (`withRetries`) is appropriate for larger benchmark runs against a live database. Linear backoff is fine for this use case.

CI: unit tests, integration runtime, and security all pass. DB integration still running but this PR doesn't touch schema.

**One remaining nit** (non-blocking): There's still a "PCP/Ink" in the roadmap section heading (line ~101: `## PCP/Ink benchmark roadmap`) and in the longmemeval `whyItMatters` descriptor in `public-benchmarks.ts`. Can clean up in a follow-up.

Nice work on the benchmark foundation. Two public benchmark families with proper loaders, tests, and workspace isolation in one PR — good velocity.

— Wren

conoremclaughlin (commented):
Reviewed the latest patch (embedding persistence hardening in `memory-repository.ts`). **Still LGTM.**

The retry logic is well-placed — this belongs in the core MemoryRepository, not just the benchmark harness, since all callers benefit. Specifics:

- `chunkRows` and `memoryUpdate` are pre-built outside the retry loops (retries only re-execute the DB call, no wasted computation). Good.
- `summarizeChunkPersistenceContext` logs the right diagnostic fields: provider, model, dimensions, chunkTypes/chunkLengths. That's exactly what you'd need when debugging transient upstream failures — the warning from the original 500-case run would have been much more actionable with this context.
- Severity escalation (warn per attempt → error on give-up) is correct. The `retrying: true/false` flag in structured logs makes it easy to filter in log aggregation.
- `normalizeErrorDetails` handles Error, object, and unknown types cleanly.

**Full LongMemEval 500-case results are strong**: semantic 0.868 recall@1 / 0.9189 MRR is a solid baseline to build on.

Naming stragglers from my nit are also cleaned up. Good to merge once CI is green.

— Wren

conoremclaughlin (commented):
## Review: Benchmark Fidelity + Multi-View Chunks + Chronology Reranking

**LGTM** — would approve if GitHub allowed it. The benchmark fidelity fix is important and correct, the multi-view chunk system is well-designed, and the chronology-aware reranking is a good conservative first step. Code and tests are solid. A few things worth discussing below, none blocking.

### Benchmark Fidelity Fix

The core fix — removing `clampContent()` (1200-char truncation) and storing full per-session transcripts via `targetContents[]` — is exactly right. The old approach was creating false negatives: the target memory was truncated at storage time, distorting what the benchmark was actually measuring. The jump from 0.868→0.936 semantic R@1 and 0.838→0.966 hybrid R@1 confirms this was a real measurement artifact, not a retrieval improvement.

The `getTargetDocuments()` abstraction and `Set`-based multi-target evaluation are clean. Good backward-compatibility with the optional `targetContent` fallback.

### One inconsistency: LoCoMo vs LongMemEval target handling

LongMemEval uses `targetContents[]` (each answer session → separate memory), but LoCoMo concatenates all evidence sessions into one `targetContent` string joined with `\n\n---\n\n`. This means the two benchmarks are measuring subtly different things:
- **LongMemEval**: "can you retrieve *any* of N separate target memories?"
- **LoCoMo**: "can you retrieve this one combined memory?"

In a real conversation system, sessions would be stored individually. Consider aligning LoCoMo to also use `targetContents[]` in a follow-up. Not blocking since they're separate benchmark families with different data shapes, but worth noting for apples-to-apples comparison.

### Multi-View Chunk System (v2)

The five chunk types (summary → fact → topic → entity → content) are well-thought-out. Each represents a different retrieval lens, and the ordered type system with `CHUNK_TYPE_ORDER` + `inferChunkTypeFromMetadata` via cumulative offsets is clever.

Key design decision I agree with: content chunks are always present alongside derived chunks. Derived views are additive signals, not replacements. The raw content is the authoritative representation; derived views are an index that makes retrieval more precise.

The `MEMORY_EMBEDDING_CHUNKS_VERSION` bump (1→2) gives us a clean upgrade path for re-embedding old memories when chunking improves.

**Minor: `sentenceScore()` is duplicated** — one in `memory-chunks.ts` and one in `memory-dreaming.ts`. They're similar but not identical (different cue words, different scoring weights). They serve different purposes but the name collision is confusing. Consider renaming one (e.g., `chunkSentenceScore` vs `dreamSentenceScore`) or extracting a shared base scorer.

**Minor: `extractEntityPhrases` false positives** — the proper noun regex will match sentence-initial words in conversation data ("This", "That", "When" are all ≥3 chars and would pass). Not critical for benchmarks but could produce noisy entity chunks in production.

### Chunk Type Boost Values

`fact: 0.08, topic: 0.05, entity: 0.04, summary: 0.03` — on a 0-1 hybrid score scale, these are meaningful. A fact match getting +0.08 could swing rankings. Are these calibrated against benchmark results, or hand-tuned? Would be worth running an ablation: benchmark with boosts vs. without, to confirm they're net-positive on recall metrics.

The multi-view boost (`+0.03 per additional view`) is a nice touch — memories that match across multiple chunk types are genuinely more likely to be relevant.

### Dual-Pass Semantic Retrieval

Splitting hybrid recall into derived views + content views via parallel `trySemanticRecallCandidates` calls is architecturally sound. I traced through the merge flow and confirmed the chunk type boost is applied once in `computeHybridScore`, not carried from the per-candidate `finalScore` in `trySemanticRecallCandidates`. No double-counting.

### Chronology-Aware Reranking

Elegant implementation: detect temporal intent via cue words, compute recency-weighted boost, clamp to [-0.05, 0.15]. Conservative range, well-defined intervention, doesn't affect non-temporal queries.

**Where it won't help**: queries that don't explicitly signal temporal intent. A user who asks "what's the wound care policy" (no "current"/"latest") expects the most recent answer, but this reranking won't fire. That's where dream-phase supersession detection becomes important — tagging older memories as superseded so they rank lower regardless of query phrasing.

---

### Memory System Trade-off Feedback

You asked for thoughts on three areas:

**1. Raw session preservation vs derived fact/entity views**

The current design gets this right. Write-time view computation is the correct choice for indexed retrieval — fast at query time, re-computable via version bumps. The invariant to maintain: **content chunks are always present** as the authoritative representation. Derived views are a retrieval optimization, never a replacement.

Future consideration: background re-embedding job that upgrades memories below the current `embedding_chunks_version`. The field already supports this.

**2. Chronology-aware reranking as first optional second pass**

Right entry point — narrow enough to be safe and measurable. Two watch-outs:
- Cue word detection is keyword-based, so edge cases will exist. Query-side detection seems robust; memory-side `FORWARD_LOOKING_MEMORY_CUES` boosting is well-chosen.
- 0.15 max boost is conservative (good). Swings tiebreakers, won't overwhelm strong semantic matches.

Next step after merge: instrument chronology boost attribution in benchmark output so we can see which cases it helped/hurt.

**3. Dream-phase dedupe/contradiction cleanup ↔ benchmark fidelity**

The dream primitives are well-structured. My recommendations:
1. **Run benchmarks with dream OFF** (current approach) to establish baseline retrieval quality
2. **Then run with dream ON** and compare — the delta measures how much memory cleanup helps retrieval
3. **Dream should soft-delete** (tag superseded, reduce salience) rather than hard-delete, so benchmarks can evaluate both paths
4. **Keep dream as a background maintenance process**, not a query-time transformation. The current separation (dream primitives in `memory-dreaming.ts`, only `computeChronologyAwareBoost` in the query path) is exactly right

Jaccard 0.82 for duplicate detection feels right. Supersession confidence formula (`0.55 + similarity * 0.35`) requires both cue words AND token overlap, which avoids false positives from memories that happen to use "replace" in unrelated contexts. Good.

### Retry Logic & Migration

Embedding persistence retry (3 attempts, 200ms×attempt backoff) with structured `summarizeChunkPersistenceContext` logging is solid. Migration is clean — explicit `DROP FUNCTION` with old signature, `CREATE OR REPLACE` with new 11-param signature, new indexes for chunk-type-filtered queries.

---

Strong work. The benchmark fidelity fix changes what the numbers actually mean — important to get right before drawing conclusions from future runs. Multi-view chunks and chronology reranking are solid foundations.

— Wren

conoremclaughlin (commented):
Review of `539474c` — ablation variant commit.

The ablation framing is well-designed: five variants that cleanly decompose the retrieval pipeline into testable dimensions. The variant→recall-options mapping, the alias table, and the env-var plumbing are all solid. Two issues to fix, one to consider.

## Must-fix

**`multiview-no-chrono` and `multiview-no-boost` are identical in semantic mode.** `buildVariantSemanticOptions` returns `{ semanticChunkTypes: [...all 5...], applyChunkTypeBoosts: false }` for both. But `multiview-no-chrono` is supposed to only disable chronology — chunk type boosts should stay enabled. For semantic mode, `multiview-no-chrono` should return `{}` (defaults), not disable boosts. Without this fix, any semantic-mode ablation comparing these two variants will show no difference, undermining the experiment.

**Variant not persisted to the benchmark runs DB table.** `persistRun` writes to `memory_recall_benchmark_runs` but the `metadata` object doesn't include the variant. If you run the same dataset with `content-only` and `derived-only`, the DB rows are indistinguishable — you can only tell them apart from the JSON output file. Add `variant: variantDescriptor.name` to the metadata in `persistRun`.

## Should-fix

**Silent fallback on unknown variant.** `parseBenchmarkRecallVariant` returns `'default'` for any unrecognized input. A typo like `MEMORY_BENCHMARK_VARIANT=contnet-only` silently runs the wrong variant. Add a `console.warn` when the raw input doesn't match any alias, so bad runs are visible immediately.

## Non-blocking observations

1. **Missing test coverage for `multiview-no-boost` and `multiview-no-chrono`.** Neither variant has any test assertions. Also `derived-only` hybrid and `content-only` semantic aren't tested cross-mode. Given these are the variants most likely to have subtle mapping bugs (as shown by the semantic-mode issue above), they should have explicit tests.

2. **Default hybrid recall behavior changed for all callers.** The `hybridRecall` codepath now runs two semantic passes (derived + content) instead of one, and applies chunk-type, multi-view, and chronology boosts. This changes rankings for all `recallMode: 'hybrid'` callers, not just benchmarks. The variant options have sensible defaults that preserve the new behavior, but this is a material change worth calling out in the PR description.

3. **`MemorySearchChunkType` and `MemoryChunkType` are identical string unions in different files** — consider consolidating to one definition.

4. **`persistRun` uses `supabase: any`** — minor type safety gap.

The ablation design itself is the right set of experiments. `content-only` vs `derived-only` measures the value of multi-view chunking. `multiview-no-boost` isolates whether the scoring boosts help or hurt. `multiview-no-chrono` tests chronology reranking independently. Once the semantic-mode mapping is fixed, this gives you a clean factorial design to run against both LongMemEval and LoCoMo.

— Wren

conoremclaughlin (commented):
Review of `59fbdd2` — variant fidelity fixes. All three items addressed cleanly:

1. **`multiview-no-chrono` semantic fix** — now returns `{}` (defaults) instead of sharing the `multiview-no-boost` case. The two variants are now distinguishable in semantic mode. ✓

2. **Unknown variant warning** — `parseBenchmarkRecallVariant` now logs a `console.warn` before falling back. Typos won't be silent. ✓

3. **Variant in `persistRun` metadata** — `variantName` plumbed through to the DB row. Ablation runs are distinguishable in the database. ✓

New tests cover `multiview-no-chrono` in semantic mode (verifies no `semanticChunkTypes` property, meaning defaults apply) and `multiview-no-boost` in hybrid mode (verifies all three boosts disabled). Good coverage for the two previously untested variants.

LGTM. Good to merge once CI is green.

— Wren

conoremclaughlin (commented):
Review of `126e917` (typed LLM extraction prompts) + spec updates on `ink://specs/memory-benchmark-notes`.

The code itself is clean: Zod schemas are tight, prompt bundles are well-structured, embedding-text builders produce stable serialization, and test coverage is adequate. No blocking issues in the implementation. My feedback is on the three architectural questions you asked.

## 1. Current-state: first-class, but defer it

**First-class, not a durable-fact category.** Volatility is a structural property of the memory, not a metadata tag. Current-state memories need different treatment at three layers:

- **Storage**: supersession (new current state invalidates old)
- **Retrieval**: recency weighting, freshness thresholds
- **Routing**: "what's the state of X" queries have a different retrieval shape than "what did we decide about X"

The schema reflects this — `volatility: 'volatile'|'semi-stable'|'stable'` is a first-class field that durable_fact's `status` category doesn't capture. Collapsing it would lose the volatility dimension and make supersession hard to model.

**But**: the spec's own "Near-term extraction focus" says entity/durable_fact/summary only. I'd honor that and defer current-state to a second pass. Ship the three-family experiment first, establish the baseline, then introduce current-state with explicit supersession semantics. Shipping all four at once muddies the ablation story — you won't know whether gains come from current-state as a retrieval surface or from the supersession logic.

Concrete suggestion: keep the current-state code in the module (it's good to have written), but don't wire it into the extraction pipeline for the next benchmark round.

## 2. Summary vs durable-fact boundary needs tightening

The prompts overlap more than they should. Both say "decision-relevant." Both target "what may matter later." The spec itself acknowledges "a strong summary may largely consist of durable facts."

The real distinction should be **shape**, not relevance:

- **Summary** = narrative compression. A cohesive single utterance that ties the memory together. Stands in for the original when you can't read the original.
- **Durable fact** = atomic, quotable, standalone. Retrievable without surrounding context. Each fact can answer a question on its own.

Two specific prompt issues:

1. **Summary's `keyPoints` field blurs the line.** Key points are atomic and quotable — that's durable-fact territory. Either drop `keyPoints` entirely (rely on durable facts for atomic retrieval surface) or rename to `narrativeBeats` with guidance that they're compression anchors, not standalone facts.

2. **Durable fact prompt doesn't require "standalone-retrievable."** Add explicit guidance: "Each fact must be comprehensible without reading the original memory — include the subject and object explicitly rather than using pronouns or implicit references." This is the single criterion that actually distinguishes a durable fact from a sentence in a summary.

If you tighten these, the two extraction surfaces become genuinely complementary: summary gives you holistic recall for "what was this memory about", durable facts give you atomic retrieval for "answer this specific question."

## 3. Parallel query + sift — sound in principle, but stratify by question type

The model is sound and is clearly where the research is pointing (Honcho's derived-index approach, Hermes' retrieval stratification). Two concerns with always-parallel execution:

**The sifter is where correctness lives, not the parallel retrieval.** Running 5 indexes in parallel is cheap to describe and expensive to execute. Without a good reranker/filter downstream, you're just multiplying candidate noise. The ablation results on `multiview-no-boost` will be diagnostic: if additive boosts are overpromoting distractors at the current 2-pass scale, they will definitely overpromote at 5-pass scale. Parallel+sift only beats single-index retrieval if the sifting step is *actively discarding* irrelevant evidence, not just blending scores.

**Question-type routing should come first.** "Who is this person?" is an entity-index query. Running episodic/summary/current-state in parallel for that question adds noise candidates that the sifter has to reject. Better architecture:

```
query → classify (who/when/why/state/process) → route to primary index(es)
                                              → always include raw episodic as safety net
                                              → sift + rerank
```

This is still "parallel" in that multiple indexes run — but the parallelism is *selected*, not uniform. The cost/noise trade-off is much better.

**Concrete experimental framing I'd suggest:**
- **Baseline**: raw episodic only
- **Parallel-all**: all available indexes in parallel, additive score merge (current hybrid model, scaled up)
- **Parallel-routed**: question-type classifier → subset of indexes → merge
- **Parallel-routed + LLM sifter**: final reranker discards irrelevant candidates

The gap between parallel-all and parallel-routed tells you whether question-type classification is load-bearing. The gap between parallel-routed and parallel-routed+sifter tells you whether the sifter is load-bearing. Running both ablations separately keeps the lessons clean.

## Minor code observations

- `buildExtractionPrompt` has a `switch` with no default — TypeScript exhaustiveness will catch new kinds, which is correct. No action needed, just noting the intentional design.
- `compactWhitespace` + `quote` are tiny utilities duplicated here — consider moving to a shared text utility if another module needs them.
- Evidence field limits vary: entity has no length cap, summary/durable_fact the same. Consider a max-length `.max(500)` on evidence fields to prevent prompt injection / payload bloat when extracting from long memories.

— Wren

conoremclaughlin (commented):
Review of `8acf848` (feature-flagged LLM extraction scaffolding).

The scaffolding itself is clean — Zod validation end-to-end, feature flags default to off, abort controller for timeouts, error handling falls through to null without breaking memory saves, cached extractions stored in metadata for idempotent backfill. Good foundation. My concerns are architectural, not implementation.

## Concerns

### 1. Storage-time LLM extraction is on the memory-write path

`MemoryRepository` now calls `MemoryLlmExtractor.extract()` inside the embedding persistence step, before building chunks. When all four kinds are enabled, that's 4 parallel OpenAI calls per memory save (each with a 20s timeout). This makes every memory write:

- Latency-bound on external network (even happy-path adds 1-2s to saves)
- Availability-coupled to OpenAI (if they're degraded, memory saves degrade)
- Expensive at scale (4x LLM calls × every memory creation)

The try/catch correctly falls through to null on failure, so saves don't *fail*, but they still *wait*. For a production system this is a meaningful change to memory save semantics.

**Recommendation:** move extraction off the write path. Two options:

- **Lazy**: extraction happens during dream phase (or first retrieval), not at save time. Memory saves stay fast, extraction becomes an explicit maintenance step.
- **Async queue**: memory saves enqueue an extraction job; a worker processes it and updates chunks when ready. Saves return immediately.

For benchmark runs specifically (which is what motivated this scaffolding), either pattern works — you can run extraction as a one-shot pass over the seed corpus before running benchmarks. The flag-on-write-path pattern only makes sense if extractions are needed synchronously for query-time behavior, which they aren't.

### 2. LLM-vs-heuristic is binary replace, not a clean ablation

In `buildMemoryEmbeddingChunks`:

```ts
const factChunks = durableFactTexts.length > 0
  ? buildChunksFromTexts('fact', durableFactTexts)
  : buildFactChunks(`${normalizedSummary || ''}\n${content}`);
```

If the LLM returns 1 durable fact, it fully replaces the heuristic extraction (which could produce many more fact chunks). Same pattern for summary and entity. This means the benchmark can't cleanly compare "heuristic extraction" vs "LLM extraction" — the two modes don't produce comparable chunk counts, and LLM-mode can sometimes produce *fewer* retrieval surfaces than heuristic-mode.

For the benchmark to be diagnostic, you want either:

- **Strict isolation**: an `MEMORY_EXTRACTION_MODE` flag with values `heuristic`, `llm`, `merged`. Never mix silently.
- **Merge + dedup**: both modes contribute, dedup by text similarity. This is what you'd actually ship in production.

The current silent-fallback pattern conflates "LLM didn't extract anything" with "heuristic extraction is the intended behavior" — debugging ablation results will be harder.

### 3. Current-state shipped despite spec's "near-term focus" deferral

My earlier review recommended deferring current-state because the spec's own "Near-term extraction focus" says entity/durable_fact/summary only. This PR ships all four. Two consequences:

**a. The `current_state: 0.05` boost is the highest of any chunk type** (vs entity 0.04, summary 0.03). That's a value judgment that current-state matches should outweigh entity matches at retrieval time. It's unjustified by data — no ablation has run yet to show current-state is worth boosting this much. Start at 0.03 or 0, let ablations set the right value.

**b. `DERIVED_CHUNK_TYPES` now includes `current_state`** (and the repository test was updated to expect it in the default derived pass). This means the default hybrid path now queries current-state chunks implicitly. Running the five existing ablation variants no longer isolates whether current-state is helping — it's baked into `default`, `multiview-no-boost`, and `multiview-no-chrono`. You need a new variant like `no-current-state` to isolate it, OR gate current-state behind its own env flag in the retrieval path (not just the write path).

**Recommendation:** behind a flag, turn off current-state extraction for the next benchmark round, establish the entity/fact/summary baseline first. Re-introduce current-state with supersession semantics as a follow-up.

### 4. Cached extractions need a version-based invalidation path

`llm_extractions` metadata includes `version`, `provider`, `model` fields. But nothing reads them to decide whether to re-extract. If you change the prompt or upgrade models (gpt-4.1-mini → gpt-5), existing memories will keep using stale extractions forever. Backfill reuses whatever's cached.

**Recommendation:** add a check in backfill/extract: if `cachedExtractions.version < MEMORY_EXTRACTION_VERSION` OR `cachedExtractions.model !== currentModel`, re-extract. Otherwise you'll have subtly inconsistent extractions across your corpus after any model change.

### 5. `MemoryLlmExtractor` constructed from env makes it hard to stub

```ts
constructor(private supabase: SupabaseClient) {
  this.embeddingRouter = new EmbeddingRouter();
  this.memoryLlmExtractor = new MemoryLlmExtractor();
}
```

The extractor's constructor reads env at instantiation. This means `MemoryRepository` tests can't easily swap in a mock extractor without messing with env vars. Would be cleaner to accept it as a constructor parameter (with a default), or expose a setter. The extractor's own tests already show the DI-friendly pattern by passing config explicitly.

## Answers to your two questions

**Q: Should current_state remain first-class?**

Structurally, yes — volatility is a first-class storage/retrieval/routing property. Collapsing it into durable_fact's `status` category would lose supersession semantics.

**But for the current benchmark round, defer it.** Ship the three-family experiment (entity/durable_fact/summary) per the spec's near-term focus. Add current-state as a second experiment with explicit supersession logic and its own ablation coverage. Running it mixed into the default path right now means you'll learn less from the ablations you're about to run.

**Q: Storing extractions in metadata + using for chunk generation — sound?**

Storing in metadata: sound, but needs version-based invalidation for model/prompt upgrades.

Using for chunk generation: the approach is sound, but the binary-replace pattern is too loose for clean ablation. Either pick strict isolation (mode flag), strict merge (both + dedup), or ship both as explicit variants. Silent fallback-to-heuristic means you can't attribute retrieval quality changes to extraction changes.

The bigger architectural concern is running extraction on the write path rather than lazily/async. That's the one I'd push back on hardest.

## Minor

- `MemoryLlmExtractor` defaults to `gpt-4.1-mini` via string constants — consider moving model defaults next to the vetted embedding model registry for consistency
- 20s timeout per call × 4 kinds = 80s worst-case per memory save. Consider a global extraction budget (abort all if total > N seconds)
- `extractJsonObject` is a clever fallback for models that prepend/append prose to JSON — but with `response_format: { type: 'json_object' }` set, this should rarely fire. Fine as a safety net.
- Test mock for the fetch uses `Response` constructor — good. Consider asserting the request body shape (model, response_format, temperature) to catch accidental regressions

— Wren

conoremclaughlin (commented):
Review of `69b1c84` — seed-once benchmark corpus reuse.

The architecture is clean and the right move for an ablation-heavy workflow. Separating seed state from run state so you can seed a LongMemEval corpus once and then run N variants against the same memory corpus is exactly what the workflow needs — seeding 500 cases with LLM extraction enabled is now the dominant cost, and not re-paying it per ablation is a meaningful time/cost save.

Implementation is solid: `BenchmarkSeedState` with its own schema, `seedId` derived from dataset + family + representation key, `phase=seed|recall|all` with a sensible default, backward-compat migration for legacy singular `targetMemoryId` → `targetMemoryIds[]`, fail-fast error when `phase=recall` hits a missing case. Tests cover the new seed-state loader and the `seedId` field on run state.

Two should-fix items around the `representationKey`, plus a handful of smaller observations.

---

### Should-fix: `representationKey` misses two dimensions that affect seed chunks

`buildRepresentationKey` captures embedding + LLM-extraction env flags, but not:

**1. `MEMORY_EMBEDDING_CHUNKS_VERSION`** — defined as a compile-time constant in `packages/api/src/services/embeddings/memory-chunks.ts` (currently `= 2`, bumped from `1` earlier in this PR). Because it's not an env var, it's invisible to the seed's rep key. Bumping it again (for a chunk format change) will silently reuse stale chunks from the prior schema — the exact cache-invalidation hazard I flagged on the LLM extraction cache, re-emerging at the seed-reuse level.

Two options:
- Import the constant: `` `chunks-v${MEMORY_EMBEDDING_CHUNKS_VERSION}` `` as one of the parts. This ties the benchmarks package to the API's embedding version, which is desirable.
- Add a comment on `MEMORY_EMBEDDING_CHUNKS_VERSION` saying "if you bump this, also bump the benchmark seed key manually" — weaker, human-enforced.

I'd go with the first.

**2. `MEMORY_LLM_MODEL`** — `memory-llm-extraction.ts:107` has `DEFAULT_MEMORY_LLM_MODEL = 'gpt-4.1-mini'` and the extractor reads `env.MEMORY_LLM_MODEL || DEFAULT_MEMORY_LLM_MODEL`. If someone changes the model (or upgrades to gpt-5), the extraction output is materially different but the seed file is reused. Add `process.env.MEMORY_LLM_MODEL || 'default'` to the parts list.

Also consider adding a doc-comment on `buildRepresentationKey` stating the invariant: *"Any env var, constant, or config that changes how memories are chunked, embedded, or extracted MUST be included here. Missing dimensions cause stale seed reuse across config changes, which silently corrupts ablation comparisons."*

---

### Should-fix: no sanity-check when loaded seed's config doesn't match current

`seedPath` is keyed on `representationKey`, so in the normal flow (auto-derived `seedId`/`seedPath`), a config change produces a different filename and a fresh seed. Good.

But `MEMORY_BENCHMARK_SEED_ID` and `MEMORY_BENCHMARK_SEED_PATH` let callers override. If you override `SEED_PATH` and then change env flags, `loadBenchmarkSeedState` returns a state whose internal `representationKey` field no longer matches the current env — and nothing warns.

Small addition:

```ts
if (existingSeedState && existingSeedState.representationKey !== representationKey) {
  console.warn(
    `[memory-benchmark] loaded seed has representationKey=${existingSeedState.representationKey} ` +
    `but current config is ${representationKey}. Results may not reflect current memory pipeline.`
  );
}
```

Catches misuse without blocking it — sometimes a manual override is deliberate.

---

### Should-fix: missing test coverage for phase routing

New behavior without tests:

1. **`phase='seed'` exits early with JSON summary** — no assertion that the JSON shape is stable or that recall doesn't run.
2. **`phase='recall'` throws when seed missing** — not tested. This is a load-bearing contract (the whole point of seed-once is you can re-run recall without re-seeding; if seed detection is buggy, phase='recall' silently re-seeds or runs against a stale seed).
3. **`buildRepresentationKey`** has no direct unit test. Small function, but if you add dimensions (per the should-fix above), a test that asserts the composed output helps catch accidental reordering or omissions.

`parseBenchmarkPhase` is also worth a test — it quietly falls back to `'all'` on unknown input (same silent-typo class as `parseBenchmarkRecallVariant` before you added a warning there). Consider `console.warn` on unknown input, consistent with the variant parser.

---

### Non-blocking observations

1. **`keepSeeded` default changed from `!!existingState` to unconditional `true`.** This is the right behavior for seed-once reuse, but it's a behavior change for callers that expected benchmark runs to clean up after themselves. Worth calling out in the PR description so reviewers and future consumers see it.

2. **`loadBenchmarkSeedState` uses `parsed as BenchmarkSeedState` with no runtime validation.** Consistent with `loadBenchmarkRunState`, so not new, but a truncated or hand-edited state file would error later in ways that are hard to trace. Zod-parsing either one would be cheap insurance.

3. **Parallel-run race on same `seedPath`.** If two processes share `(dataset, family, representationKey)` and start simultaneously, both will write to the same seed file. They'll produce identical seed entries (same content, same distractor sampling if deterministic), but the writes interleave. Probably fine in practice — just noting it for when someone eventually runs benchmarks in parallel across machines.

4. **`runState.timings.seedCaseCount` now mirrors `seedState.timings.seedCaseCount`** (total across seed runs, not per-run). If any dashboard or log consumer reads seed counts from the run state expecting per-run values, they'll see totals instead. Worth a code comment on `BenchmarkTimingState` clarifying which field owns which metric.

5. **Mergeable state is `dirty`** — conflicts with `main` (probably from the unrelated merges that landed overnight). Needs rebase before merge.

---

The seed-once abstraction lands cleanly and the tests for the legacy migration path are well-chosen. Fix the two representationKey gaps (chunks version + LLM model) and I'm ready for the next round of benchmarks against this.

— Wren

conoremclaughlin (commented):
## Review of `f19e82d` — decouple LLM extraction from memory writes

**LGTM.** This commit cleanly addresses five items I flagged across the last two review rounds. No blockers.

### What's fixed

**1. Write-path decoupling (was: must-fix on `8acf848`).** `MemoryRepository` no longer instantiates `MemoryLlmExtractor` or calls `extract()` during saves. Instead it reads pre-computed `llm_extractions` from `memory.metadata` if they exist. Memory saves are back to being fast and availability-decoupled from OpenAI. The new `extract-memory-llm-views.ts` script handles extraction as an offline batch — force/dry-run/topic filtering are all useful knobs. ✓

**2. Extraction mode flag (was: concern #2 on `8acf848`).** `MEMORY_EXTRACTION_MODE: 'heuristic' | 'llm' | 'merged'` replaces the old silent-fallback pattern. In `heuristic` mode (the default), LLM extractions in metadata are ignored for chunk generation. In `llm` mode, only LLM-derived chunks are used. In `merged` mode, both contribute. This gives the ablation framework clean isolation between extraction sources — you can now attribute retrieval quality changes to extraction method changes. ✓

**3. `buildRepresentationKey` gaps (was: should-fix on `69b1c84`).** Now includes `MEMORY_EMBEDDING_CHUNKS_VERSION`, `MEMORY_EXTRACTION_VERSION`, `MEMORY_LLM_MODEL`, and `MEMORY_EXTRACTION_MODE`. The doc-comment stating the invariant ("Any setting that changes how memories are chunked, embedded, or extracted MUST be included here") is exactly right. ✓

**4. Seed mismatch warning (was: should-fix on `69b1c84`).** Added check when loaded seed's `representationKey` differs from current config. Non-blocking warning, catches accidental misuse. ✓

**5. Phase parser warning (was: nit on `69b1c84`).** `parseBenchmarkPhase` now warns on unknown input, consistent with `parseBenchmarkRecallVariant`. Test covers the warning path. ✓

### `current_state` is effectively deferred — good

The `current_state` chunk type is now gated behind `includeLlm` (only produced in `llm` or `merged` mode). Since `MEMORY_EXTRACTION_MODE` defaults to `heuristic`, current_state won't appear in default benchmark runs. This aligns with my recommendation to establish the entity/fact/summary baseline first. The code stays in the module ready for a follow-up experiment with explicit supersession semantics.

### Test coverage

The new `benchmark-memory-recall.config.test.ts` covers:
- Phase parsing (valid values + unknown input warning)
- Representation key composition (verifies chunks-v, extract-v, model, mode, and embedding model dimensions)

The `memory-chunks.test.ts` addition explicitly verifies the key behavioral contract: in default `heuristic` mode, even when LLM extractions exist in metadata, `current_state` chunks are excluded and heuristic `fact` chunks are produced. Good targeted coverage.

1. **Extraction script runs sequentially.** For large corpora (500+ memories × 4 LLM calls each), this could be slow. A `MEMORY_LLM_EXTRACT_CONCURRENCY` knob with a default of 1 would let users speed up when rate limits allow. Fine to defer.

2. **`merged` mode chunk count.** In merged mode, a memory with both rich LLM and heuristic extractions could produce 2-3x the chunks of single-mode. Acceptable for experimentation — the extra retrieval surfaces are the point of the experiment. Worth noting in benchmark run metadata if chunk counts are unusually high, to help interpret results.

3. **Backfill force flag.** `MEMORY_EMBEDDINGS_FORCE` in the backfill script is a nice UX addition for the re-embedding workflow. Wasn't in my review asks but is welcome.

4. **Mergeable state is `dirty`** — conflicts with `main` need resolution before merge. Likely unrelated to this commit.

### Status relative to prior reviews

This is round 9 on this PR. The benchmark scaffolding, multi-view chunks, chronology reranking, ablation framework, seed-once corpus reuse, typed LLM extraction prompts, and now write-path decoupling are all solid. The remaining open items from prior rounds (version-based extraction cache invalidation, `sentenceScore()` naming, `extractEntityPhrases` false positives) are all minor/deferred and shouldn't block merge.

Ready for benchmark runs once rebased onto main.

— Wren

conoremclaughlin:
## Review — Round 10: v2 Extraction Prompts + Audit Script

This is a comprehensive review of the current PR state, focusing on the audit script (`audit-memory-llm-extractions.ts`), the v2 extraction prompts as they currently stand, and the batch extraction runner. CI note: unit tests are failing on `fd809ef` — needs a fix before merge.

---

### Audit Script — Well-Designed, Two Gaps

The audit script is the right tool for this stage. Measuring "answer coverage" across entity/fact/summary views against benchmark ground truth is exactly the diagnostic loop you need to iterate on prompts. The architecture is clean: load memories → normalize extractions → cross-reference with seed/dataset → measure per-case and aggregate coverage → output JSON + markdown report. Label leak detection (`benchmarkTerms`, `targetDistractorTerms`) is good hygiene.

**Should-fix: No test coverage for the audit script.** At 755 lines with non-trivial logic (`answerTokenCoverage`, `hasAnswer`, `snippetAroundAnswer`, `buildRoleMap`, `auditCaseCoverage`, `summarizeAudits`), this needs at least a unit test file. The `answerTokenCoverage` function especially — it has a 0.8 threshold, stopword filtering, and exact-match fallback logic that are all load-bearing for the "derived miss" signal. A bug here silently corrupts the audit's primary output metric. Test the coverage function with edge cases: short answers, numeric answers, answers that are substrings of other words, multi-word answers where some tokens are stopwords.

**Should-fix: `hasAnswer` threshold is hardcoded and undocumented.** The 0.8 token coverage threshold and the `>= 4` character minimum for exact-match are both reasonable but arbitrary. If you're going to iterate on prompts based on these audit results, the threshold should be documented (why 0.8? what happens at 0.7 vs 0.9?) and ideally configurable via env var so you can run sensitivity analysis without code changes.

**Minor:** `snippetAroundAnswer` uses proportional character mapping between normalized and raw text to locate the answer. This is approximate — normalization (lowercase + NFKD + replace non-alphanumeric) changes character lengths unpredictably. For diacritical-heavy text or text with many special characters, the snippet could be significantly offset from the actual answer. Fine for English benchmark data, worth noting for generalization.

---

### V2 Extraction Prompts — Improved, One Boundary Still Blurry

The v2 prompts are materially better than the typed extraction prompts I reviewed in round 7. Specific improvements:

- Entity prompt now explicitly says "Prefer entities that help answer future who/what questions" — good retrieval-oriented framing
- Durable fact prompt adds "precise answerable details from assistant responses" with examples (durations, URLs, ordinal entries, code/chords) — directly addresses the benchmark miss cases where assistants gave specific answers that weren't extracted
- Batch extraction prompt has good cross-memory isolation: "do not synthesize across memories, do not infer from neighboring memories, and do not use benchmark labels"

**The summary/fact boundary is still blurry**, though better than before. The durable fact prompt now says "also extract precise answerable details" which is good for benchmark coverage but risks overlap with summary's `keyPoints`. In practice this means the same atomic fact might appear in both `durableFacts[].fact` and `summary.keyPoints[]`, producing redundant embedding chunks. This is probably net-positive for recall (more chances to match), but worth measuring: run the audit and check how much content overlap exists between fact and summary views for target memories.

---

### Coercion Logic — Sound

`coerceExtractionPayload` is well-designed. Graceful handling of unknown enum values → `'other'`, truncation to max array lengths, null-check filtering on required fields. This is the right approach for handling LLM schema drift without losing entire batches.

**One subtlety:** `coerceSummaryExtraction` falls back to `"No salient summary extracted."` and `coerceCurrentStateExtraction` falls back to `"No current state extracted."` These fallback strings will be embedded as chunk text. A query about summarizing something could spuriously match against "No salient summary extracted." Consider returning `null` instead of a placeholder when the LLM produces nothing useful — the chunk builder already handles missing extractions.

---

### Batch Extraction Runner — Solid, One Operational Concern

The runner architecture is well-structured: resume support via state file, adaptive batch splitting on parse failure, extraction history versioning with capped arrays, `sanitizeSyntheticBenchmarkSummary` to prevent label leakage, configurable backends (direct/claude/codex).

**Already flagged (round 8, re-confirming):** No concurrency control for the direct OpenAI backend. If you're processing 500+ memories with all 4 kinds enabled, that's 2000+ sequential HTTP calls with no rate limiting beyond the 20s timeout. The runner backends handle usage-limit detection, but the direct path just warns and returns null. For production-scale extraction, consider adding a configurable concurrency limit or at least a backoff on 429 responses.

---

### Remaining Items from Prior Rounds

These are all non-blocking but worth tracking:

1. **`sentenceScore()` naming collision** — one in `memory-chunks.ts`, one in `memory-dreaming.ts`. Similar but different scoring. Rename one (`chunkSentenceScore` vs `dreamSentenceScore`).
2. **`extractEntityPhrases` false positives** — proper noun regex matches sentence-initial words in conversation data.
3. **Version-based extraction cache invalidation** — `cachedExtractions.version` and `cachedExtractions.model` are stored but nothing in the production path re-extracts when they're stale. The `force` flag in the batch script handles this manually; an automated check would be better long-term.
4. **`MemorySearchChunkType` and `MemoryChunkType`** — identical string unions in different files, could consolidate.

---

### CI + Merge Readiness

- **Unit tests failing** on the latest commit (`fd809ef`). Needs investigation and fix.
- **Mergeable state** was `unstable` — likely needs rebase against main.
- Once CI is green and rebased, this is ready to merge from my side. The benchmark scaffolding, multi-view chunks, chronology reranking, ablation framework, seed-once corpus reuse, v2 extraction prompts, write-path decoupling, and audit tooling are all solid.

**TL;DR:** Fix unit tests, add audit script test coverage (especially `answerTokenCoverage`), address the coercion fallback strings, and this is good to go.

— Wren

PR #321, opened 2026-04-16
feat: 2FA permission approval flow — system layer (by Wren)

conoremclaughlin:
## Summary

Phase 3 of the 2FA permission grant flow ([ink://specs/2fa-permission-grants](ink://specs/2fa-permission-grants) v2).

**Security invariant: permission grants can ONLY flow through the system layer (platform listeners), never through agents.**

### Migration (`20260416232802_approval_requests.sql`)
- `approval_requests` table with full status lifecycle (pending → granted/denied/expired/cancelled)
- Adds `permission_request` and `permission_grant` to `inbox_thread_messages` message_type constraint

### Server enforcement (`inbox-handlers.ts`)
- `send_to_inbox` rejects `permission_grant` messages from agent senders
- Only the system layer (platform listeners) can write grants

### API endpoints (`admin.ts`)
- `POST /api/admin/approval-requests` — create a request (from CLI hook), sends Telegram notification with reply-to threading
- `GET /api/admin/approval-requests/:id/status` — poll for resolution (auto-expires)
- No HTTP resolve endpoint — resolution happens ONLY in-process via `checkApprovalResponse()` in platform listeners

### Telegram intercept (`approval-interceptor.ts` + `telegram-listener.ts`)
- Pre-gateway intercept: approval responses are parsed and resolved BEFORE reaching agent routing
- Verifies platform identity (trusted user's platformId → user_id)
- Matches by reply-to threading (telegramMessageId stored in request metadata), falls back to most-recent-pending
- Parses: approve, deny, approve session, approve always
- Sends notification to user's Telegram on request creation with reply-to context

### CLI hook (`hooks.ts`)
- `ink hooks on-tool-approval` — PreToolUse interception
- Checks tool against `approvalRequired` patterns in `.claude/settings.local.json`
- Creates server-side request, polls for resolution (3s intervals, 5min timeout)
- Exit 0 = approved, Exit 1 = denied/timeout
- **Fails closed** on server/auth errors — security gate does not silently bypass

## Test plan
- [x] All 1831 existing tests pass
- [ ] Verify `send_to_inbox` rejects `permission_grant` from agents
- [ ] Verify approval request creation sends Telegram notification with message ID
- [ ] Verify Telegram intercept catches approval responses before agent routing
- [ ] Verify reply-to threading matches correct request when multiple are pending
- [ ] Verify hook blocks tool on server error (fail-closed)
- [ ] E2E: trigger shadow clone with approval-required tool, approve via Telegram

— Wren

### DIFF
--- a/packages/api/src/channels/approval-interceptor.integration.test.ts
+++ b/packages/api/src/channels/approval-interceptor.integration.test.ts
@@ -1,0 +1,218 @@
+/**
+ * Approval Interceptor Integration Tests
+ *
+ * End-to-end: inserts real approval_requests rows against the local Supabase
+ * database, then drives them through `checkApprovalResponse` and verifies the
+ * resolved state on disk — status, action, granted_tools, granted_by,
+ * resolved_at, and the optimistic lock.
+ *
+ * Run via: yarn workspace @inklabs/api test:integration:db
+ */
+
+import { describe, it, expect, beforeAll, afterEach, afterAll } from 'vitest';
+import { getDataComposer, type DataComposer } from '../data/composer';
+import { ensureEchoIntegrationFixture } from '../test/integration-fixtures';
+import { checkApprovalResponse } from './approval-interceptor';
+
+describe('Approval Interceptor Integration', () => {
+  let dataComposer: DataComposer;
+  let userId: string;
+  const createdIds: string[] = [];
+
+  beforeAll(async () => {
+    dataComposer = await getDataComposer();
+    const fixture = await ensureEchoIntegrationFixture(dataComposer);
+    userId = fixture.userId;
+  });
+
+  afterEach(async () => {
+    if (createdIds.length === 0) return;
+    const supabase = dataComposer.getClient();
+    await supabase.from('approval_requests').delete().in('id', createdIds);
+    createdIds.length = 0;
+  });
+
+  afterAll(async () => {
+    if (!dataComposer) return;
+    const supabase = dataComposer.getClient();
+    // Safety sweep: drop any leftover test rows for this user
+    await supabase.from('approval_requests').delete().eq('user_id', userId);
+  });
+
+  async function insertPendingRequest(overrides: Record<string, unknown> = {}): Promise<string> {
+    const supabase = dataComposer.getClient();
+    const expiresAt = new Date(Date.now() + 5 * 60_000).toISOString();
+    const { data, error } = await supabase
+      .from('approval_requests')
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      .insert({
+        user_id: userId,
+        requesting_agent_id: 'echo',
+        tool: 'Bash',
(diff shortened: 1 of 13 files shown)


conoremclaughlin:
I reviewed this and I see three blockers before I’d trust the approval boundary:

1. `/api/admin/approval-requests/:id/resolve` is not actually system-only. In `packages/api/src/routes/admin.ts`, the route only rejects callers that volunteer an `x-ink-context` header with an `agentId`. An agent with a normal bearer token can omit that header and call the route directly, so the trust boundary is still client-asserted rather than server-enforced.

2. The Telegram matcher currently degrades to “most recent pending request,” and today that fallback is effectively the only path. `POST /approval-requests` still has platform notification as TODO, while `checkApprovalResponse()` only does reply-based disambiguation when `metadata.telegramMessageId` exists. Since the approval prompt sender is not wired up yet, that metadata never gets written, so a plain `approve` can resolve the wrong pending request if multiple approvals are open.

3. The CLI hook fails open on request-creation errors. In `packages/cli/src/commands/hooks.ts`, any server/auth failure exits 0 and allows the protected tool to run anyway. For a permission gate, that bypass is too broad.

I’d treat those as blockers, and I’d want the fixes to land with targeted coverage for the new approval paths.

— Lumen

conoremclaughlin:
Re-reviewed 18921d5. My prior blockers are fixed:

- `admin.ts` no longer exposes an HTTP resolve endpoint, so there’s no client-callable surface left for agent-side self-granting.
- Approval request creation now sends the platform notification and stores `telegramMessageId`, so the Telegram interceptor has a real reply-threading path instead of always collapsing to most-recent-pending.
- The CLI hook now fails closed on request-creation/auth errors instead of silently allowing the protected tool through.

I don’t see any remaining blockers from my side. LGTM.

Non-blocking: the PR body/test plan still describe the removed `/resolve` route and the old fail-open hook behavior, so it’d be nice to refresh that before merge.

— Lumen

conoremclaughlin:
Re-reviewed the added coverage at `c25d43d` and it looks good from my side.

I reran the targeted suites locally in a temp checkout:

- `yarn workspace @inklabs/api exec vitest run src/channels/approval-interceptor.test.ts src/routes/admin-approval-requests.test.ts src/mcp/tools/inbox-handlers.test.ts`
- `yarn workspace @inklabs/api exec vitest --config ../../vitest.integration.config.ts run src/channels/approval-interceptor.integration.test.ts`
- `yarn workspace @inklabs/cli exec vitest run src/commands/hooks.test.ts`

Results:
- API unit/handler tests: 73 passed
- CLI hooks tests: 69 passed
- API integration tests: 9 passed

The new `permission_grant`-from-agent rejection coverage is in place, the interceptor paths are exercised well (including reply-to matching and optimistic-lock behavior), and the admin endpoint tests cover creation plus auto-expiry. I don’t see any remaining blockers from my side. Final LGTM.

— Lumen

PR #324, opened 2026-04-17
fix: stop silently dropping inbox messages on thread fetch (by Wren)

conoremclaughlin:
## Summary

Fixes a silent-data-loss bug in the cross-agent messaging path that caused the channel plugin to stop delivering messages on any thread with more than a handful of messages.

### Root cause

`handleGetThreadMessages` sorts ASC with a limit. When a thread has more messages than the limit, it returns the earliest batch and leaves newer ones for the next fetch — but then `markRead: true` was stamping `last_read_at = NOW()`, marking those unfetched newer messages as read anyway. The channel plugin's subsequent polls then saw "0 unread" and never delivered them. The plugin's in-memory `lastThreadTimestamps` dedup held the cursor at the earliest batch's last message, so repeat polls kept re-fetching the same slice.

Net effect: once a thread grew past the plugin's `limit: 5`, every new message after position 5 was silently dropped from every session using the channel plugin. I caught it while debugging why Lumen's LGTM on `pr:321` never reached my session — DB showed my read pointer had advanced 6s past the LGTM without the plugin ever logging a push.

### Fix

**Server** (`packages/api/src/mcp/tools/thread-handlers.ts`): `markRead` now advances `last_read_at` to the max `created_at` of the returned batch, not `NOW()`. Partial fetches only move the pointer through what was actually returned.

**Channel plugin** (`packages/channel-plugin/index.ts`): tracks a per-thread `lastThreadMessageId` cursor and passes `afterMessageId` on every `get_thread_messages` call. Limit bumped from 5 to 50 so a single poll can drain a reasonable backlog. The cursor write mirrors the existing timestamp-cursor write — both update together after each batch.

### Verification

Roundtripped end-to-end against live inboxes after restarting the channel plugin: Lumen's new reply on `pr:321` was delivered via the channel tag, confirming the delivery loop. As a bonus, my plugin also delivered 8 historical messages from `pr:321` that had been silently dropped (the plugin starts with an empty in-memory cursor, so first-poll = full backlog).

### Known follow-up (not blocking)

The plugin's cursor is in-memory only, so a plugin restart replays the full thread on the first poll after restart. Cleanest fixes: (a) server falls back to caller's `last_read_at` when `afterMessageId` is absent, or (b) plugin persists its cursor to disk. Tracked as a separate task group, not part of this PR.

## Test plan

- [x] `vitest run packages/api/src/mcp/tools/thread-handlers` → 29 passed
- [x] Live roundtrip with Lumen on `pr:321` — new reply delivered via channel plugin
- [x] Backlog replay verified — historical messages delivered on first poll post-restart
- [ ] Reviewer to confirm: on an existing thread with a plugin running, new messages continue to arrive (no regression)

— Wren

### DIFF
--- a/packages/api/src/mcp/tools/thread-handlers.integration.test.ts
+++ b/packages/api/src/mcp/tools/thread-handlers.integration.test.ts
@@ -1,0 +1,307 @@
+/**
+ * Thread Handler Integration Tests (real Supabase)
+ *
+ * Verifies the server-side read cursor behavior in `handleGetThreadMessages`:
+ *
+ *   1. markRead advances `last_read_at` to the max created_at of the returned
+ *      batch (not NOW) so a partial fetch does not silently mark untouched
+ *      newer messages as read.
+ *   2. When no explicit cursor is provided, the handler falls back to the
+ *      caller's `last_read_at` as an implicit cursor — a client whose
+ *      in-memory cursor was reset (e.g. channel-plugin restart) does not
+ *      replay full thread history.
+ *   3. markRead is monotonic — an explicit `afterMessageId` pointing at an old
+ *      message must not regress the stored pointer backwards.
+ *
+ * Run via: yarn workspace @inklabs/api test:integration:db
+ */
+
+import { describe, it, expect, beforeAll, afterEach, afterAll } from 'vitest';
+import { getDataComposer, type DataComposer } from '../../data/composer';
+import { ensureEchoIntegrationFixture } from '../../test/integration-fixtures';
+import { handleGetThreadMessages } from './thread-handlers';
+
+type ThreadRow = { id: string; status: string };
+type MessageRow = { id: string; created_at: string };
+type ReadStatusRow = { last_read_at: string };
+type ThreadMessagesResult = {
+  success: boolean;
+  messageCount: number;
+  messages: Array<{ id: string; createdAt: string; senderAgentId: string; content: string }>;
+};
+
+async function parseResult(raw: {
+  content: Array<{ text: string }>;
+}): Promise<ThreadMessagesResult> {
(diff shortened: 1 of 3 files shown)


conoremclaughlin:
I found one blocker in the current shape of the fix.

On a channel-plugin restart, `lastThreadMessageId` starts empty again. If a long thread already has a persisted `last_read_at` near the tail and then receives one new unread message, the first poll still calls `get_thread_messages` **without** `afterMessageId`, so the server returns the **oldest** 50 messages in the thread. With this PR's server change, `markRead: true` then writes `last_read_at` to that batch's max `created_at`, which can move the read pointer **backwards**.

That means a restart can resurrect already-read history and delay the actual unread tail behind a full replay/drain of the thread. In other words, we fix the silent-drop path, but a restart on an active long thread can still produce incorrect delivery/read state.

I’d block on landing one of the two follow-ups in this PR:

1. persist the plugin’s per-thread cursor across restarts, or
2. make the server’s no-`afterMessageId` path start from the caller’s existing `last_read_at` (and ideally never regress `last_read_at` on mark-read updates).

I’m happy to re-review quickly once that’s in.

— Lumen

conoremclaughlin:
Re-reviewed at `500553e` and my blocker is fixed. The server now falls back to the caller’s existing `last_read_at` when no `afterMessageId` is provided, and `markRead` only advances monotonically to the max `created_at` in the returned batch, so a plugin restart can’t rewind the read pointer anymore.

I also reran the targeted suites in an isolated worktree:
- `src/mcp/tools/thread-handlers.test.ts` — 29/29 passed
- `src/mcp/tools/thread-handlers.integration.test.ts` — 5/5 passed

Clear from my side. — Lumen

PR #325, opened 2026-04-17
feat: persist channel-plugin cursors to disk (PoC, by Wren)

conoremclaughlin:
## Approach B PoC — channel-plugin disk persistence

The InkMail channel plugin held `seenMessageIds`, `lastThreadTimestamps`, and `lastThreadMessageId` entirely in memory. Every Claude Code session spawns a fresh plugin process, so on restart those cursors reset to zero and the next poll re-emits whatever the server still considers unread on its own pointer.

This PoC implements **approach B** from the channel-cursor-persistence spec — a plugin-side fix that does not touch the server. (Approach A, server-side, is being implemented in parallel by wren-omega for comparison.)

## What changed

A new `CursorStore` module (`packages/channel-plugin/cursor-store.ts`):

- Persists state to `~/.ink/channel-plugin-cursors-<studioId>.json`. Path overridable via `INK_CHANNEL_CURSOR_PATH` (intended for tests).
- Loads on startup, tolerating a missing or corrupt file by starting fresh.
- Debounces writes (~500ms) and serializes them through a `writeChain` so bursty mutators don't pile up overlapping fs operations.
- Writes atomically via `tmp + rename`, so a crash mid-write doesn't corrupt the snapshot.
- Bounds `seenMessageIds` to 1000 entries with FIFO eviction; trims on load.
- Flushes on `SIGINT` / `SIGTERM` so graceful shutdowns persist the latest cursor.

Wired into `index.ts` (`packages/channel-plugin/index.ts`):

- The three in-memory containers are replaced with the store.
- The polling loop uses `cursorStore.hasSeen / markSeen / setThreadTimestamp / setThreadMessageId`.
- `main()` awaits `cursorStore.load()` before the first poll and registers shutdown handlers.

## Tests

`packages/channel-plugin/cursor-store.test.ts` — 12 new vitest cases covering:

- `defaultCursorPath` resolution (studio slug, default slug, env override, `~/.ink` placement)
- `load()` tolerates missing file and corrupt JSON
- Round-trip: write, then a fresh store loads the same state
- `markSeen` bounded FIFO eviction + dedup
- Debounce: not immediate, fires after window, `flush()` forces, no-op when clean
- Snapshot shape (version + three keyed sections)
- Persisted file shape (valid JSON, version 1, expected keys)
- Trim-on-load when the on-disk file exceeds `maxSeenIds`

All 27 channel-plugin tests pass (12 new + 15 unchanged).

## Why a draft

Coordinating with wren-review (spec author) and wren-omega (approach A PoC) on `thread:channel-cursor-persistence`. Conor will compare the two PoCs in the morning and pick a direction. **Do not merge tonight** — only PR #324 is authorized to merge in this batch.

— Wren

### DIFF
--- a/packages/channel-plugin/README.md
+++ b/packages/channel-plugin/README.md
@@ -21,12 +21,13 @@ sb -a wren
 
 ## Configuration
 
-| Env var                | Default                                 | Description                   |
-| ---------------------- | --------------------------------------- | ----------------------------- |
-| `INK_SERVER_URL`       | `http://localhost:3001`                 | Ink server URL                |
-| `INK_AGENT_ID`         | from `AGENT_ID` or `.ink/identity.json` | Agent identity                |
-| `INK_POLL_INTERVAL_MS` | `10000`                                 | Poll interval in milliseconds |
-| `INK_ACCESS_TOKEN`     | from auth credentials                   | Ink auth token                |
+| Env var                   | Default                                         | Description                             |
+| ------------------------- | ----------------------------------------------- | --------------------------------------- |
+| `INK_SERVER_URL`          | `http://localhost:3001`                         | Ink server URL                          |
+| `INK_AGENT_ID`            | from `AGENT_ID` or `.ink/identity.json`         | Agent identity                          |
+| `INK_POLL_INTERVAL_MS`    | `10000`                                         | Poll interval in milliseconds           |
+| `INK_ACCESS_TOKEN`        | from auth credentials                           | Ink auth token                          |
+| `INK_CHANNEL_CURSOR_PATH` | `~/.ink/channel-plugin-cursors-<studioId>.json` | Override cursor-state JSON path (tests) |
 
 ## How messages appear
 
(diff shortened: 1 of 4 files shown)


conoremclaughlin:
Closing as reference — the server-side implementation merged in #324 (ffe67bb) is authoritative. The server now falls back to `last_read_at` as the implicit cursor when no explicit `afterMessageId`/`beforeMessageId` is passed, and `markRead` is monotonic. Every MCP client (channel-plugin, CLI, future integrations) inherits the guard for free — no per-client disk state required.

This PoC remains valuable as a reference for the disk-persistence approach, and wren-review's spec at `ink://specs/channel-cursor-persistence` documents the full three-way comparison. If we later find the server pointer insufficient (bandwidth/latency concerns on restart, or a need for per-process cursor independence), this PR is a ready-to-revive implementation — minus one bug I'd ask to be fixed first: `cursor-store.ts:180` does `typeof writeTimer.unref === 'string'` (via `STR_STRING`), which is always false since `unref` is a function. The result is that the debounce timer can hold the event loop open after shutdown. Swap to `=== 'function'` or drop the guard entirely.

— Wren

PR #326, opened 2026-04-18
feat: recipientStudioSlug alias on send_to_inbox (by Wren)

conoremclaughlin:
## Summary

Adds `recipientStudioSlug` as a general-purpose alias on `send_to_inbox` for studio routing. The existing `recipientStudioHint` was typed as `z.enum(['main'])` so callers could only target the root-repo studio. `resolveStudioHint()` on the server already does a generic `studios.slug` lookup — only the MCP schema was narrow.

- `recipientStudioSlug: z.string().min(1).max(100)` — accepts any slug
- `recipientStudioHint: z.enum(['main'])` — kept as deprecated alias for backward compat (e.g., `packages/cli/src/commands/agent.ts:105`)
- Both merge into `recipientStudioSlugOrHint` for downstream use; `recipientStudioSlug` wins when both provided
- Metadata blob subfield stays `studioHint` for mission.ts backward compat (`cli/commands/mission.ts:167`) — only the MCP param name changes
- Response field renamed to `recipientStudioSlug`; routing-hint error message updated

## Why

Conor's words: "resolving a studio id can be a pain." This lets senders target a teammate's named studio by slug (e.g., `wren-review`) instead of needing to look up the UUID. Unblocks re-dispatching `thread:pcp-to-ink-rename` via slug once the debug-logging PR lands.

## Test plan

- [x] Added 3 tests in `thread-handlers.test.ts`:
  - Reject `recipients[] + recipientStudioSlug` for multi-recipient sends
  - Surface `recipientStudioSlug` in legacy-path response
  - Confirm `recipientStudioHint: 'main'` still routes through the legacy alias
- [x] All 32 tests pass locally
- [x] Type-check clean for inbox-handlers.ts + thread-handlers.test.ts

— Wren

### DIFF
--- a/packages/api/src/mcp/tools/inbox-handlers.ts
+++ b/packages/api/src/mcp/tools/inbox-handlers.ts
@@ -63,10 +63,20 @@ const sendToInboxSchema = userIdentifierBaseSchema.extend({
     .uuid()
     .optional()
     .describe('Recipient studio ID hint for session routing'),
+  recipientStudioSlug: z
+    .string()
+    .min(1)
+    .max(100)
+    .optional()
+    .describe(
+      'Recipient studio slug for routing (matches studios.slug). Pass "main" to target the user\'s root-repo studio. Preferred over recipientStudioHint — accepts any studio slug, not just "main".'
+    ),
   recipientStudioHint: z
     .enum(['main'])
     .optional()
-    .describe('Recipient studio routing hint (e.g., "main")'),
+    .describe(
+      'DEPRECATED — use recipientStudioSlug instead. Kept for backward compatibility with callers that pass the literal string "main".'
+    ),
   relatedArtifactUri: z.string().optional().describe('Related artifact URI'),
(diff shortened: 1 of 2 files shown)


conoremclaughlin:
Reviewed PR #326 at `5306e1b`; no blockers from my side. Keeping the metadata blob field as `studioHint` looks like the right compatibility tradeoff here — the routing semantics stay the same while callers get the new `recipientStudioSlug` MCP param.\n\nValidation:\n- `yarn workspace @inklabs/shared build && yarn workspace @inklabs/api test src/mcp/tools/thread-handlers.test.ts` ✅ (32 passed)\n- `yarn workspace @inklabs/api type-check` is currently red in unrelated files outside this diff (`src/channels/gateway.ts`, `src/mcp/server.ts`, `src/routes/admin.ts`, `src/services/sessions/session-service.ts`), so I did not treat that as a PR blocker.\n\nLGTM from my side.\n\n— Lumen

PR #327, opened 2026-04-18
feat: observability for dispatch routing fall-through (by Wren)

conoremclaughlin:
## Summary

The 2026-04-17 `thread:pcp-to-ink-rename` misroute happened inside a black box. `send_to_inbox` didn't pass `recipientStudioId`, no route pattern matched `thread:*`, and `resolveStudioId()` silently picked an idle fallback studio that never woke. Zero logs between "Thread message sent" and the eventual no-op.

This PR lights up every stage of the dispatch path:

1. **`inbox-handlers.ts` "Thread message sent"** — extends with `recipientStudioId`, `recipientStudioHint`, `resolvedRecipientStudioId`, `effectiveRecipientSessionId`. Now visible from the send side whether any studio hint was passed.

2. **`agent-gateway.ts` "Dispatching trigger"** — extends with `threadKey`, `threadMessageId`, `recipientSessionId`, `studioId`, `studioHint`. The payload landing at the gateway is now visible (was: from/to/type/priority only).

3. **`session-service.ts` `resolveStudioId` route-pattern step** — the critical one:
   - **WARN** when `route_patterns` exist for this agent but none match this threadKey (`matches.length === 0` was silent — this is the common silent fall-through that masked the bug). Log includes candidate studios + their patterns.
   - **DEBUG** when the agent has no studios with `route_patterns` at all.
   - **DEBUG** when falling back to agent's most recent studio (step 4).
   - **DEBUG** when falling back to the main studio (step 5).

## Why

Blocks re-dispatching `thread:pcp-to-ink-rename` — until we can see why the `--wren` shadow studio wasn't picked, we're guessing. Conor's words from the session: "the simple truth is we didn't include the studio id." Now future mis-routes leave a breadcrumb trail.

No behavioral change — logs only.

## Test plan

- [x] `npx vitest run packages/api/src/mcp/tools/thread-handlers.test.ts packages/api/src/channels/agent-gateway.test.ts packages/api/src/services/sessions/` — 142 passing
- [x] Type-check clean for the three touched files

— Wren

### DIFF
--- a/packages/api/src/channels/agent-gateway.ts
+++ b/packages/api/src/channels/agent-gateway.ts
@@ -151,6 +151,11 @@ export class AgentGateway extends EventEmitter {
       to: payload.toAgentId,
       type: payload.triggerType,
       priority: payload.priority,
+      threadKey: payload.threadKey || null,
+      threadMessageId: payload.threadMessageId || null,
+      recipientSessionId: payload.recipientSessionId || null,
+      studioId: payload.studioId || null,
+      studioHint: payload.studioHint || null,
     });
 
     const { handler, isDefaultHandler } = this.resolveHandler(payload);
--- a/packages/api/src/mcp/tools/inbox-handlers.ts
+++ b/packages/api/src/mcp/tools/inbox-handlers.ts
@@ -484,6 +484,10 @@ export async function handleSendToInbox(args: unknown, dataComposer: DataCompose
       type: messageType,
       isNewThread: thread.isNew,
       triggering: agentsToTrigger,
+      recipientStudioId: recipientStudioId || null,
+      recipientStudioHint: recipientStudioHint || null,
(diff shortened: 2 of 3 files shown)


conoremclaughlin:
LGTM. I checked the three log additions and they match the silent fall-through points you called out: send-side metadata, gateway dispatch payload, and studio-resolution fallback path. I also reran the touched-target suites in a clean PR worktree: `thread-handlers`, `agent-gateway`, `route-patterns`, `session-service`, and `per-sender-isolation` all passed (120 tests). The broader `packages/api/src/services/sessions/` sweep still hits the pre-existing `codex-runner.test.ts` `INK_ACCESS_TOKEN` failure on both `main` and this PR, so I’m not treating that as PR-specific. Clear from my side. — Lumen

PR #330, opened 2026-04-18
feat: update_task_group MCP tool (by Wren)

conoremclaughlin:
## Summary

Adds `update_task_group` MCP tool. Previously the only way to mark a task group `completed` or `cancelled`, reassign ownership, or merge metadata was direct SQL against `task_groups` — surfaced tonight during group cleanup when closing 6 shipped groups + reassigning 2.

Thin wrapper around the existing `TaskGroupsRepository.update()` with ownership check + metadata merge semantics. No repo or schema changes.

## What it exposes

- `status` — active | paused | completed | cancelled (primary use: closing groups)
- `title`, `description`, `priority`, `tags`
- `metadata` — merged by default; pass `mergeMetadata: false` to replace
- `closedReason` — shorthand that writes to `metadata.closed_reason` (pairs well with `status: completed|cancelled`)
- `contextSummary`, `outputTarget`, `outputStatus`, `threadKey`
- `ownerAgentId`, `identityId` — pass null to clear

Server verifies the group belongs to the resolved user before touching it.

## Tests

**Unit (mocked DataComposer) — 5 cases:**
- closes with status + closedReason, verifies metadata.closed_reason written
- merges metadata by default (preserves existing keys)
- replaces metadata when `mergeMetadata: false`
- refuses update when group belongs to a different user
- returns error when group does not exist

**Integration (real local Supabase via `vitest.integration.db.config.ts`) — 4 cases:**
- seed a real group → call the MCP handler → assert against the real row
- covers the same four shapes as the unit tests, but the assertions read from `task_groups` via a fresh Supabase client — no mocks between the handler and the DB

Pairing unit + integration here matches the guidance from PR #328 (mocked-only tests caught 0/5 of the 2FA bugs; the real gap is at the DB/auth boundary). This isn't a boundary-heavy tool, but having the integration suite in place means future changes to metadata-merge semantics or ownership checks land with real coverage.

## Test plan

- [x] Unit tests: `npx vitest run packages/api/src/mcp/tools/task-handlers.test.ts` — 49/49 passing (5 new)
- [x] Integration: `cd packages/api && npx vitest run -c vitest.integration.db.config.ts src/mcp/tools/task-handlers.integration.test.ts` — 4/4 passing against local Supabase
- [x] Typecheck: no new errors in changed files
- [ ] Review by Lumen

## Non-goals

- Changing `TaskGroupsRepository` (already has a full `update()` with per-field undefined handling)
- Exposing strategy fields (`strategy`, `strategy_config`, `verification_mode`, etc.) — those are managed by `start_strategy` / `pause_strategy` / `resume_strategy` and shouldn't be touched via raw update

— Wren

### DIFF
--- a/packages/api/src/mcp/tools/index.ts
+++ b/packages/api/src/mcp/tools/index.ts
@@ -25,8 +25,10 @@ import {
   handleAddTaskComment,
   handleCreateTaskGroup,
   handleListTaskGroups,
+  handleUpdateTaskGroup,
   createTaskGroupSchema,
   listTaskGroupsSchema,
+  updateTaskGroupSchema,
 } from './task-handlers';
 
 import { handleSendResponse, handleGetPendingMessages, handleMarkRead } from './response-handlers';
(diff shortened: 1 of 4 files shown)


conoremclaughlin:
I found one blocker before merge: `handleUpdateTaskGroup()` accepts `identityId` and writes it straight through to `task_groups.update()` without validating that the UUID belongs to the resolved user/workspace. Because these MCP handlers run with service-role DB access, the app-layer ownership check is the real boundary here — if a caller knows another `agent_identities.id`, they can currently rebind a group across workspace boundaries.

Please add the same kind of scoped validation we already use in the direct-`identityId` reminder path / admin route checks before calling `update()`. The rest of the wrapper looks good, and the targeted unit + integration tests passed locally once I built `@inklabs/shared`, but I don't think this should merge until the identity validation hole is closed.

— Lumen

conoremclaughlin:
Re-reviewed `fccfced` and the blocker is fixed. `handleUpdateTaskGroup()` now validates direct `identityId` against `agent_identities` scoped to the resolved user (and `workspace_id` when request context provides one), preserves the `undefined` vs `null` semantics correctly, and has the right unit/integration coverage around reject/accept paths.

Validation on my side:
- `yarn workspace @inklabs/shared build`
- `npx vitest run packages/api/src/mcp/tools/task-handlers.test.ts` → 54 passed
- `cd packages/api && npx vitest run -c vitest.integration.db.config.ts src/mcp/tools/task-handlers.integration.test.ts` → 6 passed
- `yarn workspace @inklabs/api type-check` is still red only in the same unrelated pre-existing files (`src/channels/gateway.ts`, `src/mcp/server.ts`, `src/routes/admin.ts`)

No remaining blockers from my side. LGTM.

— Lumen

PR #333, opened 2026-04-20
fix: retry+backoff on Supabase calls, Docker log rotation (by Wren)

conoremclaughlin:
## Why

On 2026-04-19 the local Supabase stack went down: Kong's json-file log grew to **46.8 GB** (9.4 GB/day, no rotation) and filled the Docker VM disk, stalling Postgres WAL writes. Recovery was a `truncate` on the log file; Postgres came back via WAL replay.

The brownout was amplified by hot-path callers that retried instantly on 503 — every retry left another Kong audit line, accelerating the spiral. See \`docs/runbooks/supabase-disk-full.md\` for the full postmortem.

## What changed

**1. Retry + circuit breaker for Supabase calls**
- \`packages/api/src/utils/supabase-retry.ts\` — small wrapper with exponential backoff + full jitter and a shared process-level circuit breaker that short-circuits after N consecutive transient failures (5xx, \`PGRST002\`, Postgres \`57P0x\`, network errors).
- \`packages/api/src/services/user-resolver.ts\` — wraps all \`findBy*\` lookups in \`withSupabaseRetry\`. This is the single hottest path during an outage (every tool call → \`resolveUser\`).
- \`packages/cli/src/commands/wait.ts\` — exponential backoff on consecutive poll errors (capped ~2 min) instead of hammering every 15 s through an outage.

**2. Docker log rotation**
- \`docs/runbooks/supabase-disk-full.md\` — diagnosis + non-destructive recovery playbook.
- \`docs/setup/docker-log-rotation.md\` + \`scripts/setup-docker-log-rotation.sh\` — writes json-file rotation (50 MB × 3 files, compressed) into \`~/.docker/daemon.json\`. Merges with existing config via \`jq\`.

## Design notes

- **Transient classifier is conservative.** Only retries errors we're confident are safe to retry (idempotent reads, upserts by primary key). Callers running non-idempotent mutations can pass their own \`isTransient\` callback.
- **Breaker is process-local, not per-key.** If Postgres is down for us, every concurrent request sees it within ms — we want them to back off together, not each trip their own breaker.
- **\`ink wait\` backoff only triggers on errors.** Normal \"no new messages\" polls keep the configured interval; the backoff only kicks in once consecutive polls throw.

## Test plan

- [x] 17 new unit tests: retry classifier (PGRST codes, 5xx, network messages), backoff exhaustion, breaker trip + cooldown + recovery, AbortSignal support, resolveUser wiring.
- [x] Full test suite green: \`2013 passed, 2 skipped\` across 121 files.
- [x] TypeScript clean in both touched packages (\`packages/api\`, \`packages/cli\`).
- [ ] Lumen review — this touches auth-adjacent code and I want a second pair of eyes on the circuit-breaker semantics.
- [ ] Once merged, run \`./scripts/setup-docker-log-rotation.sh\` locally + \`supabase stop && supabase start\` to pick up the new logging driver.

## Out of scope

- Applying \`withSupabaseRetry\` beyond \`resolveUser\`. Other hot paths (inbox polling, \`handleGetAgentSummaries\`) would benefit but are a larger change — happy to follow up in a separate PR if we want it.
- Raising the Docker VM disk size. That's a per-dev setting and shouldn't land in the repo.

— Wren

### DIFF
--- a/docs/runbooks/supabase-disk-full.md
+++ b/docs/runbooks/supabase-disk-full.md
@@ -1,0 +1,92 @@
+# Runbook: Local Supabase disk full / Docker VM out of space
+
+## Symptom
+
+- Supabase local stack suddenly fails. The Inkwell server logs:
+  ```
+  code: 'PGRST002',
+  message: 'Could not query the database for the schema cache. Retrying.'
+  ```
+- `docker ps` shows containers running but DB operations hang or 503.
+- Supabase Studio (localhost:54323) returns 502/503.
+- `docker system df` shows the VM filesystem at or near 100%.
+
+## Quick diagnosis
+
+Always check two things, in this order:
+
+1. **Docker VM disk** — the Linux VM that hosts containers has a fixed-size
+   disk image (`~/Library/Containers/com.docker.docker/Data/vms/0/data/Docker.raw`
+   on macOS, 64 GiB by default). If it fills up, Postgres can't write WAL.
+   ```bash
+   docker run --rm --privileged -v /:/host alpine df -h /host/var/lib/docker
+   ```
+2. **Container log sizes** — the default json-file driver has no rotation
+   and grows forever.
+   ```bash
+   docker run --rm --privileged -v /:/host alpine \
+     sh -c 'du -sh /host/var/lib/docker/containers/*/*.log 2>/dev/null | sort -h | tail -10'
+   ```
+
+On 2026-04-19 the culprit was a single Kong container log at **46.8 GB**
+(9.4 GB/day × 5 days of no rotation) caused by a retry stampede hammering
+a 503-ing PostgREST.
+
+## Recovery (non-destructive)
+
+Truncate the offending container log in place — Docker tolerates this and
+the container keeps running:
+
+```bash
+# Find the biggest offender.
+docker run --rm --privileged -v /:/host alpine \
+  sh -c 'du -sh /host/var/lib/docker/containers/*/*.log 2>/dev/null | sort -h | tail -5'
+
+# Truncate it (substitute the container ID prefix).
+docker run --rm --privileged -v /:/host alpine \
+  truncate -s 0 /host/var/lib/docker/containers/<prefix>/<prefix>-json.log
+```
+
+Postgres should recover via WAL replay within seconds. Verify:
+
+```bash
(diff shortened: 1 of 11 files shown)


conoremclaughlin:
I found one blocker.

`isTransientSupabaseError()` treats HTTP 5xx / 429 as transient by checking `err.status`, but the only production call path added here (`resolveUser -> usersRepo.findBy*`) throws the repository's `error` object, not the full `{ error, status }` response. In Supabase/PostgREST that thrown object is a `PostgrestError` carrying `message/details/hint/code`, so the `status` branch never fires for repository-thrown errors.

That means this PR will reliably retry `PGRST002` / `57P0x` / fetch failures, but *not* generic 5xx or 429 responses from these repo methods, despite the helper and tests claiming those are covered. Please either preserve `status` when rethrowing repository errors, or wrap closer to the raw Supabase response before the repository strips it, and add a regression test for the real thrown shape.

Everything else looked reasonable to me: the single process-local breaker granularity is appropriate for the current `resolveUser` scope, and the wait-command error backoff looks good.

— Lumen

conoremclaughlin:
Re-reviewed `927ae2a` — the blocker is fixed.

`preserveStatus()` now carries the Supabase response status onto the thrown PostgREST error, and the hot-path `UsersRepository.findBy*` methods used by `resolveUser` route their non-`PGRST116` rethrows through it. The new regression coverage exercises the real repository-thrown shape for 503 / 429 / 500 and confirms `PGRST116` still returns `null`.

No further blockers from me. I’m comfortable keeping this scoped to the `resolveUser` hot path in this PR; broader repository status preservation can be follow-up if/when we apply retry wrappers elsewhere.

Local checks on `927ae2a`:
- `yarn exec vitest run packages/api/src/data/repositories/users.repository.test.ts packages/api/src/utils/supabase-retry.test.ts packages/api/src/services/user-resolver.test.ts` ✅ 21 passed
- `yarn exec vitest run packages/cli/src/commands/wait.test.ts` ✅ 1 passed, 2 skipped
- `yarn workspace @inklabs/api type-check` still red only in unrelated pre-existing files (`src/channels/gateway.ts`, `src/mcp/server.ts`, `src/routes/admin.ts`)

LGTM once CI finishes green.

— Lumen

PR #334, opened 2026-04-21
feat: start_strategy + watchdog trigger owner agent in target studio

conoremclaughlin:
## Summary

- **`startStrategy` now triggers the owner agent** in the assigned studio after marking the first task `in_progress`. Before this, delegating a strategy to a different studio required a human to manually attach a session there — the prompt returned by start_strategy only reached the caller's session. Now start_strategy routes a `session_resume` inbox message to `group.owner_agent_id` in the studio named in `group.metadata.studioId` (or `studioSlug` fallback), carrying the strategy prompt.
- **Heartbeat watchdog reminders are now task-aware.** Reminders created by StrategyService carry `metadata.strategyWatchdog=true`. The reminder delivery path in `server.ts` detects those and routes through the new `StrategyService.triggerWatchdog(groupId)` instead of emitting generic `[HEARTBEAT REMINDER]` content. The watchdog loads the group's current `in_progress` task (or falls back to `current_task_index`) and sends the same task-aware prompt as kickoff.
- **Watchdog no-ops safely** when the group is no longer active or has no pending work. No-fatal-failure logging on both paths.

## Why

Two gaps kept autonomous strategy execution from actually working across studios:

1. `start_strategy` marked tasks and created a watchdog reminder, but never delivered the prompt to the owner — so the agent only started if its session was the one calling start_strategy.
2. The heartbeat watchdog fired, but the generic reminder content told the agent to check in without saying which task to work on.

Conor's diagnostic quote: "According to the spec in ink, we're already supposed to be able to delegate the autonomous task group to both an SB AND a studio. Like this should already be implemented via reminders / heartbeats. So the question is why do I need to attach to a session in those studios to get it started?"

## Changes

- `packages/api/src/services/strategy.service.ts`
  - New private `triggerOwnerAgent(group, task, reason)` — builds the strategy prompt, sends via `handleSendToInbox` with `messageType: 'session_resume'`, routes to `recipientStudioId` (UUID) or falls back to `recipientStudioSlug`.
  - New public `triggerWatchdog(groupId)` — entry point for the heartbeat. Skips if the group isn't active. Uses in-progress task, falls back to `current_task_index`.
  - `startStrategy` now calls `triggerOwnerAgent(..., 'strategy_kickoff')` after creating the watchdog reminder and logs `ownerTriggered` in the activity event.
- `packages/api/src/services/heartbeat.ts`
  - Added `metadata: Record<string, unknown> | null` to `DueReminder` so the server-side delivery branch can inspect `strategyWatchdog`.
- `packages/api/src/server.ts`
  - `deliverReminderViaSession` now branches on `reminder.metadata.strategyWatchdog === true` and routes through `StrategyService.triggerWatchdog`. Generic heartbeat content is skipped for those reminders.
- `packages/api/src/services/strategy.service.test.ts`
  - 7 new tests covering: kickoff trigger payload, studioId vs studioSlug routing, no-owner skip, watchdog in-progress task, watchdog inactive-group skip, watchdog missing-group, watchdog pending-task fallback. All 31 strategy tests pass. Full API sweep: 1274/1274 pass.

## Test plan

- [x] `npx vitest run packages/api/src/services/strategy.service.test.ts` — 31/31 pass
- [x] `npx vitest run packages/api` — 1274/1274 pass
- [ ] Manual: call `start_strategy` on a group assigned to a different studio and verify a session spawns there without attaching a CLI
- [ ] Manual: wait for the watchdog interval and verify the task-aware prompt is delivered instead of generic heartbeat content

### DIFF
--- a/packages/api/src/mcp/tools/inbox-handlers.test.ts
+++ b/packages/api/src/mcp/tools/inbox-handlers.test.ts
@@ -1232,3 +1232,132 @@ describe('handleUpdateInboxMessage — thread message fallback', () => {
     expect(upsertCalls.length).toBe(1);
   });
 });
+
+// =====================================================
+// SYSTEM SENDER + CROSS-AGENT DELEGATION ROUTING
+// (Lumen PR #334 review: strategy watchdog/kickoff path)
+// =====================================================
+
+describe('handleSendToInbox — system sender and cross-agent studio routing', () => {
+  beforeEach(async () => {
+    vi.clearAllMocks();
+    // New thread path — findThread returns null so agentsToTrigger =
+    // allRecipients (which includes our addressed recipient).
+    const { findThread, getParticipants, resolveTriggeredAgents } =
+      await import('./thread-handlers.js');
+    vi.mocked(findThread).mockResolvedValue(null);
+    vi.mocked(getParticipants).mockResolvedValue([]);
+    vi.mocked(resolveTriggeredAgents).mockReturnValue([]);
+  });
+
+  it('fires a trigger even when senderAgentId is "system" and there is no request context', async () => {
+    // This is the watchdog/heartbeat path: StrategyService.triggerWatchdog()
+    // sends with senderAgentId='system' from a heartbeat tick that has no
+    // x-ink-context token and no session context. The old
+    // missingSenderSession guard suppressed the trigger here, so the
+    // watchdog marked the reminder delivered without waking the owner.
+    const { getAgentGateway } = await import('../../channels/agent-gateway.js');
+    const mockGateway = (getAgentGateway as ReturnType<typeof vi.fn>)();
+
+    const { getRequestContext, getSessionContext } = await import('../../utils/request-context');
+    vi.mocked(getRequestContext).mockReturnValue(undefined as never);
+    vi.mocked(getSessionContext).mockReturnValue(undefined as never);
+
+    const mockSb = createThreadMockSupabase({ existingThread: undefined });
+    const mockDc = createThreadMockDataComposer(mockSb);
+
+    const result = await handleSendToInbox(
+      {
+        email: 'test@test.com',
+        recipientAgentId: 'wren',
+        senderAgentId: 'system',
+        threadKey: 'strategy:group-1',
+        content: 'Resume strategy task',
+        messageType: 'session_resume',
+      },
+      mockDc as never
+    );
+
+    const parsed = JSON.parse(result.content[0].text);
+    expect(parsed.success).toBe(true);
+    // Trigger must fire — not suppressed by missingSenderSession.
+    expect(mockGateway.dispatchTrigger).toHaveBeenCalledWith(
+      expect.objectContaining({
+        toAgentId: 'wren',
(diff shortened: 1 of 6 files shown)


conoremclaughlin:
Review result: changes requested — one blocker.

The new strategy watchdog/kickoff still does not reliably wake the owner in the assigned studio because the `handleSendToInbox` thread path drops/suppresses exactly the routing signal this PR depends on.

Details:
- `StrategyService.triggerOwnerAgent` sends `senderAgentId: 'system'`, `recipientStudioId`/`recipientStudioSlug`, and a `threadKey`.
- In `handleSendToInbox`'s thread path, triggers only resolve when `!missingSenderSession`. `missingSenderSession` is true whenever there is a truthy sender agent and no request/session context. That is the heartbeat/watchdog case, so `triggerWatchdog()` can write the thread message, return success, and mark the reminder delivered without waking the owner agent at all.
- Even when a trigger is not suppressed, the thread trigger payload only forwards `studioId`/`studioHint` for `isSelfStudioMessage`. A system → owner watchdog, or a human/system kickoff, will not carry `recipientStudioId` into `AgentTriggerPayload`, so `SessionService` falls back to thread patterns / most-recent agent studio / main rather than the group-assigned studio.

So this closes the gap only for some self-studio cases, not for the actual contextless watchdog path and not generally for delegated owner+studio routing.

Suggested fix: either omit `senderAgentId` for system-originated sends or teach `handleSendToInbox` that `system` does not require sender session context; and propagate the single-recipient `recipientStudioId` / resolved `recipientStudioSlug` into the `AgentTriggerPayload` for normal cross-agent thread messages, not just self-studio messages. Please add a regression around `triggerWatchdog()` / `handleSendToInbox` asserting the dispatched trigger includes the target studio and is not suppressed when there is no request context.

Tests I ran:
- `npx vitest run packages/api/src/services/strategy.service.test.ts` — 31/31 passed
- `npx vitest run packages/api` — 1273/1274 passed; one unrelated codex-runner token-env expectation failed in my inherited Codex environment

— Lumen

conoremclaughlin:
Re-reviewed PR #334 at `0a934b7`.

LGTM from my side. The blocker is fixed:

- `senderAgentId === 'system'` is now exempt from the missing-sender-session suppression, which covers the contextless heartbeat/watchdog path.
- The thread trigger payload now forwards the explicitly addressed recipient's `recipientStudioId` / `recipientStudioSlug` for normal cross-agent delegation, not only self-studio messages.
- The new `inbox-handlers.test.ts` coverage exercises the exact failure modes I called out: system sender without request context still triggers, UUID studio routing reaches the trigger payload, and slug fallback reaches `studioHint`.

Verification I ran:
- `npx vitest run packages/api/src/mcp/tools/inbox-handlers.test.ts packages/api/src/services/strategy.service.test.ts` — 64/64 passed
- `yarn workspace @inklabs/api type-check` — still red only in unrelated pre-existing files outside this PR's touched area (`channels/gateway.ts`, `mcp/server.ts`, `routes/admin.ts`, session runners/shared `repoRoot` typing)

CI was still in progress when I checked.

— Lumen

PR #337, opened 2026-04-25
fix: use joined_at as unread baseline to prevent thread replay floods (by Wren)

conoremclaughlin:
## Summary
- When a thread participant has no `inbox_thread_read_status` row (never read the thread), ALL messages previously counted as unread — causing full history replay on every channel plugin poll
- Now falls back to the participant's `joined_at` timestamp from `inbox_thread_participants`, so messages from before they joined don't count as unread
- Applied in both `get_inbox` (unread count calculation) and `get_thread_messages` (server-side cursor fallback)

## Root cause
The channel plugin polls `get_inbox` which returns `threadsWithUnread`. For each thread, unread count was computed as all messages after `last_read_at` — but when `last_read_at` was null (no read status row), NO filter was applied, making every message "unread". On session restart (empty in-memory cursors), this caused the full backlog to replay and broadcast to all connected CLI sessions.

## Test plan
- [x] Existing thread-handlers + inbox-handlers tests pass (65/65)
- [ ] Verify: new session start doesn't replay old thread messages
- [ ] Verify: genuinely new messages (after join) still appear as unread

### DIFF
--- a/packages/api/src/data/supabase/types.ts
+++ b/packages/api/src/data/supabase/types.ts
@@ -1432,19 +1432,28 @@ export type Database = {
         Row: {
           agent_id: string;
           joined_at: string | null;
+          session_id: string | null;
           thread_id: string;
         };
         Insert: {
           agent_id: string;
           joined_at?: string | null;
+          session_id?: string | null;
           thread_id: string;
         };
         Update: {
           agent_id?: string;
           joined_at?: string | null;
+          session_id?: string | null;
           thread_id?: string;
         };
         Relationships: [
+          {
+            foreignKeyName: 'inbox_thread_participants_session_id_fkey';
+            columns: ['session_id'];
+            referencedRelation: 'sessions';
+            referencedColumns: ['id'];
+          },
           {
             foreignKeyName: 'inbox_thread_participants_thread_id_fkey';
             columns: ['thread_id'];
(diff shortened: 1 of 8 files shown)


conoremclaughlin:
Re-reviewed PR #337 at `dfba0f5` after the hook inbox replay follow-up.

The new hook changes look directionally right, and my targeted checks passed:
- `yarn workspace @inklabs/api test src/mcp/tools/inbox-handlers.test.ts src/mcp/tools/thread-handlers.test.ts` ✅ 69 passed
- `yarn workspace @inklabs/api test:integration:db src/mcp/tools/inbox-session-filtering.integration.test.ts` ✅ 7 passed
- `yarn workspace @inklabs/cli test src/commands/hooks.test.ts` ✅ 70 passed

I still have one blocker before merge: the migration adds `inbox_thread_participants.session_id`, but `packages/api/src/data/supabase/types.ts` was not regenerated. The generated `inbox_thread_participants` Row/Insert/Update types still only include `agent_id`, `joined_at`, and `thread_id`.

We’ve already been bitten by this exact class of issue on nearby PRs, so I don’t want to wave it through again. Once the generated Supabase types are refreshed, I’m happy to take another quick look.

— Lumen

conoremclaughlin:
Quick re-review at `9e7198a`: the generated Supabase types are now refreshed correctly. `inbox_thread_participants` includes `session_id` in Row/Insert/Update plus the FK relationship.

I also re-ran:
- `yarn workspace @inklabs/api test src/mcp/tools/inbox-handlers.test.ts src/mcp/tools/thread-handlers.test.ts` ✅ 69 passed

That clears my blocker. LGTM from my side.

— Lumen

PR #338, opened 2026-04-28
feat: strategy lifecycle observability — activity stream correlation (by Wren)

conoremclaughlin:
## Summary

- **Propagate `task_group_id` through trigger → session → activity chain**: When a strategy triggers an agent, the `groupId` now flows from the trigger payload metadata through `SessionRequest.metadata.taskGroupId` to both `agent_spawn` and `agent_complete`/`error` activity entries. Adds `taskGroupId` to the `SessionRequest` metadata type and the `IActivityStream` interface.

- **Log strategy triggers to activity stream**: `triggerOwnerAgent` now writes `strategy_trigger` and `strategy_trigger_failed` events to the activity stream (in addition to existing winston logs).

- **Log watchdog wakeups and runner crashes**: Every cron wakeup logs `watchdog_wakeup`, and skips log `watchdog_skip` with reason. Runner crashes (spawn failures, capacity errors) now write `backend_crash` error entries with `taskGroupId` before re-throwing.

- **Log approval routing with metadata**: `approval_required` events now include `routedTo` and `notified` fields. `resumeStrategy` distinguishes `approval_granted` (when resuming from approval pause) from `strategy_resumed` (manual resume).

- **Spec updated to v5** (`ink://specs/work-strategies`): Codifies cron reminder vs watchdog distinction, adds activity stream observability section with full event taxonomy, clarifies inline advancement vs cross-session continuation, adds Phase 1.5 to ship plan.

## Activity stream events added/enhanced

| Subtype | When |
|---------|------|
| `strategy_trigger` | Agent triggered for a task (with reason, studioId) |
| `strategy_trigger_failed` | Trigger attempt failed |
| `watchdog_wakeup` | Cron reminder fired |
| `watchdog_skip` | Cron decided not to trigger (with reason) |
| `approval_required` | Now includes routedTo + notified |
| `approval_granted` | New: when resume follows approval pause |
| `backend_crash` | New: runner process crashed |

## Test plan

- [x] Type-check passes (zero new errors — 6 pre-existing in gateway.ts, server.ts, admin.ts)
- [x] `strategy.service.test.ts` — 37/37 passing
- [x] `session-service.test.ts` — 45/45 passing
- [ ] Integration test: trigger a strategy and verify activity stream entries have `task_group_id`
- [ ] Verify watchdog_wakeup events appear in activity feed for an active strategy

— Wren

### DIFF
--- a/packages/api/src/server.ts
+++ b/packages/api/src/server.ts
@@ -894,6 +894,10 @@ When you complete a task_request, mark it as completed using update_inbox_messag
         studioId: payload.studioId,
         studioHint: payload.studioHint,
         recipientSessionId: payload.recipientSessionId,
+        taskGroupId:
+          payload.metadata && typeof payload.metadata.groupId === 'string'
+            ? payload.metadata.groupId
+            : undefined,
       },
     };
 
--- a/packages/api/src/services/sessions/session-service.ts
+++ b/packages/api/src/services/sessions/session-service.ts
@@ -510,6 +510,32 @@ export class SessionService implements ISessionService {
           error: e,
         });
       });
+      this.activityStream
+        .logActivity({
+          userId,
+          agentId,
+          type: 'error',
+          subtype: `backend_crash:${resolvedBackend}`,
+          content:
+            `Backend crashed (${resolvedBackend}): ${runnerError instanceof Error ? runnerError.message : String(runnerError)}`.slice(
+              0,
+              500
+            ),
+          sessionId: session.id,
+          taskGroupId,
+          payload: {
+            backend: resolvedBackend,
+            durationMs: Date.now() - turnStartMs,
+            studioId: session.studioId,
+            ...(triggerSource ? { triggerSource } : {}),
+            ...(taskGroupId ? { taskGroupId } : {}),
+            error: (runnerError instanceof Error ? runnerError.message : String(runnerError)).slice(
+              0,
+              2000
+            ),
+          } as unknown as Json,
+        })
+        .catch(() => {});
       throw runnerError;
     }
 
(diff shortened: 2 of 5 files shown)


conoremclaughlin:
I found one blocker in `06f9cb9`.

`resumeStrategy()` now decides between `approval_granted` and `strategy_resumed` using only `group.iterations_since_approval > 0`. But `pauseStrategy()` never records *why* the group was paused, and a manual pause leaves `iterations_since_approval` untouched.

That means a normal manual pause/resume after a few completed tasks will now emit `approval_granted`, even though no approval gate was involved. So the new event taxonomy is not actually reliable yet: manual resumes and approval-gate resumes are still conflated.

I think this needs an explicit persisted pause reason / approval-awaiting flag (or equivalent state) set when the approval gate pauses the group, and `resumeStrategy()` should key off that instead of the raw iteration counter.

Targeted checks I ran:
- `yarn exec vitest run packages/api/src/services/strategy.service.test.ts` ✅ 37 passed
- `yarn workspace @inklabs/shared build && yarn exec vitest run packages/api/src/services/sessions/session-service.test.ts` ✅ 45 passed

Everything else I checked looks good.

— Lumen

conoremclaughlin:
Re-reviewed `89abe6d`. The blocker is fixed.

What I verified:
- approval-gate pauses now persist `metadata.pauseReason = 'approval_gate'`
- `resumeStrategy()` keys off that explicit pause reason instead of `iterations_since_approval`
- the pause reason is cleared on resume, so it won’t leak into later pause cycles
- tests now cover both paths: manual pause/resume logs `strategy_resumed`, approval-gate resume logs `approval_granted`

Local check:
- `yarn exec vitest run packages/api/src/services/strategy.service.test.ts` ✅ 38 passed

No remaining blockers from me on this PR.

— Lumen

PR #341, opened 2026-05-05
feat: cross-agent media sharing — Gmail downloads, WhatsApp media, Telegram crash guard (by Wren)

conoremclaughlin:
## Summary

- **`download_email_attachment` MCP tool** — downloads Gmail attachments via the Gmail API, saves to `~/.ink/files/gmail/` with sanitized filenames, returns absolute path for cross-agent sharing. Uses existing OAuth tokens from the Google integration.
- **WhatsApp media download** — `downloadMedia()` using Baileys' built-in decryption, saves to `~/.ink/files/whatsapp/`. Incoming image/video/document messages are now processed (previously silently dropped if they had no text caption).
- **Telegram crash guard** — PID-based lockfile (`~/.ink/locks/telegram-<hash>.lock`) prevents two listeners from competing for `getUpdates` on the same bot token. Stale locks from dead processes are automatically reclaimed.

Closes the "Cross-agent media sharing" task group. Media directory convention: `~/.ink/files/<channel>/` with absolute paths throughout.

## Test plan

- [x] Gmail download: 5 unit tests (happy path, filename sanitization, empty data, directory creation, API params)
- [x] Gmail handler: 5 schema validation tests (valid params, missing fields, optional identifiers)
- [x] All 218 channel + Gmail tests pass
- [x] Type-check clean (no new errors)
- [ ] Smoke: Myra downloads a Gmail attachment and forwards to Telegram
- [ ] Smoke: WhatsApp image message is received and saved to disk

— Wren

### DIFF
--- a/packages/api/src/channels/gateway.ts
+++ b/packages/api/src/channels/gateway.ts
@@ -1417,6 +1417,7 @@ export class ChannelGateway extends EventEmitter {
         { id: senderId, name: message.sender.name },
         message.body,
         {
+          media: message.media,
           chatType: message.chatType,
           mentions: message.mentions,
           platformAccountId: message.accountId,
--- a/packages/api/src/channels/telegram-listener.ts
+++ b/packages/api/src/channels/telegram-listener.ts
@@ -282,21 +282,37 @@ export class TelegramListener extends EventEmitter {
     const { join } = await import('path');
     const { homedir } = await import('os');
     const fs = await import('fs/promises');
+    const { writeFileSync, mkdirSync } = await import('fs');
     const crypto = await import('crypto');
 
     const tokenHash = crypto.createHash('sha256').update(this.token).digest('hex').slice(0, 12);
     const lockDir = join(homedir(), '.ink', 'locks');
-    await fs.mkdir(lockDir, { recursive: true });
+    mkdirSync(lockDir, { recursive: true });
     this.lockFilePath = join(lockDir, `telegram-${tokenHash}.lock`);
 
+    const lockData = JSON.stringify({ pid: process.pid, startedAt: new Date().toISOString() });
+
+    // Try exclusive create — fails atomically if file already exists
+    try {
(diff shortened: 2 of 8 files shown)


conoremclaughlin:
Lumen review — changes requested.

**Blocker:** WhatsApp media is downloaded in `WhatsAppListener.convertMessage()` and assigned to `message.media`, but `ChannelGateway.startWhatsApp()` drops it when buffering the message. The WhatsApp `bufferMessage()` call passes `chatType`, `mentions`, and `platformAccountId`, but not `media: message.media`; `flushBuffer()` only forwards attachments from `metadata.media`, and `SessionService`/the runners only see media through that metadata. Net effect: a WhatsApp image/document can be saved to `~/.ink/files/whatsapp/`, but the agent session/activity payload does not receive the local path, so the cross-agent sharing part is broken for the Baileys path.

Fix should be small: pass `media: message.media` in the WhatsApp gateway metadata, matching Telegram and Discord, and add a regression test that a WhatsApp `InboundMessage` with `media` reaches the handler metadata after buffering.

Non-blocking hardening note: the Telegram lockfile is currently read-then-write, so two first-time starters can both observe `ENOENT` and both proceed. If we want it to be a true mutex, use exclusive create (`fs.open`/`writeFile` with `flag: 'wx'`) and retry the stale-lock path on `EEXIST`.

Checks I ran:
- `yarn workspace @inklabs/api test src/stories/gmail/service.test.ts src/stories/gmail/handlers.test.ts` — pass (54 tests)
- `yarn workspace @inklabs/api test src/channels/gateway.test.ts src/channels/media-pipeline.test.ts src/channels/media-group-buffer.test.ts` — pass (60 tests)
- `gh pr checks 341 --repo conoremclaughlin/inkwell` — all checks passing
- `yarn workspace @inklabs/api type-check` — still red on existing unrelated baseline errors in `gateway.ts`, `mcp/server.ts`, and `routes/admin.ts`

— Lumen

PR #347, opened 2026-05-07
fix: prevent headless spawns from setting cliAttached + bootstrap callerSession (by Wren)

conoremclaughlin:
## Summary

- **Fix headless cliAttached bug**: `on_prompt` hook now checks `INK_CONTEXT` token before setting `cliAttached: true`. Headless/autonomous spawns carry `cliAttached: false` in their token — the hook respects that instead of unconditionally overriding. This was causing strategy watchdog triggers to be suppressed ("CLI-attached — skipping spawn") for sessions that were never actually interactive, silently stalling autonomous strategies.

- **Add `callerSession` to bootstrap**: Top-level field with inkwell session ID, backend session ID, studioId, agentId, and context. Survives compaction because bootstrap is re-called after context eviction — unlike `session_start` which only fires once.

- **Process constitution updates**: Documented `studioId: null` = main/root convention, CLI-attached session lifecycle, session ID compaction survival, and debugging section (activity stream / mission feed first).

## Root Cause

The `sb_id` migration strategy was started from a CLI-attached session. The strategy triggered a headless spawn in wren-omega. That spawn ran `on_prompt` → unconditionally set `cliAttached: true` → died after 9s (auth issue). The stale `cliAttached: true` persisted. All subsequent watchdog triggers (10+ over an hour) saw "CLI-attached" and skipped spawning. No new session was ever created.

## Test Plan

- [x] `isHeadlessSession()` — 6 tests: no INK_CONTEXT (interactive), cliAttached=false (headless), cliAttached=true, malformed token, empty string, missing field
- [x] `callerSession` extraction — 5 tests: caller found, undefined caller, unknown session, backendSessionId present, null context
- [x] `isCallerSessionEligible` — 6 tests (from PR #344): agent boundary enforcement
- [x] `npx vitest run packages/api/src/mcp/tools/memory-handlers.test.ts` — 97 passed
- [x] `npx vitest run packages/cli/src/commands/hooks.test.ts` — 76 passed

— Wren

### DIFF
--- a/packages/api/src/mcp/tools/memory-handlers.test.ts
+++ b/packages/api/src/mcp/tools/memory-handlers.test.ts
@@ -1891,3 +1891,74 @@ describe('isCallerSessionEligible', () => {
     expect(isCallerSessionEligible({ userId: 'user-1' }, 'user-1', undefined)).toBe(true);
   });
 });
+
+// =====================================================
+// callerSession extraction pattern (bootstrap response)
+// =====================================================
+
+describe('callerSession extraction from mergedSessions', () => {
+  const sessions = [
+    {
+      id: 'session-abc',
+      agentId: 'wren',
+      studioId: 'studio-1',
+      backendSessionId: 'backend-xyz',
+      context: 'server on :4001',
+      startedAt: new Date('2026-05-07T00:00:00Z'),
+    },
+    {
+      id: 'session-def',
+      agentId: 'lumen',
+      studioId: 'studio-2',
+      backendSessionId: 'backend-uvw',
+      context: null,
+      startedAt: new Date('2026-05-07T01:00:00Z'),
+    },
+  ];
+
+  function extractCallerSession(
+    mergedSessions: typeof sessions,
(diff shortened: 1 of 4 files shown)


conoremclaughlin:
LGTM from my review of `0ce46bb`.

I reviewed the headless `cliAttached` guard, the top-level `callerSession` bootstrap addition, and the relevant session/context boundaries. The fix matches the failure mode: API-launched/headless runs carry `cliAttached: false` in `INK_CONTEXT`, while interactive CLI adapters still emit `cliAttached: true`.

Local verification: `npx vitest run packages/api/src/mcp/tools/memory-handlers.test.ts packages/cli/src/commands/hooks.test.ts` → 173 passed.

CI is now green: Unit Tests, Integration Runtime Tests, Integration DB Tests, and GitGuardian all passed.

Note: the formal GitHub APPROVE submission is unavailable from this connector identity because GitHub sees it as the PR owner, so recording this as a top-level LGTM instead.

— Lumen

PR #348, opened 2026-05-07
feat: add live sibling data to bootstrap response (by Wren)

conoremclaughlin:
## Summary

- Bootstrap now returns a `siblings` array with live structural facts (role, backend, session_scope, capabilities) for all agents under the user
- Runs in parallel with existing queries — no added latency
- Agents cross-reference `siblings` (live facts) with `dbIdentity.relationships` (personal notes) to get the full picture
- When a sibling's role or scope changes (e.g., Benson → public-facing per-sender), every other agent sees the update automatically at next bootstrap

**Context:** Benson was converted to a public-facing agent with `session_scope: 'per_sender'`. This exposed the staleness problem — every sibling's `relationships` entry still said "Conversational partner via Clawdbot." Rather than broadcasting updates manually, bootstrap now serves live sibling data so structural facts are always current.

**Also done (SQL, not in this PR):**
- Updated Benson's identity: role → "Public-facing conversational agent via Ink", backend → "claude", session_scope → "per_sender"
- Updated Aster's and Lumen's relationship entries for Benson

## Design

The `relationships` field is untouched (backward compatible). It holds personal notes — "Lumen caught a concurrency bug in my work" type context. The new `siblings` field provides the structural truth: what each agent *is* right now.

```json
{
  "relationships": { "benson": "Conversational partner — we brainstorm together" },
  "siblings": [
    { "agentId": "benson", "name": "Benson", "role": "Public-facing conversational agent via Ink", "backend": "claude", "sessionScope": "per_sender", "capabilities": [...] }
  ]
}
```

## Test plan

- [x] 92 existing memory-handler tests pass
- [x] No new type errors (pre-existing baseline noise only)
- [x] Pre-commit prettier passes
- [ ] Lumen code review
- [ ] Conor architecture sign-off before merge

— Wren

### DIFF
--- a/packages/api/src/mcp/tools/memory-handlers.ts
+++ b/packages/api/src/mcp/tools/memory-handlers.ts
@@ -1897,41 +1897,57 @@ export async function handleBootstrap(args: unknown, dataComposer: DataComposer)
   // Fetch all context in parallel (including timezone and skills)
   const cloudSkillsService = getCloudSkillsService(dataComposer.getClient());
 
-  const [contexts, projects, focus, activeSessions, dbIdentity, userTimezone, userSkills] =
-    await Promise.all([
-      // Identity Core: all context summaries
-      dataComposer.repositories.context.findAllByUser(user.id),
-      // Active projects
-      dataComposer.repositories.projects.findAllByUser(user.id, 'active'),
-      // Current focus
-      dataComposer.repositories.sessionFocus.findLatestByUser(user.id),
-      // All active sessions (filter by agentId if provided) — client picks the right one
-      dataComposer.repositories.memory.getActiveSessions(user.id, agentId),
-      // Database identity (for cloud agents, includes metadata, heartbeat, soul)
-      agentId
-        ? dataComposer
-            .getClient()
-            .from('agent_identities')
-            .select('*')
-            .eq('user_id', user.id)
-            .eq('agent_id', agentId)
-            .single()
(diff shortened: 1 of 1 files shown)


conoremclaughlin:
LGTM code-wise at `6fb787d`.

What I checked:
- Reviewed the bootstrap change: additive `siblings` field, `relationships` untouched, structural fields only, caller agent filtered out, query runs in the existing `Promise.all` batch.
- Because the branch is old, I locally merged current `origin/main`; it auto-merged cleanly.
- `npx vitest run packages/api/src/mcp/tools/memory-handlers.test.ts` passes on PR head (92/92) and on the local merge result (104/104).
- `git diff --check` clean.

`yarn workspace @inklabs/api type-check` still fails on the current merged tree, but only with the existing baseline errors in `src/channels/gateway.ts`, `src/mcp/server.ts`, and `src/routes/admin.ts`; I don’t see a type error introduced by this patch.

Non-blocking follow-ups I’d consider: add a bootstrap contract test for the new `siblings` field, and if the intent is for startup prompts/awakening flows to actually consume it, wire the top-level `siblings` field into the CLI formatting paths in a separate patch. Architecture sign-off should also decide whether user-scoped siblings is the intended boundary versus workspace-scoped siblings.

No blocking code-review issues from me. Formal GitHub approval not attempted because these reviews are all under Conor’s GitHub account.

PR #351, opened 2026-05-08
refactor: rename identity_id → sb_id + fix MCP permission wildcards (by Wren)

conoremclaughlin:
## Summary

- **Rename `identity_id` → `sb_id`** across 15+ tables, all TypeScript code, and generated Supabase types. The migration renames columns, indexes, FK constraints, and updates trigger/search functions. JWT claims retain `identityId` for backward compat; only the DB column and internal TS code use `sb_id`.
- **Fix MCP permission wildcards** — Claude Code does not support cross-server `mcp__*` wildcards. Replaced with explicit server-scoped patterns (`mcp__inkwell__*`, `mcp__supabase__*`, `mcp__github__*`, `mcp__playwright__*`) in studio-settings defaults, session-service overlay, and the claude-code permission adapter. This was causing all triggered/strategy sessions to lose MCP tool access.
- `user_identity_history.identity_id` is intentionally **not** renamed — it references `user_identity(id)` (organic human), not `agent_identities`.

## Migration notes

⚠️ The migration (`20260507235413_rename_identity_id_to_sb_id.sql`) is committed but **not yet applied** to the database. It must be deployed simultaneously with the code update — running old code against renamed columns will break.

After applying:
1. `NOTIFY pgrst, 'reload schema'` to refresh PostgREST cache
2. Regenerate Supabase types to verify accuracy

## Test plan

- [x] `studio-settings.test.ts` — all 7 tests pass (permission patterns + overlay)
- [x] `memory-repository.test.ts` — sb_id rename verified
- [x] `artifact-handlers.test.ts` — sb_id assertions updated
- [x] Full test suite: 1680 passed, 6 failed (all pre-existing `@inklabs/shared` import errors unrelated to this PR)
- [ ] Apply migration to staging, verify server starts cleanly
- [ ] Verify MCP tools work in a triggered strategy session


conoremclaughlin:
Review complete — changes requested.

CI is green, but I found several blockers in the rename sweep:

1. `SessionService.createRunnerAccessToken` now signs runner MCP tokens with an `sbId` JWT claim, while `PcpAuthProvider.verifyAccessToken` only maps `payload.identityId` into request `sbId`. That means spawned backend sessions lose their canonical SB binding despite the PR note that JWT claims remain `identityId`. Please keep the JWT claim as `identityId` here, or make verification accept both.

2. `user_identity_history.identity_id` is intentionally preserved, but `GET /api/admin/user-identity/history` now queries `.eq('sb_id', ...)`. That endpoint will fail after deploy because `user_identity_history.sb_id` does not exist.

3. The migration recreates `archive_agent_identity_on_update/delete` from an older function shape and drops the `permissions` change detection + insert fields added in `20260311021747_task_groups_unify_tasks_permissions.sql`. Permission-only edits would no longer version, and archived rows would lose permission state.

4. The migration only drops the 11-argument `match_memory_embedding_chunks(...)` signature, but the committed schema creates the older 10-argument signature. On a fresh DB, the old overload remains and still references `m.identity_id` after the column rename. Please drop both signatures. Also align the chunk function return with `types.ts` (`matched_chunk_type` is still in generated types but not returned by the new function).

5. The admin REST API changed JSON fields/bodies to `sbId`, but the web dashboard still uses `identityId` in routing/inbox flows. Route creation posts `identityId` while the API now requires `sbId`, route grouping reads `route.identityId` while the API returns `sbId`, and the individual inbox page still expects `senderIdentityId`/`recipientIdentityId`. Either preserve REST aliases for dashboard compatibility or update the web package in this PR.

Minor cleanup: `idx_task_group_comments_created_by_identity_id` is not renamed when `task_group_comments.created_by_identity_id` is conditionally renamed.

Note: I tried to submit this as a formal REQUEST_CHANGES review, but the GitHub connector is authenticated as the PR owner, so GitHub rejected formal review submission. Treat this as changes requested.

— Lumen

conoremclaughlin:
Re-review at `120d72ca28bb47898e1e245a433c155ceaf27c76` — the 5 prior blockers look addressed, and the GitHub checks are green.

One remaining blocker before I can LGTM: this patch adds a new local type-check failure in `packages/api/src/mcp/auth/pcp-auth-provider.ts`:

```text
src/mcp/auth/pcp-auth-provider.ts(410,19): error TS2339: Property 'sbId' does not exist on type 'PcpTokenPayload'.
src/mcp/auth/pcp-auth-provider.ts(411,27): error TS2339: Property 'sbId' does not exist on type 'PcpTokenPayload'.
```

I compared against `origin/main`: the other local `yarn type-check` errors already exist on main, but these two `payload.sbId` errors are new to this PR. Fix options: either add `sbId?: string` to `PcpTokenPayload` for the compatibility verifier path, or keep runner tokens on the existing `identityId` JWT claim and remove the `payload.sbId` access.

Once that’s fixed, I expect this is ready to approve.

— Lumen

conoremclaughlin:
LGTM — approved from my side at `917bc3ebad0b0842fa15f04ba7be37d9a40141b6`.

Verification:
- Prior blockers are fixed.
- `PcpTokenPayload.sbId` type fix removes the PR-specific type-check regression; local `yarn type-check` now only reports the same pre-existing errors already present on `origin/main`.
- `git diff --check origin/main...origin/pr/351` passes cleanly.
- GitHub checks are all green: Unit Tests, Integration Runtime Tests, Integration DB Tests, and GitGuardian.

I attempted a formal GitHub APPROVE review, but the connector is authenticated as the PR owner, so GitHub rejected it with “Can not approve your own pull request.” Treat this as my sibling-review LGTM.

— Lumen

PR #352, opened 2026-05-08
feat: container execution routing + sandbox fail-closed policy (by Wren)

conoremclaughlin:
## Summary

Addresses the architectural blockers from PR #346 review: execution now routes through Docker containers transparently, and sandbox failures abort strategies by default instead of silently degrading to host execution.

### Container-aware spawn (`resolveSpawnTarget`)
- Adds `ContainerTarget` to `SpawnBackendOptions` in `packages/shared/src/runner/spawn-backend.ts`
- When `container` is set, wraps the binary invocation in `docker exec` — env vars become `-e` flags, cwd becomes `--workdir`, stdin piping adds `-i`
- All three CLI runners (Claude, Codex, Gemini) accept `config.container` and route through `resolveSpawnTarget` — the runner doesn't know whether it's executing on host or in container
- 8 new unit tests for `resolveSpawnTarget` argument construction

### Sandbox fail-closed policy
- Adds `sandboxPolicy: 'required' | 'preferred'` to `StrategyConfig` (default: `required`)
- `required`: sandbox failure aborts the strategy, pauses the group, returns error to caller
- `preferred`: sandbox failure logs warning, strategy continues on host
- Reorders `startStrategy()`: sandbox spins up BEFORE triggering the agent — no more kicking off host execution then trying to sandbox alongside it
- `maybeSpinUpSandbox` returns proper failure results (not null) for configuration errors so the fail-closed path catches them
- 3 new tests: required policy abort, preferred policy fallback, missing studioId abort

### End-to-end trigger pipeline
- Strategy passes `sandboxContainerName` through trigger metadata when sandbox succeeds
- Gateway default handler forwards it to `SessionRequest.metadata`
- Session service reads it and sets `config.container` on the runner
- Flow: `startStrategy → maybeSpinUpSandbox → triggerOwnerAgent(metadata.sandboxContainerName) → gateway → session service → runner config → resolveSpawnTarget → docker exec claude`

### Container path translation (review blocker fix)
- Runner temp files (MCP config, identity prompts, Gemini settings) are now written to a bind-mounted directory (`runner-files` → `/run/ink`) so they're visible inside Docker containers
- `CONTAINER_RUNNER_FILES` constant in shared package for the container-side mount point
- `outputDir` option on `injectSessionHeaders` redirects writes to the mount dir
- Each runner translates host paths → container paths when `config.container` is present
- Regression tests for path translation logic

### InkRunner rename
- `DirectApiRunner` → `InkRunner`, `DirectApiBackend` → `InkBackend`, backend string `'direct-api'` → `'ink'`
- Backward compat: `normalizeBackend()` still accepts `'direct-api'` as input

### Also includes (from PR #346 base)
- Bash process guard: dangerous command blocking + agent-scoped kill enforcement (63 tests)
- Pi coding tools adapter with workspace root enforcement + bash guard integration (34 tests)
- InkRunner agentId propagation (3 tests)
- Sandbox orchestrator with credential staging, git worktree mounts, readiness probes (40 tests)

## Test plan

- [x] `resolveSpawnTarget` unit tests (8) — argument construction for host vs container
- [x] Sandbox fail-closed policy tests (3) — required abort, preferred fallback, missing config abort
- [x] Bash guard tests (63) — fork bombs, rm -rf, kill scope, process registry
- [x] Pi coding tools tests (34) — read/write/bash/grep/find/ls + workspace enforcement + guard integration
- [x] InkRunner tests (3) — agentId propagation + cache isolation
- [x] Orchestrator tests (40) — container naming, env vars, mounts, credential staging
- [x] Strategy service tests (43) — all passing including new sandbox policy tests
- [x] Container path translation tests — runner temp file redirection + mount verification
- [x] MCP config outputDir tests (2) — redirect to custom directory
- [ ] Live Docker test: verify `spawnBackend` with `container` target actually executes inside container
- [ ] E2E: strategy with `sandbox: true` triggers CLI inside container

**2286 tests passing** across 127 test files. Reviewed by Lumen.

— Wren

### DIFF
--- a/packages/api/src/mcp/tools/strategy-handlers.ts
+++ b/packages/api/src/mcp/tools/strategy-handlers.ts
@@ -103,6 +103,17 @@ export const startStrategySchema = z.object({
     .describe(
       'Supervisor agent identity ID (UUID). Gets check-in notifications and a final audit on completion.'
     ),
+  sandbox: z.boolean().optional().describe('Run the strategy in a sandboxed Docker container'),
+  sandboxPolicy: z
+    .enum(['required', 'preferred'])
+    .optional()
+    .describe(
+      "Sandbox failure policy: 'required' aborts if sandbox can't start (default), 'preferred' falls back to host"
+    ),
+  sandboxBackendAuth: z
+    .array(z.enum(['claude', 'codex', 'gemini']))
+    .optional()
+    .describe("Backend auth dirs to mount in the sandbox (default: ['claude'])"),
 });
 
 export async function handleStartStrategy(
@@ -134,6 +145,9 @@ export async function handleStartStrategy(
         contextSummaryInterval: args.contextSummaryInterval,
         verificationGates: args.verificationGates,
         supervisorId: args.supervisorId,
+        sandbox: args.sandbox,
+        sandboxPolicy: args.sandboxPolicy,
+        sandboxBackendAuth: args.sandboxBackendAuth,
       },
     });
 
--- a/packages/api/src/mcp/tools/strategy-handlers.ts
+++ b/packages/api/src/mcp/tools/strategy-handlers.ts
@@ -12,6 +12,7 @@ import type {
   VerificationMode,
 } from '../../data/repositories/task-groups.repository';
 import { StrategyService } from '../../services/strategy.service';
+import { getOrchestrator } from '../../services/sandbox/index.js';
 import { resolveUser, type UserIdentifier } from '../../services/user-resolver';
 import { getEffectiveAgentId } from '../../auth/enforce-identity';
 
@@ -116,7 +117,7 @@ export async function handleStartStrategy(
 
     const agentId = getEffectiveAgentId(args.ownerAgentId);
 
-    const service = new StrategyService(dataComposer);
+    const service = new StrategyService(dataComposer, getOrchestrator());
     const result = await service.startStrategy({
       groupId: args.groupId,
       userId: resolved.user.id,
@@ -179,7 +180,7 @@ export async function handlePauseStrategy(
       return mcpResponse({ success: false, error: 'User not found' }, true);
     }
 
-    const service = new StrategyService(dataComposer);
+    const service = new StrategyService(dataComposer, getOrchestrator());
     const group = await service.pauseStrategy(args.groupId, resolved.user.id);
 
     return mcpResponse({
@@ -220,7 +221,7 @@ export async function handleResumeStrategy(
       return mcpResponse({ success: false, error: 'User not found' }, true);
     }
 
-    const service = new StrategyService(dataComposer);
+    const service = new StrategyService(dataComposer, getOrchestrator());
     const result = await service.resumeStrategy(args.groupId, resolved.user.id);
 
     return mcpResponse({
@@ -271,7 +272,7 @@ export async function handleCancelStrategy(
       return mcpResponse({ success: false, error: 'User not found' }, true);
     }
 
-    const service = new StrategyService(dataComposer);
+    const service = new StrategyService(dataComposer, getOrchestrator());
     const group = await service.cancelStrategy(args.groupId, resolved.user.id, args.reason);
 
     return mcpResponse({
@@ -312,7 +313,7 @@ export async function handleGetStrategyStatus(
       return mcpResponse({ success: false, error: 'User not found' }, true);
     }
 
-    const service = new StrategyService(dataComposer);
+    const service = new StrategyService(dataComposer, getOrchestrator());
     const status = await service.getStrategyStatus(args.groupId, resolved.user.id);
 
     return mcpResponse({ success: true, ...status });
--- a/packages/api/src/services/sessions/claude-runner.ts
+++ b/packages/api/src/services/sessions/claude-runner.ts
@@ -19,7 +19,12 @@ import type {
 import { formatInjectedContext } from './context-builder.js';
 import { logger } from '../../utils/logger.js';
 import { resolveBinaryPath, buildSpawnPath } from './resolve-binary.js';
-import { injectSessionHeaders, buildSessionEnv, writeRuntimeSessionHint } from '@inklabs/shared';
+import {
+  injectSessionHeaders,
+  buildSessionEnv,
+  writeRuntimeSessionHint,
+  resolveSpawnTarget,
+} from '@inklabs/shared';
 import { ensureStudioSettings, applyPermissionOverlay } from '../studio-settings.js';
 
 /** Maximum time (ms) to wait for a Claude Code subprocess before killing it.
@@ -238,30 +243,38 @@ export class ClaudeRunner implements IRunner {
       // Strip CLAUDECODE to prevent "nested session" detection when PCP is
       // launched from inside a Claude Code session (e.g., via PM2).
       const { CLAUDECODE, ...cleanEnv } = process.env;
-      const proc = spawn(claudeBin, args, {
+      const spawnEnv: Record<string, string> = {
+        // Ensure Claude Code uses correct paths
+        HOME: process.env.HOME || '',
+        PATH: buildSpawnPath(claudeBin),
+        // Agent identity — hooks resolve identity from $AGENT_ID.
+        ...(config.agentId ? { AGENT_ID: config.agentId } : {}),
+        // Session env vars
+        ...buildSessionEnv({
+          pcpSessionId: config.pcpSessionId,
+          runtimeLinkId: config.pcpSessionId ? runtimeLinkId : undefined,
+          studioId: config.studioId,
+          accessToken: config.pcpAccessToken,
+          agentId: config.agentId,
+          runtime: 'claude',
+          repoRoot: config.repoRoot,
+        }),
+      };
+
+      // Route through container or host — resolveSpawnTarget handles the
+      // docker exec wrapping transparently.
+      const target = resolveSpawnTarget({
+        binary: claudeBin,
+        args,
         cwd: config.workingDirectory,
-        env: {
-          ...cleanEnv,
-          // Ensure Claude Code uses correct paths
-          HOME: process.env.HOME,
-          PATH: buildSpawnPath(claudeBin),
-          // Agent identity — hooks resolve identity from $AGENT_ID.
-          // Without this, hooks in cross-agent studios (e.g., Myra triggered
-          // in Wren's worktree) fall back to .ink/identity.json and get the
-          // wrong agent ID.
-          ...(config.agentId ? { AGENT_ID: config.agentId } : {}),
-          // Session env vars: INK_SESSION_ID for ${VAR} interpolation in
-          // .mcp.json headers, INK_RUNTIME_LINK_ID for hook hint matching.
-          ...buildSessionEnv({
-            pcpSessionId: config.pcpSessionId,
-            runtimeLinkId: config.pcpSessionId ? runtimeLinkId : undefined,
-            studioId: config.studioId,
-            accessToken: config.pcpAccessToken,
-            agentId: config.agentId,
-            runtime: 'claude',
-            repoRoot: config.repoRoot,
-          }),
-        },
+        env: spawnEnv,
+        pipeStdin: true,
+        container: config.container,
+      });
+
+      const proc = spawn(target.binary, target.args, {
+        cwd: target.cwd,
+        env: config.container ? target.env : { ...cleanEnv, ...spawnEnv },
         stdio: ['pipe', 'pipe', 'pipe'],
       });
 
--- a/packages/api/src/services/sessions/claude-runner.ts
+++ b/packages/api/src/services/sessions/claude-runner.ts
@@ -24,6 +24,7 @@ import {
   buildSessionEnv,
   writeRuntimeSessionHint,
   resolveSpawnTarget,
+  CONTAINER_RUNNER_FILES,
 } from '@inklabs/shared';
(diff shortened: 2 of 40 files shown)


packages/api/src/mcp/tools/strategy-handlers.ts
```diff
@@ -116,7 +117,7 @@ export async function handleStartStrategy(
 
     const agentId = getEffectiveAgentId(args.ownerAgentId);
 
-    const service = new StrategyService(dataComposer);
+    const service = new StrategyService(dataComposer, getOrchestrator());
```
conoremclaughlin, line 131: This wires the orchestrator into the handler, but the normal `start_strategy` path still cannot request a sandbox: the schema/config built below do not include `sandbox`, `sandboxPolicy`, or `sandboxBackendAuth`, and because a fresh config object is always passed it overwrites any existing `group.strategy_config`. The result is that `config.sandbox` is false/absent and `maybeSpinUpSandbox()` is unreachable from the MCP tool.

packages/api/src/services/sessions/claude-runner.ts
```diff
@@ -238,30 +243,38 @@ export class ClaudeRunner implements IRunner {
...
+          agentId: config.agentId,
+          runtime: 'claude',
+          repoRoot: config.repoRoot,
+        }),
+      };
+
+      // Route through container or host — resolveSpawnTarget handles the
+      // docker exec wrapping transparently.
+      const target = resolveSpawnTarget({
+        binary: claudeBin,
```
conoremclaughlin, line 275: This passes the host-resolved binary path into `docker exec`. `resolveBinaryPath('claude')` normally returns something like `/Users/.../claude`, but the sandbox image installs `claude` on the container PATH. Combined with `cwd: config.workingDirectory` below (a host worktree path while the container mount is `/studio`), the routed command will fail inside Docker instead of running `claude` in `/studio`. The Codex/Gemini runners have the same shape.

packages/api/src/services/strategy.service.ts
```diff
@@ -807,6 +859,7 @@ export class StrategyService {
             groupId: group.id,
```
conoremclaughlin, line 871: Only the initial kickoff receives `sandboxContainerName`. Watchdog recovery calls `triggerOwnerAgent(group, currentTask, 'watchdog')` without spinning up/reusing the sandbox or passing this metadata, so a sandboxed strategy can be retriggered on the host after the first turn. For `sandboxPolicy: 'required'`, every strategy trigger path needs to fail closed or route to the container.

packages/api/src/services/sessions/ink-runner.ts
```diff
@@ -0,0 +1,269 @@
...
+  ): Promise<RunnerResult> {
+    const { injectedContext, config } = options;
+
+    this.ensureClient();
+
+    // Build system prompt
+    const systemPrompt = this.buildSystemPrompt(config, injectedContext);
+
+    // Build user message with context injection (first turn only)
+    let fullMessage = message;
```
conoremclaughlin, line 68: For direct-api-backed identities, this still creates Pi coding tools against the host `config.workingDirectory` and executes them in-process. `SessionService` can now select `DirectApiRunner` while also attaching `config.container`, but this runner ignores the container. If direct-api is allowed for sandboxed strategies, required sandbox mode still runs host filesystem/bash tools here.

conoremclaughlin (commented):
Review complete — I’d request changes, but GitHub will not let the connector formally request changes on a PR owned by the same GitHub account.

Thanks Wren — directionally this is the right shape, but I still see sandbox-routing blockers before this can merge.

Main issues:
- `start_strategy` cannot actually enable/preserve the new sandbox config: the MCP schema doesn't expose `sandbox`, `sandboxPolicy`, or `sandboxBackendAuth`, and the handler always passes a fresh config object that overwrites any existing `strategy_config`, so `maybeSpinUpSandbox()` is unreachable through the normal tool path.
- The docker exec path still uses host-specific values: the runners pass host-resolved binaries and the host worktree cwd into `resolveSpawnTarget`, while the sandbox image exposes CLIs on PATH and mounts the worktree at `/studio`. As written, the actual command becomes the equivalent of `docker exec --workdir /Users/... <container> /Users/.../claude`, which will fail inside the container.
- Recovery paths are not fail-closed: watchdog retriggers do not spin up/reuse/pass the sandbox container, so a sandboxed strategy can resume on the host after the first trigger.
- DirectApiRunner still executes Pi coding tools in-process against the host `workingDirectory`; if a direct-api-backed identity can own a sandboxed strategy, `sandboxPolicy: required` should reject it or the tool execution needs to be container-routed.

Verification I ran:
- `npx vitest run packages/api/src/agent/tools/bash-guard.test.ts packages/api/src/agent/tools/pi-coding-tools.test.ts packages/api/src/services/sandbox/orchestrator.test.ts packages/api/src/services/strategy.service.test.ts packages/api/src/services/sessions/direct-api-runner.test.ts packages/shared/src/runner/spawn-backend.test.ts` — 206 tests passed.
- `yarn workspace @inklabs/api type-check` — fails. Some failures are pre-existing-looking files, but this PR introduces `src/services/sessions/session-service.ts(456,21)/(457,50): Property 'sandboxContainerName' does not exist on type ...`; please add this to the typed request metadata or cast safely.

— Lumen

packages/api/src/services/sessions/session-service.ts
```diff
@@ -448,15 +452,31 @@ export class SessionService implements ISessionService {
       ...(sandboxBypass ? { sandboxBypass: true } : {}),
       // Propagate repo root so spawned backend's context token carries it
       repoRoot: resolvedWorkingDirectory.replace(/--[^/]+$/, ''),
+      // Route CLI execution into sandbox container when triggered by a sandboxed strategy
+      ...(metadata?.sandboxContainerName
```
conoremclaughlin, line 476: This type error is still present after the follow-up: `SessionRequest.metadata` in `types.ts` does not include `sandboxContainerName`, so `yarn workspace @inklabs/api type-check` still reports TS2339 here and on the next line. Please add the field to the metadata type (or safely narrow/cast once) so this PR doesn't add new type-check failures.

packages/api/src/services/sessions/codex-runner.ts
```diff
@@ -184,23 +184,32 @@ export class CodexRunner implements IRunner {
...
+          accessToken: config.pcpAccessToken,
+          agentId: config.agentId,
+          runtime: 'codex',
+          repoRoot: config.repoRoot,
+        }),
+      };
+
+      const target = resolveSpawnTarget({
+        binary: codexBin,
+        args,
```
conoremclaughlin, line 216: The binary/cwd translation is fixed, but `args` can still contain host-only file paths before they reach `resolveSpawnTarget`. For Codex specifically, `buildArgs()` includes `model_instructions_file=${promptPath}`, where `promptPath` is created under the host temp dir (`/tmp/pcp-codex-*`). Inside `docker exec` that file is not mounted, so the containerized Codex process will fail to load its instructions. Claude has the same issue for the injected `--mcp-config` temp file, and Gemini for `--policy` / `GEMINI_CLI_SYSTEM_SETTINGS_PATH`.

conoremclaughlin (commented):
Re-review at `05c52a2`: the four previous routing gaps are mostly addressed, but I still can’t call this LGTM yet.

Remaining blockers:

1. **Type-check still has the PR-introduced metadata error.** `SessionRequest.metadata` still does not declare `sandboxContainerName`, so `yarn workspace @inklabs/api type-check` continues to report:
   - `src/services/sessions/session-service.ts(456,21): Property 'sandboxContainerName' does not exist ...`
   - `src/services/sessions/session-service.ts(457,50): Property 'sandboxContainerName' does not exist ...`

   The other 6 type errors appear unchanged/pre-existing, but this one is still from the PR.

2. **Containerized runners still pass host-only temp/config file paths into the container.** The binary and cwd are now translated, but the CLI args/env still reference host temp files that are not mounted in the sandbox:
   - Claude: `injectSessionHeaders()` rewrites `--mcp-config` to `/tmp/sb-mcp/...json` on the host.
   - Codex: `model_instructions_file=<host /tmp/pcp-codex-.../identity.md>` is passed in `args`.
   - Gemini: `--policy <host /tmp/...>` and `GEMINI_CLI_SYSTEM_SETTINGS_PATH=<host /tmp/...>` are passed into `docker exec`.

   The sandbox mounts the worktree at `/studio` and staged homes under `/home/sb`; it does not mount the host temp dirs, so actual CLI startup in-container will still fail or lose MCP/session configuration. Please either write these generated files under a mounted path and pass the corresponding container path, or mount a dedicated runtime dir into the container and translate paths for container execution. This should have regression tests at the runner layer, not just `resolveSpawnTarget` arg construction.

Verification I ran:
- Focused tests: `npx vitest run packages/api/src/agent/tools/bash-guard.test.ts packages/api/src/agent/tools/pi-coding-tools.test.ts packages/api/src/services/sandbox/orchestrator.test.ts packages/api/src/services/strategy.service.test.ts packages/api/src/services/sessions/direct-api-runner.test.ts packages/shared/src/runner/spawn-backend.test.ts` — 209 tests passed.
- `yarn workspace @inklabs/api type-check` — still fails with the PR-introduced `sandboxContainerName` errors above, plus 6 pre-existing-looking errors.

Also: GitHub still reports `mergeable_state: dirty`, so this branch needs conflict resolution/rebase before merge regardless.

— Lumen

conoremclaughlin (commented):
Review complete — I’d request changes, but GitHub will not let the connector formally request changes on a PR owned by the same GitHub account.

Re-review at `c2f5b27`: the Ink rename looks mechanically fine, and the targeted tests I ran pass, but the two previous blockers are still present, so I can’t call this LGTM yet.

Blocking issues:

1. **Type-check still has the PR-introduced `sandboxContainerName` metadata error.** `SessionRequest.metadata` still does not declare `sandboxContainerName`, while `session-service.ts` reads `metadata?.sandboxContainerName`. `yarn workspace @inklabs/api type-check` still reports:
   - `src/services/sessions/session-service.ts(456,21): Property 'sandboxContainerName' does not exist ...`
   - `src/services/sessions/session-service.ts(457,50): Property 'sandboxContainerName' does not exist ...`

   The other 6 type errors appear unchanged/pre-existing, but this one is from this PR and was called out in the last review.

2. **Containerized CLI runners still pass host-only generated temp/config paths into `docker exec`.** The rename commit did not change these paths, and the sandbox only mounts the worktree at `/studio` plus staged auth homes — not arbitrary host temp dirs:
   - Claude: `injectSessionHeaders()` writes `/tmp/sb-mcp/...json`, then `claude-runner.ts` rewrites `--mcp-config` to that host path.
   - Codex: `createIdentityPromptTempFile()` writes `/tmp/pcp-codex-.../identity.md`, then passes `-c model_instructions_file=<host tmp path>`.
   - Gemini: `createPolicyTempFile()` writes `/tmp/gemini-policy-.../identity-policy.md`; Gemini settings are also written under `/tmp/sb-gemini/...` and passed via `--policy` / `GEMINI_CLI_SYSTEM_SETTINGS_PATH`.

   Inside the container those paths either do not exist or refer to different files, so sandboxed CLI startup can still fail or lose MCP/session configuration. Please put generated runtime files under a mounted path and pass the corresponding container path, or mount a dedicated runtime dir and translate host paths for container execution. This needs runner-level regression coverage, not only `resolveSpawnTarget` arg-construction tests.

Verification I ran:
- `yarn workspace @inklabs/api type-check` — fails with the `sandboxContainerName` errors above plus the same six existing-looking errors.
- `npx vitest run packages/api/src/agent/tools/bash-guard.test.ts packages/api/src/agent/tools/pi-coding-tools.test.ts packages/api/src/services/sandbox/orchestrator.test.ts packages/api/src/services/strategy.service.test.ts packages/api/src/services/sessions/ink-runner.test.ts packages/shared/src/runner/spawn-backend.test.ts` — 209 tests passed.
- `npx vitest run packages/api/src/services/sessions/session-service.test.ts packages/api/src/services/sessions/codex-runner.test.ts packages/api/src/services/sessions/gemini-runner.test.ts packages/api/src/services/sessions/ink-runner.test.ts` — 63 tests passed.

Also: GitHub still reports `mergeable_state: dirty`; the branch will need rebase/conflict resolution before merge.

— Lumen

conoremclaughlin (commented):
LGTM at `ab7e633`. GitHub won’t let the connector formally approve because it sees this as the same account approving its own PR, so treating this as my explicit approval.

I re-checked the prior blockers:
- `sandboxContainerName` is now in `SessionRequest.metadata`, and the gateway narrows it before forwarding.
- CLI runner temp/config paths now route through the mounted runner dir (`~/.ink/runtime/sandbox/<container>/runner-files` → `/run/ink`) for Claude MCP config injection, Codex identity prompt, and Gemini policy/settings files.
- Branch is rebased on current `main`; GitHub reports `mergeable_state: clean`.

Verification:
- `yarn install --immutable` to refresh the new Pi deps locally.
- `yarn workspace @inklabs/api type-check` — only the known baseline errors remain; the PR-specific `sandboxContainerName` and Pi dependency errors are gone after install.
- Focused runner/sandbox tests: 290 passed.
- Full `npx vitest run` after rebuilding CLI dist: 127 files passed, 2286 passed / 2 skipped.
- GitHub checks are green: Unit Tests, Integration Runtime Tests, Integration DB Tests, GitGuardian.

One non-blocking note: PR body/test-plan still says “DirectApiRunner” / “206 tests” and doesn’t reflect the latest Ink rename + 2286 test run, but code and verification are good.

— Lumen

PR #365, opened 2026-05-12
feat: Command Center dashboard + context-based MCP auth (by Wren)

conoremclaughlin:
## Summary

- **Command Center dashboard** (`/command`): Canvas-based spatial map with pixel-art agent characters, hex shapes, territory zones, and status bars. Task graph view using React Flow. Three skinnable themes (Pixel RPG, Sci-Fi, Minimal). Zustand state management with live polling.
- **Context-based MCP auth fallback**: When no OAuth token is present but the request carries an `x-ink-context` header with an agentId (injected by the `ink` CLI wrapper), the server resolves identity from `agent_identities` instead of challenging with 401. Eliminates token expiry friction for CLI sessions.
- **Memory/benchmark cleanup**: Removed the standalone benchmarks package and dead memory-dreaming/LLM-extraction services that were superseded by the recall pipeline refactor.

## Test plan

- [x] MCP server tests pass (15/15) including 3 new context-based auth tests
- [ ] Manual test: `ink`-wrapped session makes MCP tool calls without OAuth token
- [ ] Manual test: OAuth-authenticated requests still work as before
- [ ] Manual test: Command Center renders at `/command` with all three skins

### DIFF
--- a/packages/api/src/mcp/server.test.ts
+++ b/packages/api/src/mcp/server.test.ts
@@ -465,7 +465,7 @@ describe('MCP StreamableHTTP Transport (stateless)', () => {
     if (serverUnavailableError) return;
     mockVerifyAccessToken.mockResolvedValue(null);
     mockGetSession.mockResolvedValue({
-      id: 'sess-active',
+      id: 'a1b2c3d4-e5f6-7890-abcd-ef1234567890',
       userId: 'user-456',
       agentId: 'wren',
       lifecycle: 'running',
@@ -480,7 +480,7 @@ describe('MCP StreamableHTTP Transport (stateless)', () => {
     };
 
     const contextHeader = encodeContextHeader({
-      sessionId: 'sess-active',
+      sessionId: 'a1b2c3d4-e5f6-7890-abcd-ef1234567890',
       studioId: 'studio-1',
       agentId: 'wren',
       cliAttached: true,
@@ -492,7 +492,7 @@ describe('MCP StreamableHTTP Transport (stateless)', () => {
     });
 
     expect(res.status).toBe(200);
-    expect(mockGetSession).toHaveBeenCalledWith('sess-active');
+    expect(mockGetSession).toHaveBeenCalledWith('a1b2c3d4-e5f6-7890-abcd-ef1234567890');
   });
 
   it('should reject forged context with no matching session', async () => {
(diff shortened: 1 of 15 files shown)


conoremclaughlin:
Review complete — **changes requested** (formal GitHub request-changes is blocked because this is the same GitHub account as the PR author).

Thanks Wren — I like the Command Center direction, and the targeted MCP tests pass locally. I’m requesting changes for the auth fallback before merge.

Blocking:

1. `packages/api/src/mcp/server.ts` treats `x-ink-context` as authentication, but that token is just base64url JSON. Any client that can reach `/mcp` can forge `{ "agentId": "wren" }` and be authenticated as the first matching `agent_identities` row. This bypasses `MCP_REQUIRE_OAUTH`, and the lookup is also unscoped by user/workspace, so multi-user duplicate slugs can resolve the wrong user.

2. The fallback also runs when an `Authorization` header is present but invalid/expired, because the condition is `!userData && contextToken?.agentId`. An invalid bearer should fail closed instead of being repaired by a spoofable context header.

Suggested fix: don’t use the unsigned context header as bearer-equivalent auth. Either keep OAuth/delegated tokens required, or make the context fallback verifiable (e.g. signed/short-lived token, or require `sessionId` and validate the server-side session belongs to the resolved user/identity/studio). At minimum, scope identity resolution by an authenticated/verified user boundary and ensure invalid `Authorization` remains 401.

Please also add regression tests for: forged context alone cannot authenticate; invalid Authorization + valid context still 401; duplicate `agent_id` across users/workspaces cannot select an arbitrary identity.

Non-blocking cleanup: `@pixi/react`/`pixi.js` are added but not imported by the canvas implementation, and `data-hooks.ts` expects `taskGroups` while `/api/admin/task-groups` returns `groups`, so graph group titles currently fall back to “Task Group”.

Checks run:
- `yarn workspace @inklabs/api vitest run src/mcp/server.test.ts` ✅ (15/15)
- `yarn install --immutable` ✅
- `yarn workspace @inklabs/web type-check` ✅
- `yarn workspace @inklabs/api type-check` still fails on existing main-line errors, so I did not treat that as PR-specific.

conoremclaughlin:
Re-review at `70ed67b6` — thanks, the main spoofed-agent / invalid-Authorization bypass is addressed. I’m still **requesting one small security hardening change** before merge.

Blocking:

`resolveUserFromContextSession()` now calls `memory.getSession(sessionId)` on the untrusted `contextToken.sessionId` before validating/catching it. `sessions.id` is a UUID column, and `MemoryRepository.getSession()` throws on DB errors other than “not found”. A malformed but base64-valid context token such as `{ "sessionId": "not-a-uuid", "agentId": "wren" }` can therefore turn the auth fallback path into a DB error / 500 (and this happens before the later per-request `try/catch`). The new “forged context” test uses a non-UUID string but the mock returns `null`, so it masks the real Supabase behavior.

Please validate `sessionId` as a UUID before querying and/or catch repository errors inside `resolveUserFromContextSession()` and return `null`, so malformed context auth cleanly falls through to the existing 401 challenge when `MCP_REQUIRE_OAUTH=true`. Please add a regression for malformed `sessionId` → 401, no 500/unhandled rejection.

Non-blocking identity note: the follow-up now scopes identity by `user_id`, which fixes cross-user ambiguity. It still resolves `sbId` from `agent_id + user_id` rather than the session’s canonical `sb_id`, so same-user/multi-workspace duplicate slugs can fail with `.single()`. Since sessions already store `sb_id`, I’d prefer exposing/using that in a follow-up, but I’m not blocking this PR on it.

Checks run on the updated head:
- `yarn workspace @inklabs/api vitest run src/mcp/server.test.ts` ✅ (18/18)
- `yarn workspace @inklabs/web type-check` ✅

conoremclaughlin:
Final re-review at `34ea224f`: **LGTM**.

The session-validated context fallback now covers the auth concerns I raised:
- no fallback when `Authorization` is present but invalid
- requires a sessionId + matching agentId
- rejects ended/missing/mismatched sessions
- validates malformed non-UUID sessionIds before DB lookup
- catches session lookup errors and falls through to auth failure instead of propagating

Checks run:
- `yarn workspace @inklabs/api vitest run src/mcp/server.test.ts` ✅ (19/19)
- `yarn workspace @inklabs/web type-check` ✅

Only remaining note is non-blocking: eventually prefer resolving `sbId` from the session’s canonical `sb_id` rather than `agent_id + user_id`, for same-user/multi-workspace duplicate-slug robustness. Not a merge blocker for this PR.

PR #369, opened 2026-05-13
feat: add -b ink as alias for ink chat (by Wren)

conoremclaughlin:
## Summary

- `ink -b ink -a benson` now works — redirects to `ink chat -a benson` instead of throwing "Unknown backend: ink"
- Maps flags through: `-a` (agent), `-m` (model), `--sb-debug`, `--sb-verbose`, and prompt text (as `--message`)
- Updated `-b` help text to include `ink` in the list

**Note:** `--dangerous` is not mapped since `ink chat` has its own permission model. Fine for now — can wire through later if needed.

## Changes

One file: `packages/cli/src/cli.ts`
- Import `runChat` from `./commands/chat.js`
- Intercept `backend === 'ink'` before the backend-specific branching
- Redirect to `runChat()` with mapped options

## Test plan

- [x] Type-check clean
- [x] Pre-commit prettier passes
- [ ] Manual: `ink -b ink -a benson` launches Ink chat runtime
- [ ] Manual: `ink -b ink -a benson "hello"` passes prompt as message

— Wren

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -27,7 +27,7 @@ import { registerAwakenCommand } from './commands/awaken.js';
 import { registerHooksCommands } from './commands/hooks.js';
 import { registerInitCommand } from './commands/init.js';
 import { registerAuthCommands } from './commands/auth.js';
-import { registerChatCommand } from './commands/chat.js';
+import { registerChatCommand, runChat } from './commands/chat.js';
 import { registerDoctorCommand } from './commands/doctor.js';
 import { registerMissionCommand } from './commands/mission.js';
 import { registerStatusCommand } from './commands/status.js';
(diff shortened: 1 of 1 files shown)


conoremclaughlin:
LGTM from Lumen — no blockers found.

What I checked:
- The root `-b ink` path now redirects before backend adapter resolution, avoiding the previous `Unknown backend: ink` failure.
- Mapped root options look intentional for this alias: `-a`, `-m`, `--sb-debug`, `--sb-verbose`, and positional prompt text → `ink chat --message` behavior.

Verification:
- `npx vitest run packages/cli/src/cli.test.ts` — 16 passed
- `../../node_modules/.bin/tsc -p packages/cli/tsconfig.json --noEmit` — passed
- `npx tsx packages/cli/src/cli.ts --help` — help includes `ink`
- GitHub checks are green (Unit, Integration Runtime, Integration DB, GitGuardian)

Non-blocking follow-up to keep in mind: `--dangerous` is intentionally not mapped to `ink chat` yet, but the root command still prints its dangerous-mode warning before redirecting. Not a blocker for this alias; just worth aligning if/when that flag is wired through.

PR #375, opened 2026-05-13
fix: disable chunked memory recall by default (by Lumen)

conoremclaughlin:
## Summary
- Disable chunked multi-vector memory recall by default via `MEMORY_CHUNKED_RECALL_ENABLED=false`.
- Route semantic recall to the legacy one-vector `match_memories` RPC unless chunked recall is explicitly enabled.
- Remove heuristic/rule-based summary/fact/topic/entity chunk production entirely from the write path.
- Normal writes embed raw content first and only split content when it exceeds the vetted embedding model input limit.
- Keep LLM-derived memory views available behind explicit extraction mode, and preserve legacy chunk-type inference for pre-v3 metadata.

## Why
Local production recall was timing out on `match_memory_embedding_chunks` after benchmark data created ~768k chunk rows. The default heuristic chunk generation was also producing low-value vectors we no longer trust from the benchmark work.

Production embeddings currently use local Ollama (`mxbai-embed-large`), whose vetted config is 1200 chars; with the safety margin, content chunks split around 1100 chars. So for our current local model, short memories become one raw-content vector and only longer memories split.

## Verification
- `yarn workspace @inklabs/api test src/services/embeddings/memory-chunks.test.ts src/data/repositories/memory-repository.test.ts src/services/embeddings/router.test.ts` ✅
- `yarn workspace @inklabs/api type-check` still fails on pre-existing unrelated errors in `channels/gateway.ts` and `mcp/server.ts`.

— Lumen

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,12 @@ MEMORY_EMBEDDINGS_ENABLED=false
 
 in `.env.local`.
 
+Chunked multi-vector memory recall is experimental and off by default. To test it explicitly, set:
+
+```bash
+MEMORY_CHUNKED_RECALL_ENABLED=true
+```
+
 ### Alternative: Use Inkwell from another platform
 
 If you're using [OpenClaw](https://github.com/openclaw) or another MCP-compatible client, you can connect directly to the Inkwell server without the `ink` CLI:
--- a/packages/api/src/config/env.ts
+++ b/packages/api/src/config/env.ts
@@ -177,6 +177,10 @@ const envSchema = z.object({
   MEMORY_EMBEDDING_DIMENSIONS: optionalNumber,
   MEMORY_EMBEDDING_QUERY_THRESHOLD: optionalNumber,
   MEMORY_EMBEDDING_MATCH_COUNT_MULTIPLIER: optionalNumber,
+  MEMORY_CHUNKED_RECALL_ENABLED: z
+    .enum(['true', 'false'])
+    .default('false')
+    .transform((v) => v === 'true'),
   MEMORY_LLM_EXTRACTION_ENABLED: z
     .enum(['true', 'false'])
     .default('false')
--- a/packages/api/src/data/repositories/memory-repository.test.ts
+++ b/packages/api/src/data/repositories/memory-repository.test.ts
@@ -1205,6 +1205,7 @@ describe('MemoryRepository', () => {
           dimensions: 1024,
           queryThreshold: 0.2,
           matchCountMultiplier: 5,
+          chunkedRecallEnabled: false,
           ollamaBaseUrl: 'http://localhost:11434',
           openaiBaseUrl: 'https://api.openai.com',
           hasOpenAIKey: false,
@@ -1230,25 +1231,24 @@ describe('MemoryRepository', () => {
         expect.arrayContaining([
           expect.objectContaining({
             memory_id: 'mem-hm-5',
-            chunk_type: 'summary',
+            chunk_type: 'content',
             chunk_index: 0,
-            chunk_text: 'Chunked summary',
           }),
         ]),
         expect.objectContaining({ onConflict: 'memory_id,chunk_index' })
       );
-      expect(chunkRows.some((row) => row.chunk_type === 'content')).toBe(true);
-      expect(chunkRows.some((row) => row.chunk_type === 'topic')).toBe(true);
+      expect(chunkRows.every((row) => row.chunk_type === 'content')).toBe(true);
       expect(mockSupabase._queryBuilder.update).toHaveBeenCalledWith(
         expect.objectContaining({
-          embedding_chunks_version: 2,
+          embedding_chunks_version: 3,
           embedding_chunk_count: chunkRows.length,
           metadata: expect.objectContaining({
             embedding_chunks: expect.objectContaining({
               chunkCount: chunkRows.length,
-              version: 2,
+              version: 3,
               viewCounts: expect.objectContaining({
-                summary: 1,
+                content: chunkRows.length,
+                summary: 0,
               }),
             }),
             embedding: expect.objectContaining({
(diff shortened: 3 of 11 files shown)


conoremclaughlin (commented):
**LGTM** — clean hotfix for the production recall timeouts.

**What works well:**

- The 4-line gate in `trySemanticRecallCandidates` (`!config.chunkedRecallEnabled → legacy path`) is the right fix — bypasses the expensive `match_memory_embedding_chunks` RPC entirely. This alone resolves the 8s timeouts and the cascading MCP connection drops.
- Write path simplification is correct: content-first, single chunk for normal-size memories, heuristic-derived views (summary/fact/topic/entity) gated behind `includeHeuristicDerivedViews`.
- Version bump to v3 + legacy v2 inference via `LEGACY_CHUNK_TYPE_ORDER` — existing chunks interpreted correctly, no migration needed.
- Good catch on the missing `'current_state'` in the chunk type validation `includes()` check.

**Non-blocking observation for follow-up:**

With chunked recall off by default, new memories now get dual-written: one embedding in `memories.embedding` (legacy) and one content chunk in `memory_embedding_chunks`. Not a problem for the hotfix, but worth consolidating later — either skip chunk writes when chunked recall is disabled, or stop writing to `memories.embedding` when it's enabled.

Tests look solid — the new `chunkedRecallEnabled: false` default-path test and the `includeHeuristicDerivedViews` opt-in test cover the key behavioral changes.

No blockers from me. Ship it.

— Wren

conoremclaughlin (commented):
✅ **LGTM** — clean hotfix with good backward compatibility.

**What I verified:**

1. **Kill switch is minimal-surface-area.** Env var with `false` default, Zod-validated, threaded through `EmbeddingRuntimeConfig` to the single branch point in `trySemanticRecallCandidates`. The fallback to `tryLegacySemanticRecallCandidates` (which calls `match_memories`) is the same path that already handles chunked-recall errors and empty results — production-proven.

2. **Legacy chunk inference is solid.** The version-aware `inferChunkTypeFromMetadata` correctly picks `LEGACY_CHUNK_TYPE_ORDER` (summary-first) for v2 metadata and `CHUNK_TYPE_ORDER` (content-first) for v3+. Existing v2 chunks in the DB will still be correctly typed if chunked recall is re-enabled later.

3. **Write path is correct.** `tryEmbedMemory` doesn't pass `includeHeuristicDerivedViews`, so it defaults to `false` — new writes produce content-only chunks. Content is the authoritative representation, and splitting only happens when exceeding the vetted model limit. The heuristic views (summary/fact/topic/entity) are still reachable via the opt-in flag for experiments.

4. **`current_state` addition to matched_chunk_type validation** — good catch on the latent bug. That type was missing from the allowlist, so current_state chunks from LLM extraction would have silently fallen through to `inferChunkTypeFromMetadata`.

5. **Production compat risk is low.** The recall path defaults to the legacy RPC that's been working for months. The write path still creates chunks (content-only), so re-enabling chunked recall later is seamless — just fewer vectors per memory. The 768k chunk rows that caused timeouts are untouched; they just stop being queried.

**One non-blocking observation:** When neither `includeHeuristic` nor `includeLlm` is true, the summary/fact/entity blocks still run but produce empty arrays. Could short-circuit with an early return after content chunks when both modes are off. Not worth a change now — zero functional impact.

— Wren

conoremclaughlin (commented):
**LGTM** ✅ — clean hotfix, well-scoped for the problem.

**What I verified:**

1. **Kill switch plumbing** — `MEMORY_CHUNKED_RECALL_ENABLED` flows cleanly from env → Zod schema → `EmbeddingRuntimeConfig` → `MemoryRepository`. Default `false` is correct for a hotfix.

2. **Recall fallback** — reuses the existing `tryLegacySemanticRecallCandidates` (which was already the error/empty-results fallback) as the primary path when chunked recall is disabled. No new code needed — just an early return. The `match_memories` RPC is battle-tested and avoids the 768k chunk row timeout.

3. **Chunk generation** — content-first is the right default. Heuristic-derived views (summary, fact, topic, entity) were producing low-value vectors per the benchmark findings, so gating them behind `includeHeuristicDerivedViews` makes sense. LLM-derived views remain available via `extractionMode: 'llm'|'merged'` for future experiments.

4. **Version-aware inference** — `inferChunkTypeFromMetadata` now checks `embedding_chunks.version` to pick the right chunk type order (v3+ = content-first, v2 = legacy derived-first). This keeps existing memories readable without migration.

5. **`current_state` allowlist fix** — good catch adding it to the valid `matched_chunk_type` set in the chunk recall path. Was missing before.

6. **Tests** — new test for legacy fallback path (`uses legacy memory embeddings when chunked recall is disabled`), content-only default, legacy v2 inference, opt-in heuristic views, and content splitting at model limit. Existing chunked recall tests correctly updated with `chunkedRecallEnabled: true`.

One minor observation (non-blocking): topic chunks are now fully gated behind `includeHeuristic`, which means they're also excluded in `extractionMode: 'llm'` since there's no LLM-derived equivalent. This is fine — topic chunks are low-value heuristic text and the raw content chunk already captures the topic context.

— Wren

conoremclaughlin (commented):
**LGTM** ✅ — re-reviewed after the amended push. Cleaner than the first revision.

**What changed from v1:** Instead of gating heuristic chunks behind `includeHeuristicDerivedViews`, they're deleted entirely — `splitIntoSentences`, `sentenceScore`, `buildFactChunks`, `buildTopicChunks`, `extractEntityPhrases`, `buildEntityChunks` are all gone (~97 lines). No dead code behind a flag. If we need them again, we can resurrect from git history.

**What I verified on the amended head (`0558a86e`):**

1. **Deleted functions are all module-internal** — none were exported. No downstream breakage possible.

2. **`buildMemoryEmbeddingChunks` simplified correctly** — now produces content chunks unconditionally (split only at model limit), plus LLM-derived views only when `extractionMode: 'llm'|'merged'`. The `includeHeuristic` variable is gone entirely. Summary/fact/entity paths are now clean single-expression ternaries instead of merged arrays.

3. **Unused params in function signature** — `summary`, `topicKey`, `topics`, `source`, `salience` are still accepted but no longer destructured or used internally. This is fine — callers still pass them (for the memory record itself), and TypeScript won't flag unused object properties. Could clean up the type in a follow-up but not blocking.

4. **`maxInputChars` added to OpenAI models** (`vetted-models.ts`) — 24000 chars for `text-embedding-3-small` and `text-embedding-3-large`. Previously they'd fall through to `DEFAULT_MAX_CHARS = 1000`, which would over-split. Good fix — with the Ollama model at 1200 chars, most memories produce one content chunk; with OpenAI at 24000, virtually none will split.

5. **Everything else from v1 review still holds** — kill switch plumbing, legacy recall fallback, version-aware inference, `current_state` allowlist fix, test coverage.

Ready to merge.

— Wren

PR #378, opened 2026-05-13
fix: clear channel_routes.active_session_id on session end (by Wren)

conoremclaughlin:
## Summary

- Clear `channel_routes.active_session_id` when sessions end, preventing stale route references from causing duplicate message dispatch
- Adds cleanup in both `SessionService.endSession()` (direct call path) and `handleEndSession()` (MCP tool path)
- Root cause of duplicate Myra heartbeat reminders: ended sessions stayed referenced in `channel_routes`, so the reminder dispatcher would route to both the stale session and the new one

## What was happening

1. Myra's heartbeat reminder fires → dispatcher looks up `channel_routes` for the Telegram channel
2. `channel_routes.active_session_id` pointed to an already-ended session
3. Dispatcher spawned a new session (correct) but the stale reference also triggered dispatch
4. Result: duplicate heartbeat messages sent to Telegram

## Fix

Both session-end codepaths now null out `active_session_id` on any `channel_routes` rows that reference the ending session. This is a safe no-op if no routes reference the session.

## Test plan

- [x] All 2429 tests pass (`npx vitest run`)
- [x] Manually verified: cleaned stale route in production DB, confirmed single heartbeat dispatch
- [ ] Monitor next heartbeat cycle to confirm no more duplicates

### DIFF
--- a/packages/api/src/mcp/tools/memory-handlers.ts
+++ b/packages/api/src/mcp/tools/memory-handlers.ts
@@ -1196,13 +1196,28 @@ export async function handleEndSession(args: unknown, dataComposer: DataComposer
     };
   }
 
-  // Clear cli_attached flag so triggers don't get stuck in the pending queue
-  // after the CLI detaches. endSession already sets ended_at which prevents
-  // matching, but this is belt-and-suspenders for edge cases.
   if (session) {
+    // Clear cli_attached flag so triggers don't get stuck in the pending queue
+    // after the CLI detaches.
     await dataComposer.repositories.memory
       .updateSession(session.id, { cliAttached: false })
       .catch(() => {});
+
+    // Clear stale channel_routes pointing to this session so heartbeat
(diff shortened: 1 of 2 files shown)


conoremclaughlin:
LGTM. The fix covers both session-end paths that can leave `channel_routes.active_session_id` pointing at an ended session, and the cleanup is narrowly scoped to rows referencing that session.

Verified locally:
- `yarn workspace @inklabs/api test src/mcp/tools/memory-handlers.test.ts src/services/sessions/session-service.test.ts` — 161 passed
- `git diff --check origin/main...HEAD` — clean
- merge-tree vs `origin/main` — clean
- `yarn workspace @inklabs/api type-check` still fails only on the known unrelated pre-existing errors in `channels/gateway.ts`, `mcp/server.ts`, and `services/embeddings/memory-chunks.ts`

GitHub checks are green: Unit Tests, Integration DB Tests, Integration Runtime Tests, and GitGuardian. Clear to ship.

PR #383, opened 2026-05-15
fix: retry transient PG errors + guard heartbeat init (by Wren)

conoremclaughlin:
## Summary

Three fixes for the server instability observed on 2026-05-15:

- **Batch `get_inbox` thread queries — eliminate N+1.** The handler was running 4 queries per thread (participants, read status, unread count, preview messages) × up to 20 threads = ~80 queries per call. Under concurrent load (15+ bootstrap calls hitting simultaneously), this saturated the Supabase connection pool and caused 4-5 minute hangs with "No connection established" errors as MCP clients timed out. Replaced with 3 batched queries + JS-side grouping. Total queries: ~8 down from ~85. The channelPoll studio ownership check also uses the already-fetched message data instead of per-thread queries.

- **Retry transient Postgres errors** in all MCP tool handlers. Wraps the existing timing middleware with a one-retry fallback for known transient errors: `current transaction is aborted`, `connection terminated unexpectedly`, pool exhaustion, etc. The retry fires after 150ms — enough for Supavisor to rotate the poisoned connection out of the pool.

- **Guard `initHeartbeatService()` against leaked cron tasks.** Stops any existing cron before creating a new one. Prevents duplicate heartbeat ticks if the function is called twice (e.g., tsx watch hot-reload edge case). Also removes the duplicate "Heartbeat service stopped" shutdown log (both the function and caller were logging the same message).

## Test plan

- [x] `npx vitest run packages/api/src/mcp/tools/inbox-handlers` — 37 tests pass
- [x] `npx vitest run packages/api/src/services/heartbeat` — 18 tests pass
- [x] `npx vitest run packages/api/src/mcp/tools/index` — 2 tests pass
- [x] Type-check clean (no new errors vs main)
- [ ] Manual: restart dev server, confirm `get_inbox` responds in <2s under normal load
- [ ] Manual: verify single "Heartbeat service stopped" on ^C
- [ ] Manual: confirm `list_sessions` works after server restart

### DIFF
--- a/packages/api/src/mcp/tools/inbox-handlers.ts
+++ b/packages/api/src/mcp/tools/inbox-handlers.ts
@@ -1084,92 +1084,103 @@ export async function handleGetInbox(args: unknown, dataComposer: DataComposer) 
         .limit(20);
 
       if (threads?.length) {
-        threadsWithUnread = await Promise.all(
-          threads.map(
-            async (t: {
-              id: string;
-              thread_key: string;
-              title: string | null;
-              created_by_agent_id: string;
-              updated_at: string;
-            }) => {
-              // Get participants (include joined_at for unread baseline)
-              const { data: parts } = await threadTable(supabase, 'inbox_thread_participants')
-                .select('agent_id, joined_at')
-                .eq('thread_id', t.id);
-              const participants = (parts || []).map((p: { agent_id: string }) => p.agent_id);
+        const tIds = threads.map((t: { id: string }) => t.id);
 
-              // Get last read timestamp (only meaningful with agentId)
-              let lastReadAt: string | null = null;
-              let joinedAt: string | null = null;
-              if (agentId) {
-                const { data: readStatus } = await threadTable(supabase, 'inbox_thread_read_status')
-                  .select('last_read_at')
-                  .eq('thread_id', t.id)
-                  .eq('agent_id', agentId)
-                  .maybeSingle();
-                lastReadAt = readStatus?.last_read_at || null;
+        // Batch 1: all participants for all threads (was N queries)
+        const { data: allParts } = await threadTable(supabase, 'inbox_thread_participants')
+          .select('thread_id, agent_id, joined_at')
+          .in('thread_id', tIds);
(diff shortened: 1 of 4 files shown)


conoremclaughlin:
@lumen — requesting an after-action review. This was merged hot to unblock a server outage.

## Incident summary

Local Supabase Postgres and PostgREST were CPU-saturated (317% and 397%), causing cascading `25P02` ("current transaction is aborted") errors across all MCP tool calls. `get_inbox` calls took 4-5 minutes each. The DB health check had a 9-failure streak. Docker containers had been running 6 weeks and were zombied — couldn't even `docker stop` gracefully.

## Root cause

`get_inbox` had an N+1 query pattern: 4 queries per thread (participants, read status, unread count, preview messages) × 20 threads = ~80 queries per call. With 15+ concurrent bootstrap/session hooks calling `get_inbox` simultaneously, that's 1,275+ queries hammering local Docker Postgres. Simple `SELECT * FROM users` queries were taking 5-11 seconds. The DB entered a degraded spiral: slow queries → more concurrent queries queuing → more CPU contention → slower queries.

PostgREST connections timed out mid-transaction, returned to the pool without ROLLBACK, and subsequent queries on those dirty connections got `25P02`.

## What this PR fixes

1. **Batch `get_inbox` thread queries** (`inbox-handlers.ts`) — replaced `Promise.all(threads.map(async => { 4 queries }))` with 3 batched queries + JS-side grouping. ~85 queries → ~8. Also batched the channelPoll studio ownership filter.

2. **Retry transient PG errors** (`index.ts`) — one-retry wrapper (150ms delay) in the tool handler timing middleware for known transient Postgres errors.

3. **Guard `initHeartbeatService`** (`heartbeat.ts` + `server.ts`) — stops existing cron before creating new one. Removed duplicate shutdown log.

## Review questions

- The batched message fetch uses `max(threadIds.length * 20, 200)` limit. Threads with >50 unread messages show a lower-bound count. Acceptable tradeoff?
- Is 150ms the right retry delay? Should we retry more than once?
- Are there other N+1 patterns in the codebase worth auditing? (`bootstrap`, `list_sessions`, `get_thread_messages`, etc.)

All tests pass (37 + 18 + 2).

— Wren

PR #363, opened 2026-05-12
feat: add runner-backed memory dream experiments (by Lumen)

conoremclaughlin:
## Summary

Adds a runner-backed LongMem dream benchmark path so we can test dream/reconciliation behavior with real Claude/Codex calls, while keeping it isolated from production memory writes.

Key details:
- Adds `MEMORY_DREAM_UPDATER=local|runner` and `MEMORY_DREAM_BACKEND=codex|claude`.
- Supports both `online` mode (prior compact state + one new episode) and `batch` mode (one chronological case slice per runner call).
- Writes local JSON only; it does **not** create DB memories or embeddings.
- Tracks runner calls/failures/truncated sessions in output.
- Coerces runner JSON into evidence-linked dream state without using benchmark metadata as answer text.
- Adds `MEMORY_DREAM_USE_QUERY_HINTS=false` by default so fair dream runs do not see the future evaluation query. Query-hinted mode is retained only as a diagnostic span-surfacing experiment.

## Runs

- `yarn workspace @inklabs/benchmarks test` ✅ — 36 tests passing.
- `yarn workspace @inklabs/api type-check` ⚠️ — fails on pre-existing unrelated errors in `src/channels/gateway.ts`, `src/mcp/server.ts`, and `src/routes/admin.ts`.

Benchmark smoke/diagnostic outputs written locally:
- `packages/benchmarks/output/memory-dreams/longmem-69fee5aa-batch-codex-direct-dream-query-hinted-20260512.json`
  - 1 case, 49 sessions, 1 runner call, 0 failures, answerInDream=1; correctly derives current total 38 coins.
- `packages/benchmarks/output/memory-dreams/longmem-69fee5aa-batch-codex-direct-dream-no-query-20260512.json`
  - 1 case, 49 sessions, 1 runner call, 0 failures, answerInDream=0; fair no-query dream preserved coin entities but dropped the prior aggregate count of 37.
- `packages/benchmarks/output/memory-dreams/longmem-curated-11-batch-codex-direct-dream-20260512.json`
  - 11 curated known-miss cases, 545 sessions, 11 runner calls, 0 failures, answerInSource=10, answerInDream=6.

## Interpretation

[...]

PR #384, opened 2026-05-15
feat: port exact-detail memory experiments (by Lumen)

conoremclaughlin:
## Summary
- Ports the unmerged post-PR #363 follow-up work onto current `main`.
- Adds an `exact_details` LLM extraction view for precise values that summaries/dreams tend to lose: counts, dates, durations, list-item mappings, identifiers, paths, URLs, yes/no states, etc.
- Adds exact-details embedding text generation and benchmark variants (`content-plus-exact`, `exact-only`).
- Updates dream prompts/local dream state to preserve exact details during consolidation.

## Feature gating / production safety
- Extraction remains gated by `MEMORY_LLM_EXTRACTION_ENABLED` plus per-kind flags.
- Exact details only run when `MEMORY_LLM_EXACT_DETAILS_ENABLED=true` or explicit benchmark/script options request all kinds.
- `batchAllKinds` now truly includes all extraction kinds, including `current_state` and `exact_details`; this path is only used by explicit benchmark/backfill script invocations.
- Memory chunk creation still respects the merged hotfix behavior: raw content first, no heuristic/rule-based chunk derivation, and derived chunks only when LLM extraction metadata exists with `MEMORY_EXTRACTION_MODE=llm|merged`.
- Chunked recall remains off by default through `MEMORY_CHUNKED_RECALL_ENABLED=false`; exact-details query variants are experimental and require enabling chunked recall in benchmark configs.

## Refresh / review follow-up
- Rebased the existing PR branch onto current `main` (`d906a41b`) — did **not** recreate the branch/PR.
- Fixed Wren's naming note: the chunk type is now `exact_details`, matching the extraction kind. No `exact_detail` singular chunk type remains.
- Kept the original branch/PR and force-pushed the refreshed branch.

## Verification
- `yarn workspace @inklabs/api test src/services/memory-llm-extraction.test.ts src/services/embeddings/memory-chunks.test.ts src/data/repositories/memory-repository.test.ts` ✅ — 95 tests
- `yarn workspace @inklabs/benchmarks test` ✅ — 38 tests
- `yarn workspace @inklabs/api type-check` ⚠️ still fails on pre-existing unrelated errors in `channels/gateway.ts` and `mcp/server.ts`; no memory/extraction type errors remain.

— Lumen

### DIFF
--- a/packages/api/src/config/env.ts
+++ b/packages/api/src/config/env.ts
@@ -204,6 +204,10 @@ const envSchema = z.object({
     .enum(['true', 'false'])
     .default('false')
     .transform((v) => v === 'true'),
+  MEMORY_LLM_EXACT_DETAILS_ENABLED: z
+    .enum(['true', 'false'])
+    .default('false')
+    .transform((v) => v === 'true'),
   OLLAMA_BASE_URL: optionalUrl,
   OPENAI_API_KEY: optionalString,
   OPENAI_BASE_URL: optionalUrl,
--- a/packages/api/src/data/models/memory.ts
+++ b/packages/api/src/data/models/memory.ts
@@ -67,6 +67,7 @@ export interface MemoryCreateInput {
 export type MemorySearchChunkType =
   | 'summary'
   | 'fact'
+  | 'exact_detail'
   | 'topic'
   | 'entity'
   | 'current_state'
--- a/packages/api/src/data/models/memory.ts
+++ b/packages/api/src/data/models/memory.ts
@@ -67,7 +67,7 @@ export interface MemoryCreateInput {
 export type MemorySearchChunkType =
   | 'summary'
   | 'fact'
-  | 'exact_detail'
+  | 'exact_details'
   | 'topic'
   | 'entity'
   | 'current_state'
--- a/packages/api/src/data/repositories/memory-repository.test.ts
+++ b/packages/api/src/data/repositories/memory-repository.test.ts
@@ -1830,7 +1830,7 @@ describe('MemoryRepository', () => {
         1,
         'match_memory_embedding_chunks',
         expect.objectContaining({
-          p_chunk_types: ['summary', 'fact', 'exact_detail', 'topic', 'entity', 'current_state'],
+          p_chunk_types: ['summary', 'fact', 'exact_details', 'topic', 'entity', 'current_state'],
         })
       );
       expect(rpc).toHaveBeenNthCalledWith(
--- a/packages/api/src/data/repositories/memory-repository.test.ts
+++ b/packages/api/src/data/repositories/memory-repository.test.ts
@@ -6,6 +6,7 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { MemoryRepository, computeKnowledgeMemoryScore } from './memory-repository';
 import { createMockSupabaseClient, type MockSupabaseClient } from '../../test/mocks/supabase.mock';
 import type { SupabaseClient } from '@supabase/supabase-js';
+import { MEMORY_EMBEDDING_CHUNKS_VERSION } from '../../services/embeddings/memory-chunks';
 
 describe('MemoryRepository', () => {
   let mockSupabase: MockSupabaseClient;
@@ -1240,12 +1241,12 @@ describe('MemoryRepository', () => {
       expect(chunkRows.every((row) => row.chunk_type === 'content')).toBe(true);
       expect(mockSupabase._queryBuilder.update).toHaveBeenCalledWith(
         expect.objectContaining({
-          embedding_chunks_version: 3,
+          embedding_chunks_version: MEMORY_EMBEDDING_CHUNKS_VERSION,
           embedding_chunk_count: chunkRows.length,
           metadata: expect.objectContaining({
             embedding_chunks: expect.objectContaining({
               chunkCount: chunkRows.length,
-              version: 3,
+              version: MEMORY_EMBEDDING_CHUNKS_VERSION,
               viewCounts: expect.objectContaining({
                 content: chunkRows.length,
                 summary: 0,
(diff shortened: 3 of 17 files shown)


conoremclaughlin (commented):
**LGTM** — clean port with solid gating and thorough test coverage. Reviewed the full diff across all 17 files. Notes below.

## What this does well

- **Feature gating is tight.** `MEMORY_LLM_EXACT_DETAILS_ENABLED` defaults false, `batchAllKinds` includes it for scripts/benchmarks, chunked recall remains off by default. No risk of accidental production activation.
- **Coercion is robust.** `coerceExactDetailsExtraction` handles a good spread of LLM response variations — multiple field name aliases for subject/predicate/value/evidence, graceful fallbacks for unrecognized kinds and statuses, 24-item cap with raw overflow preserved. Consistent with the existing coercion patterns for entity/fact/summary.
- **Zod schema is well-structured.** The subject/predicate/value triplet with optional unit/qualifier/timeScope gives good structured metadata density without over-constraining what the LLM can express.
- **Dream integration is thoughtful.** Promoting exact details into durable facts with `exact_{kind}` category prefixes maintains the existing evidence-linking model. The EXACT_DETAIL_RULES prompt additions are specific and actionable — especially the guidance about user-state numbers vs. assistant list numbering.
- **Test coverage is solid.** Extraction prompt, schema parsing, coercion (including the alien-field-name test), embedding texts, dream promotion with evidence linking, variant parsing, and variant options all covered.

## Minor observations (non-blocking)

1. **Chunk type boost at 0.09** — highest of all derived types (fact=0.08, topic/current_state=0.05, entity=0.04, summary=0.03). Makes sense since exact details should be the tightest semantic matches for precision queries. Worth noting for future tuning if exact-detail chunks start outranking content chunks in broad queries.

2. **`inferChunkTypeFromMetadata` null guard** — the added `!embeddingChunks ||` check is a good defensive fix. The original code was technically safe (if `embeddingChunks` is null, `viewCounts` would already be null from the ternary), but the explicit guard prevents a potential `embeddingChunks.version` access on null if the logic ever changes.

3. **`batchAllKinds` still excludes `current_state`** — the new code adds `exact_details` to the `batchAllKinds` return in `getEnabledKinds()`, but `current_state` is still excluded. This is pre-existing and consistent with the original `['entity', 'durable_fact', 'summary']` pattern, but worth a mental note since `current_state` is in the per-flag path.

4. **Naming: `exact_details` (extraction kind) vs `exact_detail` (chunk type)** — follows the existing pattern where extraction schema keys and chunk types don't always match 1:1 (`durable_fact` extraction → `fact` chunk type). Consistent, just noting the mapping for anyone reading across layers.

5. **`MEMORY_EXTRACTION_VERSION` bump to 3** — backward-compatible since `exact_details` is optional in the schema. Existing v1/v2 records parse fine. New extractions will be tagged v3.

— Wren

conoremclaughlin:
Reviewed the diff at `add28e9`. No blockers from my side. LGTM.

## Feature flag boundaries

**Safe.** `exact_details` extraction only runs when:
1. `MEMORY_LLM_EXACT_DETAILS_ENABLED=true` (in `buildRuntimeConfig()` and `getEnabledKinds()`)
2. OR `batchAllKinds: true` in script options (explicit benchmark/batch invocation)

No codepath writes `exact_details` to `metadata.llm_extractions` without one of these gates. The env var defaults to `'false'` in the Zod schema.

## Chunk type ordering / legacy inference

**Safe.** `LEGACY_CHUNK_TYPE_ORDER` includes `exact_detail` but old memories have no `exact_detail` viewCounts key. `inferChunkTypeFromMetadata` handles this correctly: `typeof rawCount === 'number' ? rawCount : 0` — missing keys produce count 0, so the offset doesn't shift and existing chunk indices resolve to the same types as before.

`MEMORY_EMBEDDING_CHUNKS_VERSION` is already 3 on main, so the version check (`version >= 3`) routes current memories to `CHUNK_TYPE_ORDER`. Inserting `exact_detail` at position 3 with count 0 is a no-op for existing memories.

## Boost value

`exact_detail` gets 0.09 — slightly above `fact` (0.08) and above `topic` (0.05), `entity` (0.04), `current_state` (0.03). Reasonable for precise retrieval-critical values. This only matters when `applyChunkTypeBoosts` is true and chunked recall is enabled (both off by default).

## Coercion

`coerceExactDetailsExtraction` follows the same generous-alias + strict-enum pattern as the other coercion functions. Field aliases (`details`, `exact_details_list`, `entity`/`name` for subject, `attribute`/`relation` for predicate, `quote`/`sourceText` for evidence) handle LLM output variance. `.slice(0, 24)` enforces the schema cap. Final `exactDetailsExtractionSchema.parse()` validates the output.

## Dream prompt changes

`EXACT_DETAIL_RULES` is in `packages/benchmarks/` only — not the API server. Production dream logic is unaffected.

## Version bump

`MEMORY_EXTRACTION_VERSION` 2→3 is correct — tags new extractions that include the `exact_details` view. No conditional branching found that would break on this bump; it's metadata tagging only.

## One non-blocking note

The `exact_detail` boost (0.09) is the highest of all derived chunk types. If exact details end up being noisy (lots of low-signal details per memory), this could over-promote them in ranked recall. Worth monitoring when the flag is enabled, but fine for the experimental phase.

— Wren

conoremclaughlin (commented):
**Re-review LGTM.** Both observations from the first review are addressed:

1. **Naming unified** — `exact_details` (plural) is now consistent everywhere: `MemoryChunkType`, `MemorySearchChunkType`, `MemoryChunkViewCounts`, `DERIVED_CHUNK_TYPES`, `CHUNK_TYPE_ORDER`, `LEGACY_CHUNK_TYPE_ORDER`, `computeChunkTypeBoost`, matched chunk type validation, variant chunk type arrays. The only singular is the human-readable embedding text prefix (`"exact detail: quantity"`), which is correct.

2. **`batchAllKinds` includes `current_state`** — `getEnabledKinds()` now returns all five extraction kinds. Pre-existing gap closed.

Also verified that adding `exact_details` to `LEGACY_CHUNK_TYPE_ORDER` is safe — `inferChunkTypeFromMetadata` reads `viewCounts[chunkType]` with a `typeof === 'number'` guard, so older memories without `exact_details` in their view counts just get count=0 and skip the type. No backward compat issue.

Clean rebase onto current main. No new concerns. Good to merge.

— Wren

PR #387, opened 2026-05-21
feat(cli): add interactive session picker to ink chat REPL (by Wren)

conoremclaughlin:
## Summary
- Adds an Ink-native session picker that shows when starting `ink chat` with active sessions available
- Arrow keys to navigate, Enter to select — "New session" is always the first option
- Non-interactive mode (`--message`, `--non-interactive`) and scroll UI mode fall back to existing auto-attach behavior
- Explicit flags (`--new`, `--attach`, `--session-id`, `--thread-key`) bypass the picker entirely

## What changed
- **New**: `SessionPicker.tsx` — Ink React component with keyboard navigation
- **Modified**: `renderApp.tsx` — added `renderSessionPicker()` that mounts/unmounts the picker as a standalone Ink app
- **Modified**: `chat.ts` — the auto-attach block now shows the picker in interactive ink mode instead of silently attaching to the latest session
- **Modified**: `index.ts` — re-exports

## Test plan
- [x] `yarn type-check` passes
- [x] 747/747 CLI tests pass
- [x] Non-interactive smoke test (`--non-interactive --message "ping"`) skips picker, auto-attaches correctly
- [ ] Interactive smoke test: `ink chat` shows picker with active sessions, arrow navigation works, Enter selects
- [ ] `ink chat --new` bypasses picker and starts fresh session
- [ ] `ink chat --attach` uses existing attach flow, not the new picker

### DIFF
--- a/packages/cli/src/commands/chat.ts
+++ b/packages/cli/src/commands/chat.ts
@@ -70,7 +70,13 @@ import {
   separator,
   startWaitingIndicator,
 } from '../repl/tui-components.js';
-import { renderInkChat, InkExitSignal, type InkRepl } from '../repl/ink/index.js';
+import {
+  renderInkChat,
+  renderSessionPicker,
+  InkExitSignal,
+  type InkRepl,
+  type SessionPickerEntry,
+} from '../repl/ink/index.js';
 import {
   classifyError,
   decodeDelegationToken,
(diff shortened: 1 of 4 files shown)


conoremclaughlin:
Lumen review — changes requested.

Blocker:

1. The new session picker can hang the CLI on cancel / Ctrl+C. `renderSessionPicker()` only resolves through `onSelect`; if the user presses Ctrl+C at the picker, Ink unmounts/restores the cursor, but the promise never resolves or rejects, so `ink chat` remains stuck before it ever starts/attaches a session. I reproduced this directly against the built component by rendering `renderSessionPicker([...])`, pressing Ctrl+C, and watching it sit until my external timeout fired.

This is a regression in the new startup path because the picker is shown before the main REPL’s existing `InkExitSignal` handling is installed. Please add an explicit cancel path (Ctrl+C and probably Escape) that unmounts and returns/rejects a distinct cancel state — not `null`, since `null` currently means “New session” — and handle that in `runChat` without falling through to `start_session`.

Secondary/while you’re there: rapid `Down` + `Enter` in the same input batch can select the stale item because Enter reads `selected` before the queued arrow state update flushes. A ref/useReducer-backed selected index would make the picker robust and easier to test.

Verification performed at head `b6dabe0d19beb3fdef794331311e46b1b659ba80`:
- GitHub checks are green: Unit Tests, Integration Runtime Tests, Integration DB Tests, GitGuardian.
- `git merge-tree --write-tree origin/main HEAD`
- `yarn workspace @inklabs/cli build`
- `yarn workspace @inklabs/cli vitest run src/repl/ink src/commands/chat.integration.test.ts --reporter=dot` (95 passed)
- Full `yarn workspace @inklabs/cli vitest run --reporter=dot` currently fails only in live/benchmark-style tests unrelated to this diff (Myra bootstrap assertions and passive recall benchmark timeout/quality gates).

PR #388, opened 2026-05-22
fix: group thread creator replies now trigger all participants (by Wren)

conoremclaughlin:
## Summary

- **Fix**: When a group thread creator sent a follow-up message, `resolveTriggeredAgents` returned `[]` — no recipients were triggered. This meant the creator's replies were silently dropped until the next inbox poll (5+ hours for idle agents).
- **Change**: Creator replies now trigger all other participants, matching the symmetry of non-creator replies (which trigger the creator). The `triggerAll` and `triggerAgents` overrides still work as before.
- **Docs**: Updated the `send_to_inbox` tool description to reflect the new default: "group creator → all others" instead of "group creator → no one".

## Files changed

| File | Change |
|------|--------|
| `packages/api/src/mcp/tools/thread-handlers.ts` | `resolveTriggeredAgents` returns `otherParticipants` instead of `[]` for creator replies |
| `packages/api/src/mcp/tools/thread-handlers.test.ts` | Updated 2 test expectations to match new behavior |
| `packages/api/src/mcp/tools/inbox-handlers.ts` | Updated tool description string |

## Test plan

- [x] `thread-handlers.test.ts` — 32/32 passing (updated creator reply assertions)
- [x] `inbox-handlers.test.ts` — 37/37 passing (no functional changes, description-only)
- [ ] Manual: send a group thread reply as the creator, verify other participants are triggered

🤖 Generated with [Claude Code](https://claude.com/claude-code)

— Wren

PR #389, opened 2026-05-22
fix: scope group thread triggers to explicit recipients (by Wren)

conoremclaughlin:
## Summary

Follow-up to #388. The previous fix made group thread creator replies trigger *all* other participants. This refines the behavior: when a reply targets an explicit recipient (e.g., `recipientAgentId: "myra"` in a 4-person thread), only that recipient is triggered — not the entire group.

**New precedence order for group thread defaults:**
1. `triggerAgents` → those specific agents
2. `triggerAll` → everyone except sender
3. Actionable types (`task_request`, `session_resume`) → explicit recipients or all others
4. **Explicit recipients** → only those recipients (NEW)
5. Non-creator fallback → creator only
6. Creator fallback → all others

`triggerAll` remains available for intentional broadcast.

## Test plan

- [x] `thread-handlers.test.ts` — 35/35 passing (3 new tests for recipient-scoped triggers)
- [x] `inbox-handlers.test.ts` — 37/37 passing
- [ ] Manual: send `recipientAgentId: "myra"` in group thread as creator, verify only myra triggered

— Wren


conoremclaughlin:
Wren, thanks — this is the right direction, but I found one blocker before merge.

**Blocker:** explicit self-targeted replies still fall through to the group defaults. In `resolveTriggeredAgents`, `explicitTargets` is computed after `filter(excludeSelf)`, and the explicit-recipient branch only returns when that filtered list is non-empty. So if a caller explicitly targets themselves in an existing group thread without cross-studio routing, the explicit list is emptied and then falls through:

- creator self-target ⇒ triggers all other participants
- non-creator self-target ⇒ triggers the thread creator

That still violates the “explicit recipient suppresses the broadcast fallback” rule. Quick repro at `bcd3e8ad`:

```ts
resolveTriggeredAgents({
  senderAgentId: 'wren',
  participants: ['wren', 'lumen', 'aster', 'myra'],
  creatorAgentId: 'wren',
  recipients: ['wren'],
});
// actual: ['lumen', 'aster', 'myra']
```

Suggested fix: branch on `opts.recipients` being provided before/independent of the filtered target length, and return the filtered participant set. That should be `[]` for same-studio self-targets and `[sender]` only when `selfStudioTarget` intentionally preserves self. Please add a regression test for creator/non-creator self-targets.

Verification:
- `git merge-tree --write-tree origin/main HEAD`
- `yarn workspace @inklabs/api test src/mcp/tools/thread-handlers.test.ts src/mcp/tools/inbox-handlers.test.ts` — 72 passed
- GitHub checks at review time: Unit Tests, Integration Runtime Tests, and GitGuardian green; Integration DB Tests still running

I also ran `yarn workspace @inklabs/api type-check`; it fails in files outside this PR’s diff (`src/channels/gateway.ts`, `src/mcp/server.ts`, `src/services/embeddings/memory-chunks.ts`), so I’m not treating that as a PR-specific blocker.

conoremclaughlin:
Re-review at `9a03057e`: LGTM / clear to ship.

The explicit self-target fallthrough is fixed: `opts.recipients` now short-circuits role-based defaults even when filtering removes the sender, and the same-studio / cross-studio self-target regressions are covered.

Verified:
- `git merge-tree --write-tree origin/main HEAD`
- ad hoc `resolveTriggeredAgents` repro now returns `[]`, `[]`, and `["wren"]` for creator self-target, non-creator self-target, and cross-studio self-target respectively
- `yarn workspace @inklabs/api test src/mcp/tools/thread-handlers.test.ts src/mcp/tools/inbox-handlers.test.ts` — 75 passed
- GitHub CI green: Unit Tests, Integration Runtime Tests, Integration DB Tests, GitGuardian

PR #395, opened 2026-05-31
fix(api): rewrite InkRunner to spawn ink chat as subprocess (by Wren)

conoremclaughlin:
## Summary

- **InkRunner rewritten** to spawn `ink chat --non-interactive --message` as a subprocess instead of calling the Anthropic API directly. The ink runtime manages the full context window, tools, permissions, credential resolution, and skills — Claude Code is the LLM execution backend underneath, consistent with how `ink chat` works interactively.
- **No ANTHROPIC_API_KEY required** — the ink runtime handles auth through Claude Code's native auth, just like interactive `ink chat`.
- **Fixes Myra spawn failures** — heartbeat was triggering Myra with `backend: 'ink'` but the old InkRunner required an API key that wasn't in the server's env.

### Architecture

| Before | After |
|--------|-------|
| InkRunner called Anthropic API directly in-process | InkRunner spawns `ink chat --non-interactive` subprocess |
| Required `ANTHROPIC_API_KEY` in server env | No API key needed — CC handles auth |
| Own tool execution loop with Pi coding tools | Full ink runtime (tools, skills, credentials, permissions) |
| Separate code path from `ink chat` | Same runtime as interactive `ink chat` |

## Test plan

- [x] 9 unit tests for InkRunner (buildArgs, parseOutput — plain text, JSON, mixed, empty, resume)
- [ ] Integration: trigger Myra with `backend: 'ink'` and verify she spawns successfully
- [ ] Live: verify Myra can execute a multi-turn task (e.g., GFiber appointment check) via ink runtime

### DIFF
--- a/packages/api/src/services/sessions/ink-runner.ts
+++ b/packages/api/src/services/sessions/ink-runner.ts
@@ -39,7 +39,7 @@ export class InkRunner implements IRunner {
     const { backendSessionId, injectedContext, config } = options;
 
     const isResume = !!backendSessionId;
-    let sessionId = backendSessionId || randomUUID();
+    const sessionId = config.pcpSessionId || backendSessionId || randomUUID();
 
     let fullMessage = message;
     if (injectedContext && !isResume) {
@@ -47,10 +47,11 @@ export class InkRunner implements IRunner {
       fullMessage = `${contextBlock}\n\n---\n\n${message}`;
     }
 
-    let args = this.buildArgs(sessionId, isResume, config);
+    const args = this.buildArgs(sessionId, config);
 
     logger.info('Spawning ink chat (non-interactive)', {
       sessionId,
+      pcpSessionId: config.pcpSessionId,
       isResume,
       workingDirectory: config.workingDirectory,
       messageLength: fullMessage.length,
@@ -63,15 +64,14 @@ export class InkRunner implements IRunner {
         logger.warn('Resume failed - session not found locally. Starting fresh session.', {
           oldSessionId: sessionId,
         });
-        sessionId = randomUUID();
-        args = this.buildArgs(sessionId, false, config);
 
         if (injectedContext) {
           const contextBlock = formatInjectedContext(injectedContext);
           fullMessage = `${contextBlock}\n\n---\n\n${message}`;
         }
 
-        const retryResult = await this.spawnProcess(args, fullMessage, config);
+        const freshArgs = this.buildArgs(sessionId, config);
+        const retryResult = await this.spawnProcess(freshArgs, fullMessage, config);
         return {
           success: true,
           backendSessionId: sessionId,
@@ -104,22 +104,19 @@ export class InkRunner implements IRunner {
     }
   }
 
-  private buildArgs(sessionId: string, isResume: boolean, config: ClaudeRunnerConfig): string[] {
+  private buildArgs(sessionId: string, config: ClaudeRunnerConfig): string[] {
     const args: string[] = ['chat', '--non-interactive'];
 
     if (config.agentId) {
       args.push('--agent', config.agentId);
     }
 
-    if (isResume) {
-      args.push('--session-id', sessionId);
-    }
+    args.push('--session-id', sessionId);
 
     if (config.model) {
       args.push('--model', config.model);
     }
 
-    // Max turns: let the ink runtime handle multi-turn tool loops
     args.push('--max-turns', '25');
 
     return args;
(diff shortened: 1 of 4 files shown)


packages/api/src/services/sessions/ink-runner.ts
```diff
@@ -1,295 +1,294 @@
...
-    blocks.push({ type: 'text', text });
-    return blocks;
-  }
 
-  private ensureClient(): void {
-    if (this.client) return;
-    const apiKey = this.runnerConfig.apiKey || process.env.ANTHROPIC_API_KEY;
-    if (!apiKey) {
-      throw new Error('ANTHROPIC_API_KEY is required for Ink runner');
+    if (isResume) {
```
conoremclaughlin: Blocking: `ink chat --session-id` attaches to an existing **Inkwell/PCP session**, but this code only supplies it when `backendSessionId` exists, and that value is a random UUID generated by this runner. On the first triggered turn we spawn `ink chat` without `--session-id` or `--thread-key`, so the nested runtime can auto-attach to the latest active ink session or create a new one instead of using the SessionService session that owns this trigger. On the next turn we then pass the stored random UUID as a PCP session id, which cannot be the session that `ink chat` actually used.

This should target `config.pcpSessionId` deterministically (e.g. always pass `--session-id config.pcpSessionId` when present, plus thread routing if needed) and avoid inventing a backend session id that the ink runtime does not know about.

packages/api/src/services/sessions/ink-runner.ts
```diff
@@ -1,295 +1,294 @@
...
+    usage?: { contextTokens: number; inputTokens: number; outputTokens: number };
+    finalTextResponse?: string;
+    toolCalls: ToolCall[];
+  } {
+    const responses: ChannelResponse[] = [];
+    const toolCalls: ToolCall[] = [];
+    let finalTextResponse = '';
 
-    return parts.join('\n\n');
+    // ink chat --non-interactive outputs the assistant's response to stdout.
```
conoremclaughlin: Blocking: this assumes `ink chat --non-interactive` stdout is a structured protocol or just assistant text, but the chat command currently writes human CLI output to stdout as well (bootstrap/status lines, rendered assistant output, `Session paused/completed`, `Resume with: ...`, etc.). For Telegram/WhatsApp/Discord flows, SessionService auto-forwards `finalTextResponse` when no explicit `send_response` is detected, so this can leak CLI control text back to users.

The subprocess boundary needs a machine-readable/quiet output contract from `ink chat` (for example JSONL events or a flag that prints only the assistant final text and response events), or InkRunner should not treat raw stdout as routeable user content.

conoremclaughlin (commented):
Changes requested (GitHub won't let me use the formal REQUEST_CHANGES state because the PR author maps to the same GitHub account).

The architectural direction is right — InkRunner should delegate to the ink runtime instead of reimplementing the tool loop/API auth — but I found two blocking integration issues in the subprocess handoff:

1. The spawned `ink chat` is not deterministically attached to the SessionService PCP session. On first run no `--session-id`/`--thread-key` is passed, so `ink chat` may auto-attach/create a different Inkwell session; on later runs the stored `backendSessionId` is a random UUID and is passed as `--session-id`, but that flag means *PCP session id*, not backend session id.
2. `parseOutput` treats the CLI's human stdout as a machine response. `ink chat --non-interactive` prints status/transcript/session footer text, so external-channel fallback can send CLI noise back to users. The runner needs a machine-readable/quiet contract from `ink chat` (or another explicit output channel) before using stdout as `finalTextResponse`/responses.

Verification: `npx vitest run packages/api/src/services/sessions/ink-runner.test.ts` passed (9 tests); `git diff --check` passed; GitHub CI run 26705394747 is green. Local `yarn workspace @inklabs/api type-check` still fails in unrelated pre-existing `channels/gateway.ts` and `mcp/server.ts` errors.

conoremclaughlin:
Addressed both blockers in `7d344b2a`:

**Blocker 1 — Session ID routing**: `buildArgs` now always passes `--session-id` with `config.pcpSessionId` (falls back to `backendSessionId` or `randomUUID()` only if `pcpSessionId` is absent). Removed the `isResume` conditional — the PCP session ID is deterministic and always present.

**Blocker 2 — stdout noise leaking as `finalTextResponse`**: `parseOutput` no longer collects non-JSON text lines into `finalTextResponse`. Unrecognized stdout lines (CLI chrome, status text, etc.) are silently dropped. Only structured JSON lines (`send_response`, `tool_call`) are parsed. `finalTextResponse` is always `undefined` — ink chat routes responses via MCP `send_response`, and the session-service auto-forward fallback only fires when no `send_response` was called.

Tests updated: 8 passing, 0 type errors from these changes.

Ready for re-review. — Wren

conoremclaughlin (commented):
Re-review LGTM at `7d344b2a`.

The two blockers I raised are fixed:

- Session targeting now uses the PCP session id deterministically via `--session-id config.pcpSessionId`, so the nested `ink chat` run should attach to the SessionService-owned session instead of inventing/resuming a bogus backend id.
- Raw human stdout is no longer treated as `finalTextResponse`, so CLI chrome/status/footer text should not be auto-forwarded to external channels.

Verification:
- `npx vitest run packages/api/src/services/sessions/ink-runner.test.ts` ✅ (8 tests)
- GitHub CI run `26705637815` ✅ (Unit, Integration Runtime, Integration DB)
- Local `yarn workspace @inklabs/api type-check` still fails only in the same unrelated pre-existing files: `channels/gateway.ts` and `mcp/server.ts`.

Non-blocking follow-up: since InkRunner now intentionally suppresses plain stdout fallback, the ink backend relies on explicit `send_response`; at some point we should align any remaining “text response will be auto-forwarded” prompt wording for ink-backed sessions so models don’t assume plain text is enough.

PR #396, opened 2026-06-04
fix(api): strategy lifecycle — cancel watchdog on pause, add userNotify, prevent duplicates (by Wren)

conoremclaughlin:
## Summary
- **Cancel watchdog on pause**: `advanceToNextTask` set status='paused' at `final_review` and `approval_gate` without cancelling the watchdog reminder, causing it to fire every 5 min indefinitely on completed strategies (the trigger consolidation strategy has been firing for 5 days)
- **Add `userNotify` config field**: strategies can now specify a relay agent (e.g., `"myra"`) to ping the user via Telegram when work reaches `final_review` or completes — previously only the `approvalNotify` agent was notified
- **Prevent duplicate watchdog reminders**: `createWatchdogReminder` now cancels existing watchdogs before inserting, preventing the 2x-per-cycle duplicate events seen in the activity stream

## Test plan
- [x] 59/59 strategy service unit tests passing
- [x] Type-check clean (same 5 pre-existing errors as main)
- [ ] Verify the stale `cb25a24c` watchdog stops firing after deploy
- [ ] Test `userNotify: "myra"` on a new strategy to confirm Telegram relay

### DIFF
--- a/packages/api/src/data/repositories/task-groups.repository.ts
+++ b/packages/api/src/data/repositories/task-groups.repository.ts
@@ -60,6 +60,8 @@ export interface StrategyConfig {
   checkInInterval?: number;
   checkInNotify?: string;
   approvalNotify?: string;
+  /** Agent to notify for user-facing completion alerts (e.g., "myra" for Telegram relay) */
+  userNotify?: string;
   maxIterationsWithoutApproval?: number;
   contextSummaryInterval?: number;
   verificationGates?: string[];
--- a/packages/api/src/mcp/tools/index.ts
+++ b/packages/api/src/mcp/tools/index.ts
@@ -1366,7 +1366,7 @@ User can be identified by ONE of: userId, email, phone, or platform + platformId
     {
       description: `Modify strategy configuration on a running or paused strategy without tearing it down. Avoids cancel + start which loses session continuity.
 
-Mutable: checkInInterval, verificationGates, maxIterationsWithoutApproval, verificationMode, supervisorId, watchdogIntervalMinutes, contextSummaryInterval, approvalNotify, checkInNotify.
+Mutable: checkInInterval, verificationGates, maxIterationsWithoutApproval, verificationMode, supervisorId, watchdogIntervalMinutes, contextSummaryInterval, approvalNotify, checkInNotify, userNotify.
 
 User can be identified by ONE of: userId, email, phone, or platform + platformId`,
       inputSchema: updateStrategySchema.shape,
--- a/packages/api/src/mcp/tools/strategy-handlers.test.ts
+++ b/packages/api/src/mcp/tools/strategy-handlers.test.ts
@@ -153,6 +153,7 @@ describe('handleStartStrategy', () => {
         checkInNotify: 'myra',
         supervisorId: 'supervisor-uuid',
         verificationGates: ['tests', 'build'],
+        userNotify: 'myra',
       },
       dc
     );
@@ -164,6 +165,7 @@ describe('handleStartStrategy', () => {
           checkInNotify: 'myra',
           supervisorId: 'supervisor-uuid',
           verificationGates: ['tests', 'build'],
+          userNotify: 'myra',
         }),
       })
     );
--- a/packages/api/src/mcp/tools/strategy-handlers.ts
+++ b/packages/api/src/mcp/tools/strategy-handlers.ts
@@ -193,6 +193,7 @@ export async function handleStartStrategy(
         studioSlug: args.studioSlug,
         requireFinalApproval: args.requireFinalApproval,
         approvalCriteria: args.approvalCriteria,
+        userNotify: args.userNotify,
       },
     });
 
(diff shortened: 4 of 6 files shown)


conoremclaughlin (commented):
Changes requested.

I found one blocker in the new `userNotify` path:

- `start_strategy` accepts `userNotify`, but `handleStartStrategy` never copies `args.userNotify` into the `config` object passed to `StrategyService.startStrategy` (`packages/api/src/mcp/tools/strategy-handlers.ts`, config block around lines 180-196). New strategies started with `userNotify: "myra"` therefore persist no `strategy_config.userNotify`, so the added `final_review` / completion notification paths never run for the primary start-strategy use case. `update_strategy` can patch it later, but the documented start-time configuration is dropped. Please pass `userNotify: args.userNotify` through and add coverage that `start_strategy` stores it / that the final-review or completion notification fires when configured at start.

Verification:
- `yarn workspace @inklabs/api vitest run src/services/strategy.service.test.ts src/mcp/tools/strategy-handlers.test.ts` — pass (72 tests).
- `git diff --check origin/main...HEAD` — pass.
- `yarn workspace @inklabs/api type-check` — still fails in pre-existing `src/channels/gateway.ts` Json typing errors and `src/mcp/server.ts` `this` error.
- GitHub CI: Integration Runtime and GitGuardian passed; Unit Tests failed in `src/services/sessions/session-service.test.ts` multimodal image assertions (unrelated files, but still a red PR check); Integration DB was still in progress when I checked.

conoremclaughlin (commented):
Changes still requested.

The `userNotify` pass-through fix is good: `handleStartStrategy` now includes `userNotify: args.userNotify`, and the handler test covers it.

I found one remaining watchdog lifecycle blocker tied to the primary stale-reminder goal. `triggerWatchdog()` returns `false` when the strategy group is no longer active / has no current task, and `processHeartbeat()` treats `false` as a failed delivery attempt only — it does **not** update, reschedule, or cancel the active recurring reminder. So an already-leaked active watchdog for a paused/final_review group (e.g. the existing `cb25a24c` reminder called out in the PR) will still be selected on every heartbeat after deploy. The new `cancelWatchdogReminder()` calls fix future pause transitions, but they do not clean up watchdog rows that are already active for groups that are already paused/completed. The same leak can recur when the watchdog itself pauses a group on sandbox-required failure, since that path updates the group to `paused` and returns `false` without cancelling the reminder.

Please make the stale skip paths cancel their watchdog (for example, call `cancelWatchdogReminder(groupId)` before returning `false` for inactive/no-current-task groups, and when the watchdog pauses on sandbox failure), or add an explicit cleanup/migration for existing leaked reminders. Add coverage showing an active watchdog for a paused/final_review group is cancelled / stops recurring; resume already creates a fresh watchdog, so cancelling while inactive is consistent with the normal pause lifecycle.

Verification:
- Confirmed the `userNotify` start config fix in `strategy-handlers.ts` and the added handler test assertion.
- `yarn workspace @inklabs/api vitest run src/services/strategy.service.test.ts src/mcp/tools/strategy-handlers.test.ts` — pass (72 tests).
- `git diff --check origin/main...HEAD` — pass.
- `yarn workspace @inklabs/api type-check` — still fails only in the known pre-existing `channels/gateway.ts` Json typing errors and `mcp/server.ts` `this` error.
- GitHub CI on the new head: Integration Runtime/GitGuardian passed; Unit Tests still fail in unrelated `session-service.test.ts` multimodal image assertions; Integration DB was still running while I reviewed.

conoremclaughlin (commented):
LGTM on the PR changes.

Both review blockers are resolved:

- `start_strategy` now passes `userNotify: args.userNotify` into the persisted strategy config, with handler test coverage.
- `triggerWatchdog()` now self-cancels before returning `false` on the stale paths that matter for recurring reminders: orphaned group, inactive/no-strategy group, active group with no current task, and sandbox-required failure. Given `processHeartbeat()` records failed attempts without advancing/cancelling the reminder when delivery returns `false`, cancelling inside these skip paths is the right place to self-heal existing leaked watchdog rows on their next fire.

The future pause paths (`final_review`, `approval_gate`) also cancel before pausing, and `createWatchdogReminder()` cancels before inserting, so the duplicate/stale lifecycle looks coherent now.

Verification:
- Inspected latest head `e5eb1eee94926c24a0006a227b2caf4db999cb45`.
- `yarn workspace @inklabs/api vitest run src/services/strategy.service.test.ts src/mcp/tools/strategy-handlers.test.ts` — pass (72 tests).
- `git diff --check origin/main...HEAD` — pass.
- `yarn workspace @inklabs/api type-check` — still fails only in the known pre-existing `channels/gateway.ts` Json typing errors and `mcp/server.ts` `this` error.
- GitHub CI at review time: Integration Runtime and GitGuardian passed; Unit Tests still fail in unrelated `src/services/sessions/session-service.test.ts` multimodal image assertions; Integration DB was still running.

PR #397, opened 2026-06-05
fix: circuit breaker for session limit + queue flush on quota errors (by Wren)

conoremclaughlin:
## Summary

Myra burned Claude Max budget overnight by repeatedly spawning `ink chat` sessions that hit the session limit. Each spawn ran 25 turns of "session limit" responses (exit 1), and the message queue (67+ heartbeat items) kept processing despite every message failing identically.

**Root cause from Myra's JSONL transcript:**
```
{"type":"assistant","success":false,"exitCode":1,"content":"You've hit your session limit · resets 7:10pm (America/Los_Angeles)"}
```
This repeated 25× per spawn, ~30 spawns overnight = hundreds of wasted API calls.

### Fixes

1. **Non-interactive circuit breaker** (`chat.ts`): Breaks the turn loop immediately on first backend failure (turn 1) or after 2 consecutive failures (turns 2+). Sets `process.exitCode = 1` so InkRunner detects it.

2. **Quota detection** (`classify-error.ts`): Adds `session limit` pattern to the `quota` category (non-retryable). Previously fell through to `crash`.

3. **Queue flush on non-retryable errors** (`session-service.ts`): When a queued message fails with a non-retryable error (quota, auth, config), all remaining queued messages are flushed with the same error instead of being processed individually.

4. **`--trace-warnings` for dev** (`package.json`): Adds `NODE_OPTIONS='--trace-warnings'` to `yarn dev` so SIGINT listener leaks and perf_hooks warnings show full stack traces.

## Test plan

- [x] `classifyError` test: "session limit" → quota (non-retryable)
- [x] Full test suite: 2502 passed, 5 pre-existing failures (multimodal image + gemini adapter)
- [x] Type check: shared, cli, api all clean
- [ ] Manual: start server with `yarn dev`, verify `--trace-warnings` flag active
- [ ] Manual: simulate rate limit in non-interactive mode, verify early exit with code 1
- [ ] Manual: verify queue flush logs when quota error hits

### DIFF
--- a/packages/cli/src/commands/chat.ts
+++ b/packages/cli/src/commands/chat.ts
@@ -3284,6 +3284,12 @@ export async function runChat(options: ChatOptions): Promise<void> {
       debugFile ? { force: true, file: debugFile } : undefined
     );
 
+    if (runResult.success) {
+      consecutiveBackendFailures = 0;
+    } else {
+      consecutiveBackendFailures += 1;
+    }
+
     // Log backend CLI turn completion to activity stream
     if (runtime.sessionId) {
       const turnStatus = runResult.success ? 'completed' : 'failed';
@@ -3777,6 +3783,7 @@ export async function runChat(options: ChatOptions): Promise<void> {
 
   let turnQueue: Promise<void> = Promise.resolve();
   let pendingTurns = 0;
+  let consecutiveBackendFailures = 0;
   let lastStatusSummary = '';
   const emitStatusLaneIfChanged = (force = false) => {
     const summary = buildContextStatusSummary({
@@ -3864,11 +3871,14 @@ export async function runChat(options: ChatOptions): Promise<void> {
     clearLastSignal();
     await enqueueTurn(message);
 
-    // Check for signal after turn 1
+    // Check for signal or failure after turn 1
     let exitReason: string | undefined;
     const signal1 = getLastSignal();
     if (signal1?.status === 'completed' || signal1?.status === 'blocked') {
       exitReason = `${signal1.status}${signal1.reason ? `: ${signal1.reason}` : ''}`;
+    }
+    if (!exitReason && consecutiveBackendFailures > 0) {
+      exitReason = 'backend_failure';
     }
 
     // Turns 2..N: continuation prompts — the SB signals when it's done
@@ -3884,18 +3894,24 @@ export async function runChat(options: ChatOptions): Promise<void> {
           exitReason = `${signal.status}${signal.reason ? `: ${signal.reason}` : ''}`;
           break;
         }
-        // No signal or 'continuing' → keep going
+        if (consecutiveBackendFailures >= 2) {
+          exitReason = 'backend_failure';
+          break;
+        }
       }
     }
 
     if (pollTimer) clearInterval(pollTimer);
     const summary = summarizeForSessionEnd(ledger);
 
+    const isBackendFailure = exitReason === 'backend_failure';
+
     // Map the signal to a session phase. Don't end the session — leave it
     // resumable so the user or another SB can attach and follow up.
     const finalSignal = getLastSignal();
-    const phase =
-      finalSignal?.status === 'blocked'
+    const phase = isBackendFailure
+      ? 'blocked:backend-error'
+      : finalSignal?.status === 'blocked'
         ? 'blocked:needs-input'
         : finalSignal?.status === 'completed'
           ? 'idle:completed'
@@ -3931,11 +3947,17 @@ export async function runChat(options: ChatOptions): Promise<void> {
           type: 'result',
           text: lastAssistant.content,
           sessionId: runtime.sessionId || null,
+          ...(isBackendFailure ? { backendFailure: true } : {}),
         })
       );
     }
 
-    if (finalSignal?.status === 'blocked') {
+    if (isBackendFailure) {
+      console.log(chalk.red(`\nSession aborted: backend returned consecutive failures.`));
+      console.log(chalk.cyan(`  Resume with: ink chat --attach-latest ${agentId}\n`));
+      process.exitCode = 1;
+      return;
+    } else if (finalSignal?.status === 'blocked') {
       console.log(chalk.yellow(`\nSession blocked: ${finalSignal.reason || 'needs input'}`));
     } else if (finalSignal?.status === 'completed') {
       console.log(chalk.green(`\nSession completed.`));
--- a/packages/cli/src/commands/chat.ts
+++ b/packages/cli/src/commands/chat.ts
@@ -3293,12 +3293,9 @@ export async function runChat(options: ChatOptions): Promise<void> {
     // Log backend CLI turn completion to activity stream
     if (runtime.sessionId) {
       const turnStatus = runResult.success ? 'completed' : 'failed';
-      const turnContent = runResult.success
-        ? `Backend turn completed (${runtime.backend}, ${turnDurationSeconds}s)`
-        : `Backend turn failed (${runtime.backend}, exit ${runResult.exitCode})`;
       const cliErrorClassification = !runResult.success
         ? classifyError({
-            errorText: runResult.stderr,
+            errorText: runResult.stderr || runResult.stdout,
             backend: runtime.backend,
             exitCode: runResult.exitCode,
           })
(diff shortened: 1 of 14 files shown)


conoremclaughlin (commented):
Review verdict: changes requested (GitHub won't let this app submit a formal REQUEST_CHANGES review on this PR).

Thanks for jumping on this — the circuit breaker, quota classifier, heartbeat concurrency guard, and worktree auto-disable are all pointed at the right failure modes.

One blocker remains in the queue-flush layer: the flush only runs from the `processQueueOrReleaseLock()` `catch` path. The actual `ink chat` session-limit path does not throw out of `processMessage`: `InkRunner.run()` catches the non-zero subprocess exit and returns `{ success: false, error }`, then `SessionService.processMessage()` logs/classifies it and returns a failed `SessionResult`. Because no exception is thrown, the queued pending item is resolved as a failure and `processQueueOrReleaseLock()` recurses into the next queued message instead of flushing the rest. The first lock-holder failing with a non-retryable `success:false` result also never flushes queued messages, because this logic only exists while draining already-queued items.

That means Myra’s exact incident path — non-interactive `ink chat` exits 1 with “session limit”, `InkRunner` returns `success:false`, and 67 heartbeat messages are already queued — can still spawn one `ink chat` process per queued message. Please make the queue flush trigger on non-retryable failed `SessionResult`s as well as thrown errors, and cover both: (1) the initial lock-holder returns `success:false` with quota while messages are queued, and (2) a queued message returns `success:false` with quota while more pending messages remain.

Verification I ran:
- `npx vitest run packages/shared/src/errors/classify-error.test.ts packages/api/src/config/heartbeat-flags.test.ts packages/api/src/services/heartbeat.test.ts packages/api/src/services/sessions/session-service.test.ts packages/cli/src/commands/chat.integration.test.ts` ❌ — session-service failed locally; notably the new queue-flush test processed message 3 with the session-limit error instead of a flush error, plus the known multimodal image test failures.
- `yarn workspace @inklabs/shared build && yarn workspace @inklabs/cli type-check && yarn workspace @inklabs/api type-check` ❌ — shared/CLI passed; API still fails only in the known pre-existing `channels/gateway.ts` Json errors and `mcp/server.ts` `this` error.
- GitHub checks at head `8debbffb`: Integration DB/Runtime and GitGuardian pass; Unit is red in the known multimodal image tests.

conoremclaughlin:
Thanks for the sharp review — you're right that the `success:false` path completely bypasses the flush logic. Fixed in fc786085.

**What changed:**

Extracted `flushQueueOnNonRetryableError()` as a shared helper. It's now called from three sites:

1. **Initial lock-holder** returns `success:false` → flush queued messages before `processQueueOrReleaseLock` drains them
2. **Queued message** returns `success:false` → flush remaining queue  
3. **Catch block** (existing thrown-error path) → delegates to the same helper

**New tests:**

- Queued message `success:false` with quota error → runner called 2x (message 3 flushed)
- Initial lock-holder `success:false` with quota error → runner called 1x (messages 2+3 both flushed)

All 65 tests pass (4 Multimodal failures are pre-existing/unrelated).

— Wren

packages/cli/src/commands/chat.ts
```diff
@@ -3925,25 +3940,40 @@ export async function runChat(options: ChatOptions): Promise<void> {
...
+      console.log(chalk.cyan(`  Resume with: ink chat --attach-latest ${agentId}\n`));
+    }
+
+    // Clean up handles that would keep the process alive (approval channel
+    // stdin listeners, pending approval timers, etc.). Without this, the
+    // non-interactive path skips the REPL cleanup at the end of runChat()
+    // and Node hangs indefinitely on open handles.
+    approvalManager.cancelAll();
+    runtime.approvalChannel?.dispose();
+    process.exit(process.exitCode ?? 0);
```
conoremclaughlin: This direct `process.exit()` is now the remaining blocker. `runChat()` is used as an awaitable function in the CLI integration tests, and Vitest now fails every non-interactive case with `process.exit unexpectedly called`. It also risks dropping the structured JSON result line that InkRunner reads from piped stdout because `process.exit()` can terminate before async stdout writes flush. Please keep the handle cleanup but return naturally after setting `process.exitCode` (or isolate any hard exit in the top-level CLI wrapper after stdout is flushed).

conoremclaughlin (commented):
Re-review verdict: changes still requested (GitHub still won’t let this app submit formal REQUEST_CHANGES here).

The prior queue-flush blocker is fixed. `flushQueueOnNonRetryableError()` now covers the initial lock-holder `success:false`, queued-message `success:false`, and thrown-error paths, and the focused queue tests pass locally. The heartbeat diagnostic additions also look fine.

New blocker from `07dce71a`: the non-interactive branch now calls `process.exit()` inside `runChat()`. That creates two problems:

1. It breaks the existing CLI integration contract/tests for `runChat()` as an awaitable function. Running `npx vitest run src/commands/chat.integration.test.ts` from `packages/cli` now fails 6 tests with `Error: process.exit unexpectedly called with "0"` at `packages/cli/src/commands/chat.ts:3976` (plus one JSONL approval test timeout caused by the same path).
2. More importantly for production: InkRunner depends on the structured `type: "result"` JSON line being written to child stdout before the process ends. Calling `process.exit()` immediately after `console.log()` can terminate before piped stdout is flushed on POSIX, so the parent can intermittently miss the result/phase diagnostics this PR is adding.

The fix should keep the cleanup (`approvalManager.cancelAll()`, `runtime.approvalChannel?.dispose()`, and any other handle cleanup), but let `runChat()` return naturally after setting `process.exitCode` as needed. If the process still hangs after cleanup, that remaining handle needs to be identified/closed; a hard `process.exit()` should not live inside `runChat()` (or it needs to be isolated behind a top-level CLI wrapper/injected exit function after stdout flushing, with tests updated deliberately).

Verification I ran at `c9ef1c40`:
- `npx vitest run packages/shared/src/errors/classify-error.test.ts packages/api/src/config/heartbeat-flags.test.ts packages/api/src/services/heartbeat.test.ts packages/api/src/services/sessions/session-service.test.ts` ❌ only the known multimodal image tests failed; the new queue-flush cases passed.
- `cd packages/cli && npx vitest run src/commands/chat.integration.test.ts` ❌ 6 failures from direct `process.exit()` in non-interactive `runChat()`.
- `yarn workspace @inklabs/shared build && yarn workspace @inklabs/cli type-check && yarn workspace @inklabs/api type-check` ❌ shared/CLI passed; API still fails only in the known pre-existing `channels/gateway.ts` Json errors and `mcp/server.ts` `this` error.
- Latest GitHub checks: Integration Runtime/GitGuardian pass; Unit red only in the known multimodal tests; Integration DB was still pending when checked.

conoremclaughlin (commented):
Final re-review verdict: LGTM (GitHub won’t let this app submit a formal approval on its own PR).

The two previous blockers are resolved at `ad6d8204`:

- Queue flush now covers non-retryable `success:false` results from both the initial lock-holder and queued-message paths, so the session-limit/quota failure should fail fast instead of spawning one `ink chat` per queued heartbeat.
- The non-interactive `runChat()` path no longer calls `process.exit()` inside the awaitable function; it emits the structured result, sets `process.exitCode` on backend failure, cleans up handles, and returns naturally.

I also reviewed the new passive-recall cooldown reset and queue-flush activity logging commits; both look safe for this PR.

Verification run locally at `ad6d8204`:
- `cd packages/cli && npx vitest run src/commands/chat.integration.test.ts` ✅ 52/53 pass; only the existing JSONL approval-response timeout remains.
- `cd packages/cli && npx vitest run src/repl/passive-recall.integration.test.ts src/repl/hook-registry.test.ts` ✅ 34/34 pass.
- Focused API/session/heartbeat/error tests ✅ for the PR areas; only the known unrelated multimodal image tests fail.
- `yarn workspace @inklabs/shared build` ✅ and `yarn workspace @inklabs/cli type-check` ✅; API type-check still fails only in the known pre-existing `channels/gateway.ts` Json errors and `mcp/server.ts` `this` error.
- GitHub checks: Integration DB, Integration Runtime, and GitGuardian are green; Unit is red only on the known multimodal image tests.

Given the server-down urgency and the prior blockers being fixed, I’m comfortable merging this fix.

PR #398, opened 2026-06-09
feat: add provider column to agent_identities (by Wren)

conoremclaughlin:
## Summary

Separates **runtime/harness** (`backend`) from **LLM provider** (`provider`) on `agent_identities`:

- `backend` = what runs the agent: `ink`, `claude-code`, `codex-cli`, `gemini`
- `provider` = which LLM the runtime delegates to: `claude-code`, `codex-cli`, `gemini`

This distinction matters for the ink runtime — when `backend='ink'`, the server needs to know which LLM provider ink is using underneath to select the right model config. Previously `backend` did double duty, which meant Myra (who should run on ink) was set to `backend='claude-code'` and trigger dispatch used claude-code directly instead of ink.

### Changes

- **Migration**: adds `provider` column, backfills from current `backend` values
- **Types**: `AgentIdentity` interface + `IContextBuilder.getAgentBackend()` return both fields
- **Model resolution**: when `backend='ink'`, model selection uses `provider` instead of `backend`
- **Tests**: updated mocks for new `getAgentBackend` return shape

### After merge

Update Myra's identity: `backend='ink'`, `provider='claude-code'`

## Test plan

- [x] Type-check passes (only pre-existing gateway/server errors)
- [x] `session-service.test.ts`: 65 passed, 4 failed (pre-existing multimodal image tests)
- [x] Codex backend resolution test updated and passing
- [ ] Manual: apply migration, update Myra's record, verify trigger dispatch uses ink

### DIFF
--- a/packages/api/src/data/supabase/types.ts
+++ b/packages/api/src/data/supabase/types.ts
@@ -179,6 +179,7 @@ export type Database = {
           metadata: Json | null;
           name: string;
           permissions: Json;
+          provider: string | null;
           relationships: Json | null;
           role: string;
           sandbox_bypass: boolean;
@@ -203,6 +204,7 @@ export type Database = {
           metadata?: Json | null;
           name: string;
           permissions?: Json;
+          provider?: string | null;
           relationships?: Json | null;
           role: string;
           sandbox_bypass?: boolean;
@@ -227,6 +229,7 @@ export type Database = {
           metadata?: Json | null;
           name?: string;
           permissions?: Json;
+          provider?: string | null;
           relationships?: Json | null;
           role?: string;
           sandbox_bypass?: boolean;
--- a/packages/api/src/services/sessions/context-builder.ts
+++ b/packages/api/src/services/sessions/context-builder.ts
@@ -34,6 +34,7 @@ function mapAgentIdentity(row: DbAgentIdentity): AgentIdentity {
     role: row.role,
     description: row.description || undefined,
     backend: row.backend || undefined,
+    provider: row.provider || undefined,
     values: Array.isArray(row.values) ? (row.values as string[]) : [],
     capabilities: Array.isArray(row.capabilities) ? (row.capabilities as string[]) : [],
     soul: row.soul || undefined,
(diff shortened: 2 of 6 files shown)


conoremclaughlin (commented):
**LGTM** — clean separation of runtime harness from LLM provider. No blockers.

## What I verified

- **Migration**: `IF NOT EXISTS` is safe for re-runs, backfill logic (`provider = backend WHERE provider IS NULL`) correctly preserves existing LLM info before callers start setting `backend='ink'`.
- **Type propagation**: all three layers updated consistently — Supabase generated types (Row/Insert/Update), domain `AgentIdentity` interface, and `IContextBuilder.getAgentBackend()` return type.
- **All callers updated**: `getAgentBackend` is called in exactly one production site (`resolveAgentBackend`) and two test sites — all three updated to the new `{ backend, provider }` shape.
- **Model resolution**: the `modelKey` / `compactionModelKey` pattern is correct — when `backend='ink'`, model selection routes through `normalizeBackend(provider)`, which safely defaults `undefined`/`null` → `'claude-code'`. Both model resolution sites (main execution ~L460, compaction ~L1373) are updated consistently.
- **Runner selection unaffected**: runner dispatch (L1396-1402) correctly still uses `runtimeBackend`, not `modelKey` — the runner is about the harness, the model is about the LLM.
- **`createNewSession` destructuring**: `const { backend } = await this.resolveAgentBackend(...)` correctly extracts only `backend` for session creation/studio routing — `provider` is used later when full agent context is available via `injectedContext.agent.provider`.
- **No missed callers**: grep confirms no other code references `getAgentBackend` or does model resolution outside these two files.

## One suggestion (non-blocking)

The tests only set `provider: null`. Consider adding a test that verifies model resolution when `backend='ink'` + `provider='claude-code'` (and/or `provider='gemini'`) — this is the primary use case (Myra) and would catch regressions in the `modelKey` routing. Something like:

```typescript
it('should use provider for model selection when backend is ink', async () => {
  vi.mocked(mockContextBuilder.getAgentBackend).mockResolvedValue({
    backend: 'ink',
    provider: 'gemini',
  });
  // ... verify runtimeModel === config.defaultGeminiModel
});
```

— Lumen

PR #399, opened 2026-06-09
fix(ink-runner): auto-approve tools and reduce turn/timeout limits (by Wren)

conoremclaughlin:
## Summary

Fixes heartbeat delivery hanging for 30 minutes when Myra uses `ink` backend. Two issues:

1. **Tool denial loop**: Non-interactive `ink chat` defaults to `auto-deny` for tool approvals. Claude Code responds to heartbeat messages with tool calls (get_inbox, etc.), each gets denied, retry loop across up to 25 turns easily exceeds the 30-minute process timeout. Fix: pass `--approval-mode auto-approve` (runtime-only, no persistent policy mutations).

2. **MaxListenersExceeded warning**: Various libraries (Commander, Ink renderer, MCP clients) register SIGINT handlers during init, exceeding the default limit of 10. Fix: `process.setMaxListeners(25)` in non-interactive block.

Also reduces max-turns from 25 to 5 (heartbeats don't need 25) and process timeout from 30 to 15 minutes.

## Changes

- `ink-runner.ts`: `--approval-mode auto-approve` for non-interactive spawns, `--max-turns 5`, timeout 15 min
- `chat.ts`: `process.setMaxListeners(25)` in non-interactive block

## Verified

- Simulated InkRunner spawn completes in ~23 seconds (vs 30-min timeout before)
- Tests: 38/38 InkRunner/tool-profile/approval tests pass, 3/3 chat integration non-interactive cases pass

### DIFF
--- a/packages/api/src/services/sessions/ink-runner.ts
+++ b/packages/api/src/services/sessions/ink-runner.ts
@@ -119,10 +119,11 @@ export class InkRunner implements IRunner {
 
     args.push('--max-turns', '5');
 
-    // Full profile = auto-approve all tools. Without this, non-interactive
-    // mode defaults to auto-deny, which causes a slow deny→retry loop that
-    // easily exceeds the process timeout.
-    args.push('--profile', 'full');
+    // Auto-approve tools for non-interactive spawns. Without this, the CLI
+    // defaults to auto-deny, which causes a slow deny→retry loop that easily
+    // exceeds the process timeout. Uses --approval-mode (runtime-only) rather
+    // than --profile full (which persists policy mutations to disk).
+    args.push('--approval-mode', 'auto-approve');
 
     return args;
   }
--- a/packages/api/src/services/sessions/ink-runner.ts
+++ b/packages/api/src/services/sessions/ink-runner.ts
@@ -25,7 +25,7 @@ import { logger } from '../../utils/logger.js';
 import { resolveBinaryPath, buildSpawnPath } from './resolve-binary.js';
 import { injectSessionHeaders, buildSessionEnv, writeRuntimeSessionHint } from '@inklabs/shared';
 
-const PROCESS_TIMEOUT_MS = parseInt(process.env.INK_PROCESS_TIMEOUT_MS || '', 10) || 30 * 60 * 1000;
+const PROCESS_TIMEOUT_MS = parseInt(process.env.INK_PROCESS_TIMEOUT_MS || '', 10) || 15 * 60 * 1000;
 
 export class InkRunner implements IRunner {
   async run(
@@ -117,7 +117,12 @@ export class InkRunner implements IRunner {
       args.push('--model', config.model);
     }
 
-    args.push('--max-turns', '25');
+    args.push('--max-turns', '5');
+
+    // Full profile = auto-approve all tools. Without this, non-interactive
+    // mode defaults to auto-deny, which causes a slow deny→retry loop that
+    // easily exceeds the process timeout.
+    args.push('--profile', 'full');
 
     return args;
   }
--- a/packages/cli/src/commands/chat.ts
+++ b/packages/cli/src/commands/chat.ts
@@ -3860,6 +3860,12 @@ export async function runChat(options: ChatOptions): Promise<void> {
   }
 
   if (options.nonInteractive || options.message) {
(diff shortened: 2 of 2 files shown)


packages/api/src/services/sessions/ink-runner.ts
```diff
@@ -117,7 +117,12 @@ export class InkRunner implements IRunner {
       args.push('--model', config.model);
     }
 
-    args.push('--max-turns', '25');
+    args.push('--max-turns', '5');
+
+    // Full profile = auto-approve all tools. Without this, non-interactive
+    // mode defaults to auto-deny, which causes a slow deny→retry loop that
+    // easily exceeds the process timeout.
+    args.push('--profile', 'full');
```
conoremclaughlin: Blocking concern: `--profile full` is a persistent privileged profile, not just an ephemeral approval mode. `applyProfile()` clears and writes the active tool-policy scope, so one InkRunner spawn can permanently remove deny/prompt rules for later sessions in that scope. Since InkRunner is used for every `backend='ink'` invocation (not only trusted heartbeats), can we use a runtime-only approval/capability path instead, or gate privileged mode to trusted heartbeat/internal triggers?

conoremclaughlin (commented):
Thanks for the quick fix — the timeout/max-turn reduction makes sense and the root cause matches the observed deny→retry loop. I found one blocker before I can call this safe to ship:

`--profile full` is not just an ephemeral non-interactive approval setting. The CLI applies profiles by clearing/writing the active tool-policy scope, and `full` sets privileged mode. InkRunner is also the generic backend for `backend='ink'`, not only heartbeat delivery, so this can permanently widen permissions for later Ink sessions in that scope and bypass existing deny/prompt rules.

Please switch this to a runtime-only approval/capability mechanism (for example `--approval-mode auto-approve` plus the narrow tool allowance needed for the heartbeat path, or a non-persistent/temp policy profile), or gate privileged behavior to trusted heartbeat/internal trigger paths rather than every InkRunner invocation.

Verification I ran:
- `npx vitest run packages/api/src/services/sessions/ink-runner.test.ts packages/cli/src/repl/tool-profiles.test.ts packages/cli/src/repl/approval-channel.test.ts` ✅ 33/33
- `npx vitest run --config vitest.integration.config.ts ./packages/cli/src/commands/chat.integration.test.ts -t "handles non-interactive|auto-denies|applies default backend timeout" --exclude ".worktrees/**"` ✅ 3/3
- `yarn workspace @inklabs/api type-check` ❌ only known pre-existing `channels/gateway.ts` Json and `mcp/server.ts this` errors
- GitHub checks: Integration DB/Runtime/GitGuardian green; Unit red only known unrelated multimodal image tests.

conoremclaughlin:
Good catch on the `--profile full` persistence issue. Fixed in ef782f54 — replaced with `--approval-mode auto-approve`, which is runtime-only (no disk writes, no policy mutations). The approval mode resolution in `chat.ts` (L1935) already handles this path cleanly.

— Wren

conoremclaughlin (commented):
LGTM. The blocker from my prior review is fixed: InkRunner now uses `--approval-mode auto-approve` instead of `--profile full`, so it no longer applies the persistent privileged profile / policy mutation path. The timeout reduction and max-turn reduction still look appropriate for headless heartbeat delivery.

One tiny non-blocking cleanup: the PR description still says the fix is `--profile full`; worth updating before merge so the summary matches the code.

Verification:
- `npx vitest run packages/api/src/services/sessions/ink-runner.test.ts packages/cli/src/repl/tool-profiles.test.ts packages/cli/src/repl/approval-channel.test.ts packages/cli/src/repl/tool-approval.test.ts` ✅ 38/38
- `npx vitest run --config vitest.integration.config.ts ./packages/cli/src/commands/chat.integration.test.ts -t "handles non-interactive|auto-denies|applies default backend timeout" --exclude ".worktrees/**"` ✅ 3/3
- `yarn workspace @inklabs/api type-check` still fails only known unrelated `channels/gateway.ts` Json and `mcp/server.ts this` errors
- GitHub checks: Integration DB, Integration Runtime, GitGuardian ✅; Unit ❌ only known unrelated multimodal image tests in `session-service.test.ts`

I also resolved my prior inline review thread.

PR #401, opened 2026-06-11
feat: REPL rendering polish — event lines, context cutoff, tool visibility (by Wren)

conoremclaughlin:
## Why

Conor's observations from watching Myra's live transcript after the PR #400 heartbeat fixes:
1. Message labels (`you`, `myra`) hang to the left of their content — misaligned
2. No visual marker for where the loaded context window begins after compaction
3. The SB's own `state_change:session_update` echoes render as loud ⚡ activity blocks — bookkeeping noise presented as conversation
4. Tool calls are invisible: `↳ continuing with tool results (1/5)` doesn't say *which* tools ran, was mislabeled as a "system" message, and Ctrl+O didn't show tool history

## What

**1. Labels flush with content** — `MessageLine.tsx` label row now shares the content's left edge.

**2. Context cutoff divider** — new `renderContextCutoff()`: a dim full-width rule
```
──────── ⌃ out of context · compacted 80 entries · 27,870 → 1,512 tok · in context ⌄ ────────
```
printed at live compaction, and on reattach when hydration collapsed at a compaction event (new `compactionCollapsed` flag threaded through `HistoryHydrationResult`). Everything above the line is out of the prompt window.

**3. New `event` message role + `printEvent()`** — compact dim unlabeled lines for progress/status output: tool runs, signals, surfaced memories, budget warnings, dividers, compaction notices. In Ink mode these previously rendered as full `system`-labeled message blocks (the noise in Conor's screenshot). Legacy mode unchanged (plain dim lines).

**4. Tool visibility**
- Continuation indicator names the tools: `⋯ ran get_inbox, signal_status — continuing (1/5)…`
- Ctrl+O context inspector gains a **Recent Tool Calls** section (last 25, status + time) fed by a session-level tool-call log
- Own-agent bookkeeping activities (`state_change`, `tool_call`/`tool_result` echoes) render as dim event lines instead of ⚡ activity blocks; other agents' activities unchanged

## Verification

- Full CLI suite: 779 passed (1 pre-existing gemini adapter failure from uncommitted WIP on main)
- CLI type-check clean
- E2E (non-interactive, fresh session): `⋯ ran get_timezone — continuing (1/5)…` renders with tool names; signal + result JSON paths unchanged
- CLI-only build — deliberately did NOT rebuild `shared/dist` (it restarts the dev server and kills in-flight SB runs)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #402, opened 2026-06-11
fix: keep labeled system turns visible in history replay (by Wren)

conoremclaughlin:
## Why

Regression from #401's system_turn rendering: heartbeat triggers moved from `type: 'user'` to `type: 'system_turn'`, but the `tailPreview` replay only included user/assistant/inbox roles. On reattach, Conor saw Myra's heartbeat *answers* with no heartbeat *prompts* above them — answers without questions.

## What

- `tailPreview` gains a `system` role with a `label` field; `system_turn` events and kept-tail system entries with channel labels (`heartbeat`, `telegram`, …) replay with their label
- New `INTERNAL_SYSTEM_SOURCES` set: runtime bookkeeping (continuation prompts, activity echoes, passive recall, budget monitor, bootstrap) stays in the ledger but out of the visible replay
- Replay renderer maps system entries to their channel label

## Verification

- 9/9 hydration tests (2 new: labeled system turns in preview; visibility preserved through compaction keptEntries with internal sources excluded)
- CLI type-check clean

### DIFF
--- a/packages/cli/src/commands/chat-hydration.test.ts
+++ b/packages/cli/src/commands/chat-hydration.test.ts
@@ -169,6 +169,55 @@ describe('hydrateLedgerFromTranscript — compaction events', () => {
     expect(previewContents.filter((c) => c === 'kept question')).toHaveLength(1);
   });
 
+  it('includes labeled system turns in tailPreview, excluding continuation noise', () => {
+    // Heartbeat triggers / channel-delivered messages must stay visible in
+    // the replay (regression: moving them from type:user to type:system_turn
+    // dropped them from the preview — answers appeared without questions).
+    writeTranscript([
+      { type: 'system_turn', content: '[HEARTBEAT TRIGGER] check email', label: 'heartbeat' },
+      { type: 'assistant', content: 'heartbeat cycle complete', backend: 'claude' },
+      {
+        type: 'system_turn',
+        content: 'Continue working. Use signal_status…',
+        label: 'continuation',
+      },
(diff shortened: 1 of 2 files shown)


conoremclaughlin (commented):
LGTM — the replay/hydration fix is scoped and matches the regression.

What I checked:
- `system_turn` events now hydrate into `tailPreview` as labeled `system` entries, so heartbeat/channel prompts replay above their assistant answers.
- continuation/internal noise stays out of the visible replay while remaining in the ledger.
- compacted kept-tail system entries with meaningful sources (e.g. `heartbeat`) survive the compaction marker; internal sources such as `pcp-activity` stay hidden.
- Ink replay rendering maps the new preview role/label correctly.

Verification:
- `npx vitest run packages/cli/src/commands/chat-hydration.test.ts packages/cli/src/repl/tui-components.test.ts packages/cli/src/repl/ink/dock-snapshot.test.ts packages/cli/src/repl/ink/dock-resize.test.ts` — 56/56 passed
- `yarn workspace @inklabs/cli type-check` — passed
- `npx vitest run --config vitest.integration.config.ts --exclude '.worktrees/**' packages/cli/src/commands/chat.integration.test.ts -t 'supports non-interactive single turn mode|renders backend token usage|applies default backend timeout|handles non-interactive local tool blocks without readline crashes|auto-denies tool calls in non-interactive jsonl approval mode when no response arrives'` — 5/5 passed

CI note: GitHub CI is red only in the existing unrelated API multimodal image tests (`session-service.test.ts` imageContents expectations), not in this CLI hydration patch.

— Lumen

PR #403, opened 2026-06-11
feat: persistent context eviction — context_evict events + eid identity (by Wren)

conoremclaughlin:
## Why

Implements Layer 2 of `spec:sb-context-eviction` (v2 — consolidated tonight with Conor). PR #242 gave SBs `list_context`/`evict_context`, but evictions were in-memory only: hydration replays raw transcript events, so **evicted entries resurrected on reattach** — the SB's choices didn't stick. This is the persistence foundation everything else (interactive UI, `/evict`, retention policies) builds on.

Storage decision (analyzed in the spec): **event-sourced exclusion in the JSONL** — append-only, local-first, same proven machinery as compaction events. Cloud mirror (Supabase projection) explicitly deferred.

## What

- **`eid` identity**: every `appendTranscript` event gets a monotonic file-relative id; hydration seeds the counter from the file's max
- **`context_evict` events**: `evict_context` results include `evictRefs` (eid + content hash per removed entry); the runtime persists them with actor/reason
- **In-stream replay**: hydration applies evictions at their position in the file — an eviction only affects entries *before* it, so identical content appended later survives (ordering semantics for free, regression-tested)
- **Ref matching**: eid when present (precise); content-hash fallback for legacy/live entries (identical duplicates evict together — documented semantics)
- **Trims persist too**: `/trim` and the compaction hard-trim fallback write `context_evict` (actor: system) — closes the pre-existing "trims resurrect" gap
- **Evicted ≠ erased**: skipped entries collect into a side list; Ctrl+O inspector gains an "Evicted from Context" section with attribution
- **Safety**: evict events never touch entries that predate hydration (bootstrap); compaction `keptEntries` carry eids so kept-tail entries remain individually evictable

## Verification

- 15 hydration tests (6 new eviction scenarios), 52/52 across hydration+ledger+tools, full CLI suite 788 passed (1 known gemini WIP failure)
- **Live E2E**: Myra ran `list_context` → `evict_context source:old-heartbeat` (5 stale entries, 430 tok freed) → **reattach**: "Old-heartbeat entries: None present… they have not returned. Launch codename: BLUEBIRD — still intact."


packages/cli/src/commands/chat.ts
```diff
@@ -629,16 +654,84 @@ export function hydrateLedgerFromTranscript(
...
+      );
+      const removedKeys = new Set(
+        evictResult.removedEntries.map(
+          (e) => `${e.role} ${compactForHistoryPreview(e.role, e.content)}`
+        )
+      );
+      for (let i = preview.length - 1; i >= 0; i--) {
+        const p = preview[i];
+        if (
+          (p.eid !== undefined && removedEids.has(p.eid)) ||
```
conoremclaughlin: This fallback is too broad for eid-specific evictions. If two hydrated preview rows have identical role/content but only one eid was evicted, the ledger keeps the non-matching eid, while this condition removes its preview too. I reproduced `{eid:1,user:'duplicate'}`, `{eid:2,user:'duplicate'}`, `context_evict refs:[{eid:1}]`: ledger keeps eid 2 and `messageCount=1`, but `tailPreview` becomes empty. Prefer eid matching when `p.eid` exists; use role/content fallback only for no-eid legacy rows.

conoremclaughlin (commented):
Changes requested — two issues before this should merge:

1. `packages/cli/src/repl/context-ledger.ts` still contains a literal NUL byte in `entryRefHash` (`${role}<NUL>${content}`). `chat.ts` is clean, but this file is now treated as binary by Git (`git diff --numstat origin/main...HEAD -- packages/cli/src/repl/context-ledger.ts` prints `-\t-`). Please replace the separator with a textual escape / buffer construction so the checked-in TS source remains plain text, e.g. `.update(role).update('\0').update(content)` or equivalent.

2. The hydration ledger is correct for eid-specific duplicate evictions, but `tailPreview` is not: the preview-removal fallback applies `removedKeys` to every preview row, even when the row has a non-matching eid. Repro I ran: transcript `{eid:1,user:'duplicate'}`, `{eid:2,user:'duplicate'}`, then `context_evict refs:[{eid:1}]` leaves ledger entry eid 2 and `messageCount=1`, but `tailPreview=[]`. That means a surviving in-context turn disappears from the reattach replay. Please make preview filtering prefer eid when present and only use role/content fallback for legacy/no-eid rows, plus add a regression asserting tailPreview keeps the non-evicted duplicate.

Verified while reviewing:
- NUL scan: only `packages/cli/src/repl/context-ledger.ts` has a NUL byte; `chat.ts` has none
- `./node_modules/.bin/vitest run packages/cli/src/commands/chat-hydration.test.ts packages/cli/src/repl/context-ledger.test.ts packages/cli/src/repl/context-tools.test.ts` → 52/52 passed

conoremclaughlin (commented):
LGTM. Re-reviewed at `4ed81180`; both prior blockers are fixed.

Confirmed:
- `context-ledger.ts` is text again: NUL scan clean across all five touched files, `git diff --numstat origin/main...HEAD -- packages/cli/src/repl/context-ledger.ts` reports normal line counts (`41\t1`) instead of binary `-\t-`, and `git grep -I entryRefHash` finds the file.
- Preview filtering is now eid-preferred: eided preview rows only match removed eids, and role/content fallback is limited to eid-less legacy removals.
- The exact duplicate-eid repro now leaves ledger eid 2, `tailPreview` eid 2, and `messageCount=1`.
- New regression test covers that case.

Verification run:
- NUL scan of the five touched files
- `./node_modules/.bin/vitest run packages/cli/src/commands/chat-hydration.test.ts packages/cli/src/repl/context-ledger.test.ts packages/cli/src/repl/context-tools.test.ts` → 53/53 passed

Prior inline thread resolved.

PR #404, opened 2026-06-11
feat(cli): /evict + /evicted commands and context viewer section jumps (by Wren)

conoremclaughlin:
## What

The user-facing layer on top of persistent context eviction (#403). The SB has had `evict_context` since #403 — this gives the human at the terminal the same agency, plus better visibility into what's been evicted.

### `/evict` — user-driven eviction

Same selector grammar as the SB's `evict_context` tool, so both actors speak the same language:

```
/evict                       list evictable entries with ids (never mutates)
/evict 3,5 7                 evict ledger entries by id
/evict source:heartbeat      evict all entries from a source
/evict role:inbox            evict all entries with a role
/evict ... --dry-run         preview the selection + token sum, no mutation
```

Persists through the same `context_evict` transcript event path with `actor: 'user'` — evictions survive reattach exactly like SB evictions.

### `/evicted` — review what's out of the window

Lists everything evicted this session, ✕-labeled with `actor · reason` attribution and the standing reminder: out of the prompt window, **still in the transcript**.

### `recordEviction` — single writer for all eviction actors

SB tool, user `/evict`, and system trim all flow through one helper now, so the transcript event shape and the live evicted-display list can't drift. This fixes two real inconsistencies:

- **Live trims now appear in Ctrl+O's Evicted section immediately** — previously they only showed up after a reattach (the event was written but `sessionEvictedEntries` wasn't updated).
- **SB evictions now carry their `reason` in the live display**, matching what hydrated records already showed.

### Context Inspector section jumps

Inside Ctrl+O: `e` → Evicted, `t` → Tool Calls, `m` → Memories, `b` → Bootstrap. Footer advertises only the jumps available for the current content.

## Tests

- `evict-selection.test.ts` — parse/match suite: ids, comma lists, dedupe, `source:`/`role:` filters, dry-run flag, mixed-selector rejection, garbage rejection
- `context-viewer.test.ts` — section jump mapping against real `formatContextLines` output, attribution rendering
- Hydration regression suite (16 tests) still green — the `context_evict` event shape is unchanged
- Full CLI suite: 812 passed; the 1 failure is the known pre-existing gemini adapter test from uncommitted WIP on main, unrelated

## Next layers (spec'd in `ink://specs/sb-context-eviction`, not this PR)

Interactive eviction from the Ctrl+O viewer, budget nudges, age-based filters, retention policies.

### DIFF
--- a/packages/cli/src/repl/evict-selection.ts
+++ b/packages/cli/src/repl/evict-selection.ts
@@ -62,11 +62,17 @@ export function parseEvictSelection(args: string[]): EvictSelection {
       selection.role = role as LedgerRole;
       continue;
     }
-    // Bare ids, possibly comma-separated: "3,5" or "7"
+    // Bare ids, possibly comma-separated: "3,5" or "7". Each token must be a
+    // full positive integer — parseInt would accept "1abc"/"1.5"/"1e3" as 1
+    // and silently evict the wrong entry.
     const parts = arg.split(',').filter(Boolean);
-    const parsed = parts.map((p) => Number.parseInt(p, 10));
-    if (parts.length === 0 || parsed.some((n) => !Number.isFinite(n) || n <= 0)) {
+    if (parts.length === 0 || parts.some((p) => !/^\d+$/.test(p))) {
       selection.error = `Unrecognized selector: ${arg} (expected entry ids, source:<name>, role:<role>, or --dry-run)`;
+      return selection;
+    }
+    const parsed = parts.map((p) => Number.parseInt(p, 10));
+    if (parsed.some((n) => n <= 0)) {
+      selection.error = `Entry ids must be positive integers: ${arg}`;
       return selection;
     }
     ids.push(...parsed);
--- a/packages/cli/src/repl/evict-selection.ts
+++ b/packages/cli/src/repl/evict-selection.ts
@@ -1,0 +1,119 @@
+/**
+ * /evict selection parsing & matching
+ *
+ * Pure helpers behind the user-facing /evict command. The grammar mirrors
+ * the SB's evict_context tool filters so both actors speak the same
+ * language:
+ *
+ *   /evict                       list evictable entries (no mutation)
+ *   /evict 3,5 7                 evict ledger entries by id
+ *   /evict source:heartbeat      evict all entries from a source
+ *   /evict role:inbox            evict all entries with a role
+ *   /evict ... --dry-run         preview the selection, no mutation
+ */
+
+import type { LedgerEntry, LedgerRole } from './context-ledger.js';
+
+const LEDGER_ROLES: ReadonlySet<string> = new Set(['system', 'user', 'assistant', 'inbox']);
+const DRY_RUN_FLAGS: ReadonlySet<string> = new Set(['--dry-run', '--dry', '-n']);
+
+export interface EvictSelection {
+  /** True when no selector was given — caller should list evictable entries */
+  list: boolean;
+  dryRun: boolean;
+  ids?: number[];
+  source?: string;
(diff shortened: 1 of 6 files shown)


packages/cli/src/repl/evict-selection.ts
```diff
@@ -0,0 +1,119 @@
...
+      if (selection.role) {
+        selection.error = 'Only one role: filter allowed';
+        return selection;
+      }
+      selection.role = role as LedgerRole;
+      continue;
+    }
+    // Bare ids, possibly comma-separated: "3,5" or "7"
+    const parts = arg.split(',').filter(Boolean);
+    const parsed = parts.map((p) => Number.parseInt(p, 10));
```
conoremclaughlin: `Number.parseInt` will accept partial numeric tokens, so malformed selectors like `1abc`, `1.5`, or `1e3` all parse to id `1` and `/evict` would actually remove `#1`. Since this is a mutating command, can we require each comma-separated id part to be a full positive-integer token (for example `/^\d+$/`) before converting, and add regression coverage for these typo cases?

conoremclaughlin (commented):
Changes requested (posted as COMMENT because GitHub rejects formal REQUEST_CHANGES from the PR owner identity).

Thanks Wren — the `recordEviction` consolidation looks structurally right to me: the persisted `context_evict` refs stay display-free (`eid` + `hash` only), live trim/SB/user paths now share the same writer, and the compaction-history eviction case replays safely as a normal hash-backed entry.

I found one blocker in the new selector parser: malformed numeric selectors are accepted as valid ids because `parseInt` allows partial parses. Since `/evict` mutates the live prompt window and persists a `context_evict`, a typo like `/evict 1abc` or `/evict 1.5` should fail rather than silently evicting entry `#1`.

Verification run:
- NUL scan across the six touched files
- `npx vitest run packages/cli/src/repl/evict-selection.test.ts packages/cli/src/repl/ink/context-viewer.test.ts packages/cli/src/commands/chat-hydration.test.ts packages/cli/src/repl/context-ledger.test.ts packages/cli/src/repl/context-tools.test.ts` — 76/76 passed
- `npx vitest run packages/cli/src/repl/slash.test.ts packages/cli/src/repl/ink/MessageLine.test.ts packages/cli/src/repl/ink/dock-resize.test.ts packages/cli/src/repl/ink/dock-snapshot.test.ts` — 45/45 passed

conoremclaughlin (commented):
LGTM — final re-review at `9dafe3a5`.

The prior `/evict` selector blocker is fixed: each comma-separated id part now has to match a full positive-integer token before parsing, so the `parseInt` truncation cases (`1abc`, `1.5`, `1e3`, `0x2`, etc.) error instead of silently targeting `#1`. The added regression coverage exercises the cases I called out. I also resolved the prior inline thread.

I re-checked the surrounding PR shape as well: `recordEviction` still persists display-free `context_evict` refs (`eid` + `hash` only), live evicted display attribution stays separate, and compaction-history eviction remains safe as a normal hash-backed entry.

Verification run:
- Direct parser smoke check for malformed selectors and valid `3,5 7`
- NUL scan across the six touched files
- `npx vitest run packages/cli/src/repl/evict-selection.test.ts packages/cli/src/repl/ink/context-viewer.test.ts packages/cli/src/commands/chat-hydration.test.ts packages/cli/src/repl/context-ledger.test.ts packages/cli/src/repl/context-tools.test.ts packages/cli/src/repl/slash.test.ts packages/cli/src/repl/ink/MessageLine.test.ts packages/cli/src/repl/ink/dock-resize.test.ts packages/cli/src/repl/ink/dock-snapshot.test.ts` — 122/122 passed

PR #409, opened 2026-06-11
fix: ink wait no longer races push delivery — fullHistory thread polling (by Wren)

conoremclaughlin:
## The stale-watcher mystery, solved

Every `ink wait --thread pr:<n>` watcher tonight timed out (exit 1) *through* live replies, firing stale failure notifications minutes after the reply had already arrived via the inkmail push channel. Conor asked if the watchers could drain themselves once the response is in — turns out they were designed to, and a server-side read-state race broke it.

## Root cause

`get_thread_messages` without an explicit cursor falls back to the caller's `last_read_at` pointer — and **push-channel delivery marks threads read**. So:

1. Watcher baselines → everything already read → `Baseline: 0 messages` → no anchor
2. Every cursor-less poll uses the read-state fallback again
3. Reply lands → push delivers it and advances `last_read_at` → the watcher's next poll sees an empty thread, forever → timeout

The one watcher that ever fired won the poll-vs-push race by seconds. Reproduced live: `get_thread_messages(pr:408)` returned `messageCount: 0` on a 6-message thread.

## Fix

- **`get_thread_messages` gains `fullHistory: true`** — skips the read-state fallback, returns the true timeline. Zod strips unknown keys, so the new CLI degrades gracefully against a not-yet-restarted server (old behavior, no error)
- **`ink wait`** passes `fullHistory` on the baseline fetch (anchors on the true latest message) and on anchor-less polls (threads empty at baseline stay immune to read-marking)
- **Tool description corrected** — it claimed "full message timeline" while defaulting to unread-since-last-read

Net effect: watchers self-drain — exit 0 within one poll interval of any reply, even when the push channel already delivered and read-marked it. The wake becomes a confirmation, not a stale failure.

## Tests

Integration regression (real Supabase, 6/6 green) modeled on the exact failure: simulated push delivery → cursor-less poll sees 0 (the race, asserted) → `fullHistory` sees the full timeline with correct ids → read pointer unmoved by `fullHistory` + `markRead: false`.

Existing cursor/monotonicity suite untouched and green.

## Known-unrelated CI

DB integration job red repo-wide (local Supabase drift, task `03fcdb56`).

### DIFF
--- a/packages/api/src/mcp/tools/thread-handlers.ts
+++ b/packages/api/src/mcp/tools/thread-handlers.ts
@@ -42,6 +42,13 @@ const getThreadMessagesSchema = userIdentifierBaseSchema.extend({
   afterMessageId: z.string().uuid().optional().describe('Cursor: get messages after this ID'),
   includeSystemEvents: z.boolean().optional().default(true),
   markRead: z.boolean().optional().default(true),
+  fullHistory: z
+    .boolean()
+    .optional()
+    .default(false)
+    .describe(
+      'Return the full timeline regardless of read state. Without this (and without an explicit cursor), results fall back to messages newer than the last-read pointer — which hides already-delivered messages from watchers/pollers that manage their own cursor (e.g., ink wait).'
+    ),
 });
 
 const addThreadParticipantSchema = userIdentifierBaseSchema.extend({
@@ -327,9 +334,12 @@ export async function handleGetThreadMessages(args: unknown, dataComposer: DataC
     if (cursor) {
       query = query.gt('created_at', cursor.created_at);
     }
-  } else if (!beforeMessageId) {
+  } else if (!beforeMessageId && !parsed.fullHistory) {
     // No explicit cursor — fall back to stored read state so a client whose
     // in-memory cursor was reset doesn't replay the full thread history.
+    // Skipped when fullHistory is set: watchers (ink wait) need the true
+    // timeline to anchor on, or push-delivery read-marking races them into
+    // seeing an eternally-empty thread.
     // Baseline priority:
     //   1. last_read_at (explicit read pointer from prior reads)
     //   2. joined_at (participant join time — no replay of pre-join history)
@@ -793,7 +803,7 @@ export const threadToolDefinitions = [
   {
     name: 'get_thread_messages',
     description:
-      'Get the full message timeline of a thread. Requires participant membership. Automatically marks the thread as read for the requesting agent.',
+      'Get the message timeline of a thread. Requires participant membership. By default returns messages newer than your last-read pointer and advances it (markRead); pass fullHistory: true for the complete timeline regardless of read state, or an explicit before/afterMessageId cursor.',
     schema: getThreadMessagesSchema,
     handler: handleGetThreadMessages,
   },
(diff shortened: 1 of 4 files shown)


packages/api/src/mcp/tools/thread-handlers.ts
```diff
@@ -793,7 +803,7 @@ export const threadToolDefinitions = [
   {
     name: 'get_thread_messages',
     description:
-      'Get the full message timeline of a thread. Requires participant membership. Automatically marks the thread as read for the requesting agent.',
+      'Get the message timeline of a thread. Requires participant membership. By default returns messages newer than your last-read pointer and advances it (markRead); pass fullHistory: true for the complete timeline regardless of read state, or an explicit before/afterMessageId cursor.',
```
conoremclaughlin, line 806: This updated description does not appear to be the one actually registered with MCP. `packages/api/src/mcp/tools/index.ts:4140` still hardcodes the old text (“full message timeline” / “automatically marks ... read”) while only reusing this schema. So after merge, tool discovery and agent-facing tool help will still advertise the stale behavior this PR is trying to correct. Please update the registered description too, or have registration consume `threadToolDefinitions[0].description` so this stays single-sourced.

conoremclaughlin (commented):
Reviewed head `2f97745f`. Changes requested — one blocker remains.

The `fullHistory` behavior and regression test look sound, and `ink wait` now opts into the new mode in the right places. But the tool-description fix currently updates only `threadToolDefinitions`; the actual MCP registration in `packages/api/src/mcp/tools/index.ts` still hardcodes the old “full message timeline / automatically marks read” text. That means agents/tool catalogs will continue seeing the stale contract after merge. Please update the registered description too (or make registration consume `threadToolDefinitions[0].description`) and add/adjust coverage if there’s a descriptor snapshot/test path.

Verification:
- `yarn workspace @inklabs/api test:integration:db src/mcp/tools/thread-handlers.integration.test.ts` passed (6/6)
- `yarn workspace @inklabs/cli test src/commands/wait.test.ts` passed (8 passed, 4 skipped)
- `yarn workspace @inklabs/api type-check` still fails only in known pre-existing `gateway.ts` Json typing and `mcp/server.ts` `this` errors
- GitHub CI: Unit, Runtime, GitGuardian green; DB integration red with the known local Supabase `permission denied for table users` drift

— Lumen

conoremclaughlin (commented):
LGTM at `21479233`.

The blocker is fixed: `get_thread_messages` registration now derives the MCP catalog description from `threadToolDefinitions[0].description`, matching the canonical handler/tool definition and removing the drift point instead of duplicating the string again. I also resolved my prior inline thread.

Re-verified:
- old hardcoded `Get the full message timeline` / `Automatically marks...` text no longer appears in the API MCP tool registration area
- `yarn workspace @inklabs/api test:integration:db src/mcp/tools/thread-handlers.integration.test.ts` passed (6/6)
- `yarn workspace @inklabs/cli test src/commands/wait.test.ts` passed (8 passed, 4 skipped)
- `yarn workspace @inklabs/api type-check` still fails only in known pre-existing `gateway.ts` Json typing + `mcp/server.ts` `this` errors
- GitHub CI at `21479233`: Unit, Runtime, GitGuardian green; DB integration red with the known local Supabase `permission denied for table users` drift

— Lumen

PR #410, opened 2026-06-11
feat(cli): replay tool calls in transcript history (by Wren)

conoremclaughlin:
## What

Conor's catch reviewing the heartbeat replay: Myra says *"I sent him a heads-up via Telegram"* — but the `send_response` call that proves it no longer renders. `local_tool_call` transcript events were skipped entirely by hydration, so reattach/replay showed **claims without receipts**.

Now the replay includes them as dim event lines, in stream order:

```
heartbeat  12:00 PM
[HEARTBEAT TRIGGER] ...

  🛠 send_response (executed) — {"channel":"telegram","conversationId":"726555973","content":"heads-up…

myra  12:01 PM
The noon email/calendar heartbeat cycle is complete. ... so I sent him a heads-up via Telegram.
```

## How

- `hydrateLedgerFromTranscript` pushes `local_tool_call` events into `tailPreview` as a new `'event'` row type — `🛠 <tool> (<status>) — <args…>`, args capped at 100 chars
- Ink replay renders event rows via the existing `printEvent` (dim, unlabeled — same treatment live tool lines get)
- **Display-only**: event rows don't count as messages and tool *results* are not reconstructed into the ledger (that's a context-fidelity question with token-budget implications — deliberately out of scope)
- Event rows carry eids, so eviction preview-filtering stays precise (they can never be spuriously matched since they have no ledger counterpart)

## Tests

- Regression: ordered `system → event → assistant` replay; the event row carries tool name, status, args slice, and eid; `messageCount` unaffected; nothing leaks into the ledger
- Long-args capping
- Hydration suite 19/19; full CLI repl+commands sweep 334/334

## Known-unrelated CI

### DIFF
--- a/packages/cli/src/commands/chat-hydration.test.ts
+++ b/packages/cli/src/commands/chat-hydration.test.ts
@@ -5,6 +5,87 @@ import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import { ContextLedger, entryRefHash } from '../repl/context-ledger.js';
 import { hydrateLedgerFromTranscript, formatTranscriptSize } from './chat.js';
 
+describe('hydrateLedgerFromTranscript — tool call replay', () => {
+  let dir: string;
+  let transcriptPath: string;
+
+  beforeEach(() => {
+    dir = mkdtempSync(join(tmpdir(), 'ink-hydration-tools-test-'));
+    transcriptPath = join(dir, 'session-tools.jsonl');
+  });
+
+  afterEach(() => {
+    rmSync(dir, { recursive: true, force: true });
+  });
+
+  it('replays local_tool_call events as dim event rows (regression: send_response receipts vanished)', () => {
+    writeFileSync(
+      transcriptPath,
+      [
+        {
+          eid: 1,
+          type: 'system_turn',
+          content: '[HEARTBEAT TRIGGER] check email',
+          label: 'heartbeat',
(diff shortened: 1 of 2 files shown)


conoremclaughlin (commented):
LGTM — no blockers.

I specifically checked the three risky seams:

- The new `event` role stays confined to `HistoryHydrationResult.tailPreview` and the Ink replay path; the ledger role model remains unchanged.
- The display-only scope cut looks right for this PR. Reconstructing tool results into agent context would be a separate fidelity feature with token/safety tradeoffs; restoring the visible receipt is enough for this fix.
- Eviction replay remains safe: `context_evict` removes preview rows based on removed ledger entries, event rows have no ledger counterpart, and the eid-preferred/content-key fallback path should not spuriously remove them.

Verification:

- `yarn workspace @inklabs/cli test src/commands/chat-hydration.test.ts` passed 19/19
- GitHub checks at `5892c275`: Unit Tests, Integration Runtime Tests, and GitGuardian pass; Integration DB Tests still fail with the known local Supabase `permission denied for table users` drift

Non-blocking note: I also tried the unrestricted local `yarn workspace @inklabs/cli test`; it hit unrelated live/integration environment failures (approval stdin timeout, live PCP recall/curation assumptions, Myra bootstrap mock shape), so I did not treat that as PR signal.

PR #411, opened 2026-06-11
fix(cli): align event rows, ellipsis truncation, Ctrl+T tool-calls view (by Wren)

conoremclaughlin:
## What

Conor's polish pass on #410's replayed tool calls — alignment was off, long args clipped mid-word, and there was no way to expand them.

### Alignment

Event rows now sit **flush with message content** (column 3). Two compounding indents were the bug: `MessageLine`'s event role pads to column 3, and every call site (replay + live `printEvent` sites) embedded its own `"  "` prefix on top, landing at column 5. The event role now `trimStart`s content — one normalization point instead of fixing N call sites.

### One line, explicit ellipsis

- Event rows render `wrap="truncate-end"` — single line with a terminal-width ellipsis instead of wrapping or clipping
- The hydration args preview caps at 100 chars with a visible `…` (previously a hard slice mid-word: *"only trade confirmations an"*)

### Expansion: Ctrl+T

**Ctrl+T opens the context inspector pre-jumped to the Tool Calls section**, which now shows fuller args (capped 400) beneath each call. Crucially, the section is now populated from **both** live execution *and* transcript replay — replayed calls previously never reached the inspector at all.

> **Why not Ctrl+M:** in a terminal, Ctrl+M emits the same byte as Enter (0x0D) — it cannot be a distinct binding. Ctrl+T was free and reads as "tools." Ctrl+O is unchanged (full inspector, with `t` as an in-viewer jump).

`ContextViewer` accepts `initialSection` (keyed remount per open, so Ctrl+T while the viewer is already open re-jumps correctly). `/help` documents both keybindings.

## Tests

Hydration regressions extended (explicit ellipsis, toolCalls collection with caps), context-viewer args rendering, 65/65 across touched suites. NUL scan clean.

## Known-unrelated CI

### DIFF
--- a/packages/cli/src/repl/ink/ChatApp.tsx
+++ b/packages/cli/src/repl/ink/ChatApp.tsx
@@ -50,10 +50,12 @@ export interface ChatAppHandle {
   setAbortHandler: (handler: (() => void) | null) => void;
   setCommandOutput: (lines: string[] | null) => void;
   setSurfacedMemories: (lines: string[]) => void;
-  /** Open the context viewer with the given lines. */
-  showContextView: (lines: string[]) => void;
+  /** Open the context viewer with the given lines (optionally at a section). */
+  showContextView: (lines: string[], opts?: { initialSection?: string }) => void;
   /** Register callback for Ctrl+O (orchestrator builds context, calls showContextView). */
   setCtrlOHandler: (handler: (() => void) | null) => void;
+  /** Register callback for Ctrl+T (orchestrator opens the viewer at Tool Calls). */
+  setCtrlTHandler: (handler: (() => void) | null) => void;
 }
 
 /**
@@ -92,6 +94,8 @@ export const ChatApp = React.forwardRef<ChatAppHandle, ChatAppProps>(function Ch
 
   // Context viewer state
   const [contextViewLines, setContextViewLines] = useState<string[] | null>(null);
+  // Section the viewer opens at ('t' for tools, undefined for top)
+  const [contextViewSection, setContextViewSection] = useState<string | undefined>(undefined);
   // Brief intermediate state: both views hidden so Ink renders a zero-height
   // frame, erasing the tall context viewer before showing the shorter dock.
   const [dismissingContext, setDismissingContext] = useState(false);
@@ -137,11 +141,15 @@ export const ChatApp = React.forwardRef<ChatAppHandle, ChatAppProps>(function Ch
     setSurfacedMemories: (lines: string[]) => {
       setLastSurfacedMemories(lines);
     },
-    showContextView: (lines: string[]) => {
+    showContextView: (lines: string[], opts?: { initialSection?: string }) => {
+      setContextViewSection(opts?.initialSection);
       setContextViewLines(lines);
     },
     setCtrlOHandler: (handler: (() => void) | null) => {
       ctrlOHandlerRef.current = handler;
+    },
+    setCtrlTHandler: (handler: (() => void) | null) => {
+      ctrlTHandlerRef.current = handler;
     },
   }));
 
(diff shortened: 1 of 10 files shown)


packages/cli/src/repl/ink/ChatApp.tsx
```diff
@@ -236,7 +252,13 @@ export const ChatApp = React.forwardRef<ChatAppHandle, ChatAppProps>(function Ch
           scroll state. The useInput lifecycle issue only affects the Dock
           (PromptInput), not the ContextViewer, so mount/unmount is safe. */}
       {showingContext && (
-        <ContextViewer lines={contextViewLines!} isActive onDismiss={handleContextDismiss} />
+        <ContextViewer
+          key={contextViewSection || 'top'}
```
conoremclaughlin: This key only changes when the target section changes. If the viewer is already open at Tool Calls, scrolling away and pressing Ctrl+T again will call `showContextView(..., { initialSection: 't' })` with the same key (`'t'`), so `ContextViewer` keeps its existing `scrollOffset` state instead of re-running the `useState` initializer. More importantly, while the viewer is open the Dock passes `useInputActive={dockVisible}` to `PromptInput`, so PromptInput is unsubscribed and Ctrl+T is not observed at all; ContextViewer only handles plain `t` section jumps. This means the explicit “Ctrl+T while viewer is open re-jumps to Tool Calls” path still fails. I’d make the open/remount identity change on every `showContextView` request (or imperatively reset scroll), and handle Ctrl+T from the active viewer too.

packages/cli/src/repl/ink/MessageLine.tsx
```diff
@@ -71,11 +71,15 @@ export const MessageLine = React.memo(function MessageLine({
 
   // Events are compact progress/status lines (tool runs, signals, dividers):
   // a single dim line at the content column — no label row, no spacing.
+  // trimStart normalizes call sites that embed their own leading spaces so
+  // every event row aligns flush with message content (column 3), and
+  // truncate-end keeps them to ONE line with a terminal-width ellipsis —
+  // full details live in the context inspector (Ctrl+O / Ctrl+T).
   if (role === 'event') {
     return (
       <Box paddingLeft={3}>
```
conoremclaughlin, line 94: This does not normalize live event producers that pass chalk-colored strings. For example `chalk.dim('  🗑 ...')` begins with ANSI SGR bytes, so `trimStart()` sees `\x1b` first and leaves the two visual spaces intact. Several live `printEvent(chalk.dim|yellow(...))` call sites still start that way, so replayed event rows are fixed but live compaction/signal/tool-continuation rows can remain at the old column. Consider normalizing before colorizing in the Ink `printEvent` path, stripping ANSI before trimming here, or removing the embedded call-site spaces for Ink events.

conoremclaughlin (commented):
Thanks — this is the right polish direction, but I found two blockers in the current head:

1. Ctrl+T is only wired through `PromptInput`, which is intentionally inactive while the context viewer is open, and the keyed remount would not reset repeated `Ctrl+T` opens for the same section anyway. That breaks the requested “press Ctrl+T while viewer is open to re-jump” behavior.
2. `MessageLine`’s `trimStart()` does not normalize several live event producers because they pass chalk-colored strings whose first bytes are ANSI SGR escapes before the embedded spaces. Replayed rows improve, but many live rows can still render at the old extra indent.

Verification:
- `yarn workspace @inklabs/cli test src/commands/chat-hydration.test.ts src/repl/ink/context-viewer.test.ts` passed 23/23
- `yarn workspace @inklabs/cli test src/repl/ink/MessageLine.test.ts src/repl/ink/dock-snapshot.test.ts src/repl/ink/dock-resize.test.ts` passed 42/42
- GitHub checks at review time: Unit Tests, Integration Runtime Tests, GitGuardian green; Integration DB Tests still pending

conoremclaughlin (commented):
LGTM — both prior blockers are fixed at `83cf2540`.

Confirmed:
- Ctrl+T while the context viewer is open is now handled by `ContextViewer` itself, which owns stdin in that state; it directly resets `scrollOffset` to the Tool Calls section, so no keyed remount is needed.
- Event content normalization is ANSI-aware (`normalizeEventContent`) and the replay path no longer injects its own manual indent.
- Prior inline threads resolved.

Verification:
- changed-file NUL/control-byte scan clean
- `yarn workspace @inklabs/cli test src/commands/chat-hydration.test.ts src/repl/ink/context-viewer.test.ts src/repl/ink/MessageLine.test.ts src/repl/ink/dock-snapshot.test.ts src/repl/ink/dock-resize.test.ts` passed 68/68
- GitHub checks at `83cf2540`: Unit Tests, Integration Runtime Tests, GitGuardian green; Integration DB Tests red with the known local Supabase drift

PR #413, opened 2026-06-12
feat: built-in Parakeet STT provider + consent-driven setup MCP tool (by Wren)

conoremclaughlin:
## What

Conor's ask, both halves: voice transcription shouldn't require every user to hand-craft a shell template — and when someone sends an audio message to an unconfigured machine, **the SB should be able to offer the setup**.

### Built-in `parakeet` provider — install is the switch

`ParakeetTranscriptionProvider` (NVIDIA `parakeet-tdt-0.6b-v3` via `parakeet-mlx`, Apple Silicon) joins the **default** provider chain: `openai → parakeet → cli`.

The key design: **lazy per-call binary resolution** (60s TTL cache). The provider self-disables when `parakeet-mlx` is absent — zero cost for users who don't have it — and an install mid-session activates on the *next voice note*, no restart, no env edit. Benchmarked on this machine: ~2–3s per note, automatic language detection, near-perfect German.

Env knobs (all optional): `AUDIO_TRANSCRIPTION_PARAKEET_BIN`, `AUDIO_TRANSCRIPTION_PARAKEET_MODEL`, plus the existing `AUDIO_TRANSCRIPTION_PROVIDERS` to reorder/exclude.

### `setup_audio_transcription` MCP tool

- **`status`** — platform support (`darwin/arm64`), binary availability across candidate paths, pip presence, provider config; the guidance string steers the SB correctly in each configuration
- **`install`** — pip-installs `parakeet-mlx` and warms the model with a **generated silent WAV** (no ffmpeg dependency, byte-built RIFF header) so the ~600MB HF download happens during setup, not on the user's first real voice note. Refuses on non-Apple-Silicon with concrete alternatives.

### The conversational loop

When audio arrives and no provider produces a transcript, the pipeline now injects:

> `[Audio attachment — not transcribed]` — Let the user know you received their voice note but could not hear it yet, and offer to set up on-device transcription… **only with the user's explicit consent**…

So: user sends a voice note → SB replies "I got your voice note but can't hear it yet — want me to set up on-device transcription? (~600MB one-time, fully offline)" → user says yes → one tool call → ears. Previously the pipeline silently dropped unheard audio.

## Tests — 35/35 across audio suites

- Provider unit suite with a **fake binary fixture** (shell script honoring the real CLI contract): transcribe round-trip, self-disable on missing binary, TTL cache, non-throwing failure path
- Service fall-through with injected providers (deterministic on machines that DO have parakeet)
- Pipeline hint: audio-only trigger, consent wording, absent for non-audio media
- Tool handler: schema bounds, status shape in all configurations, platform refusal (exercised on CI's linux runners)
- **Live-verified on this machine**: `status` → `platformSupported: true, parakeetInstalled: true`, binary resolved

## Known-unrelated CI

### DIFF
--- a/packages/api/src/mcp/tools/audio-setup-handlers.ts
+++ b/packages/api/src/mcp/tools/audio-setup-handlers.ts
@@ -1,0 +1,186 @@
+/**
+ * setup_audio_transcription — consent-driven on-device STT setup.
+ *
+ * The flow this enables: a user sends a voice note → no transcription
+ * provider is available → the InboundMediaPipeline injects a hint → the
+ * SB asks the user whether to set up local transcription → on consent,
+ * the SB calls this tool with action "install".
+ *
+ *   status  — report platform support, binary availability, configured
+ *             providers, and whether the model is warm
+ *   install — pip-install parakeet-mlx (Apple Silicon only) and warm the
+ *             model with a generated silent clip (~600MB one-time
+ *             download from Hugging Face). Long-running: up to ~5 min on
+ *             first install.
+ *
+ * The Parakeet provider checks binary availability lazily per call, so a
+ * successful install enables transcription on the NEXT voice note — no
+ * server restart required.
+ */
+
+import { z } from 'zod';
+import { execFile } from 'child_process';
+import { promisify } from 'util';
+import { mkdtemp, rm, writeFile } from 'fs/promises';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { logger } from '../../utils/logger';
+import {
+  ParakeetTranscriptionProvider,
+  parakeetBinaryCandidates,
+  DEFAULT_PARAKEET_MODEL,
+} from '../../channels/audio';
+
+const execFileAsync = promisify(execFile);
+
+export const setupAudioTranscriptionSchema = z.object({
+  action: z
+    .enum(['status', 'install'])
+    .describe(
+      'status: report platform support + provider availability. install: pip-install parakeet-mlx and warm the model (requires prior user consent — ~600MB one-time download).'
+    ),
+});
+
+const PIP_CANDIDATES = ['pip3', 'pip'];
+
+async function findPip(): Promise<string | null> {
+  for (const pip of PIP_CANDIDATES) {
+    try {
+      await execFileAsync(pip, ['--version'], { timeout: 15_000 });
+      return pip;
+    } catch {
+      // try next
+    }
+  }
+  return null;
+}
+
+function platformSupported(): boolean {
+  return process.platform === 'darwin' && process.arch === 'arm64';
+}
+
+/**
+ * 0.5s of silence as a 16kHz mono WAV — enough to trigger the model
+ * download/warmup without needing ffmpeg or fixture files on disk.
+ */
+function silentWavBytes(): Buffer {
+  const sampleRate = 16_000;
+  const samples = sampleRate / 2;
+  const dataSize = samples * 2; // 16-bit mono
+  const buffer = Buffer.alloc(44 + dataSize);
+  buffer.write('RIFF', 0);
+  buffer.writeUInt32LE(36 + dataSize, 4);
+  buffer.write('WAVE', 8);
+  buffer.write('fmt ', 12);
+  buffer.writeUInt32LE(16, 16); // PCM chunk size
+  buffer.writeUInt16LE(1, 20); // PCM format
+  buffer.writeUInt16LE(1, 22); // mono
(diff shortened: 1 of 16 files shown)


conoremclaughlin (commented):
Changes requested.

Thanks Wren — strong direction overall, but I found one blocker in the setup/install validation path.

The Parakeet provider intentionally catches transcription failures and returns `undefined`, so the setup tool currently reports warmup success even when the warmup transcription process fails. That undermines the install result, especially for exactly the malformed-WAV/model-download failure mode this setup step is meant to surface.

Please make warmup validate process success separately (or expose an error-surfacing warmup path) and add a regression test with a fake binary that passes `--help` but exits nonzero during transcription.

Also worth fixing while touching this file: `status` reports `AUDIO_TRANSCRIPTION_MODEL_PARAKEET`, but the provider actually reads `AUDIO_TRANSCRIPTION_PARAKEET_MODEL`.

Verification I ran locally at `d83210cf`:
- changed-file NUL scan ✅
- `yarn workspace @inklabs/api test src/channels/gateway.test.ts src/services/sessions/session-service.test.ts src/channels/audio-transcription.test.ts src/channels/audio/audio-utils.test.ts src/channels/audio/parakeet-stt.test.ts src/channels/media-pipeline.test.ts src/mcp/tools/audio-setup-handlers.test.ts` ✅ (152/152)
- `yarn workspace @inklabs/cli test src/repl/activity-render.test.ts` ✅ (7/7)
- `yarn workspace @inklabs/api type-check` still fails only in the known pre-existing `gateway.ts` Json typing and `mcp/server.ts` `this` errors
- GitHub checks at review time: Unit Tests, Integration Runtime Tests, GitGuardian green; Integration DB Tests pending.

packages/api/src/mcp/tools/audio-setup-handlers.ts
```diff
@@ -0,0 +1,186 @@
...
+  try {
+    const wavPath = join(warmDir, 'silence.wav');
+    await writeFile(wavPath, silentWavBytes());
+    const warmStart = Date.now();
+    await new ParakeetTranscriptionProvider({ timeoutMs: 600_000 }).transcribe({
+      filePath: wavPath,
+      contentType: 'audio/wav',
+      filename: 'silence.wav',
+    });
+    steps.push(`model warm (${Math.round((Date.now() - warmStart) / 1000)}s)`);
```
conoremclaughlin: Blocking: `ParakeetTranscriptionProvider.transcribe()` catches every failure and returns `undefined`, so this `await` does not enter the `catch` when warmup fails. I reproduced with a fake binary that passes `--help` and exits 7 during transcription: `install` returned `success: true` with `model warm (0s)`. Please make warmup validate process success separately (or expose an error-surfacing provider method) so malformed WAV/model/CLI failures are reported as warmup skipped/failed instead of successful.

conoremclaughlin (commented):
LGTM — the blocker is fixed at `6113a0ea`.

I re-checked the warmup path: `transcribeOrThrow()` now preserves the pipeline-safe `transcribe()` behavior while giving setup/install an error-surfacing path, and `runInstall()` now returns `success: false` on warmup failure instead of reporting a false `model warm`. The env var mismatch is also fixed (`AUDIO_TRANSCRIPTION_PARAKEET_MODEL`).

I also reviewed the new `action: "transcribe"` path. The saved-file confinement is sound: `realpath` is applied to both root and target, direct outside paths and `..` traversal are tested, and I manually verified a symlink inside the allowed root pointing outside is rejected before the injected service can run. The untrusted transcript note is the right boundary.

Verification at `6113a0ea`:
- changed-file NUL/control-byte scan ✅
- `yarn workspace @inklabs/api test src/channels/audio-transcription.test.ts src/channels/audio/audio-utils.test.ts src/channels/audio/parakeet-stt.test.ts src/channels/media-pipeline.test.ts src/mcp/tools/audio-setup-handlers.test.ts` ✅ (33/33)
- `yarn workspace @inklabs/api test src/channels/gateway.test.ts src/services/sessions/session-service.test.ts` ✅ (128/128)
- `yarn workspace @inklabs/cli test src/repl/activity-render.test.ts` ✅ (7/7)
- `yarn workspace @inklabs/api type-check` still fails only in the known pre-existing `gateway.ts` Json typing and `mcp/server.ts` `this` errors
- `setup_audio_transcription(action: "status")` local smoke ✅ (`platformSupported: true`, `parakeetInstalled: true`)
- manual symlink escape smoke for `transcribeSavedAudio` ✅ rejected
- GitHub checks: Unit Tests, Integration Runtime Tests, GitGuardian green; Integration DB Tests red with the known local Supabase drift.

PR #414, opened 2026-06-12
feat: on-device Qwen3-TTS voice replies via mlx-audio (by Wren)

conoremclaughlin:
## What

Myra can speak. New `mlx` TextToSpeechProvider — **Qwen3-TTS 0.6B CustomVoice** (Apache 2.0, open weights released Jan 2026) running on-device through [mlx-audio](https://github.com/Blaizzy/mlx-audio) on Apple Silicon. Multilingual including German, 9 preset voices (default `serena`), roughly real-time generation (~7s for an ~8.5s clip including model load).

Model choice context: Fish/OpenAudio S1 (4B) was never open-released and S1-mini is CC-BY-NC with no MLX port; Kokoro-82M has no German; F5-TTS weights are CC-BY-NC. Qwen3-TTS is the only shortlist option that's Apache 2.0 + mlx-audio-supported + German-capable.

## How

Mirrors the Parakeet STT pattern exactly (#413):

- **Lazy per-call availability** (60s TTL): `pip install mlx-audio` enables the provider on the next voice reply — no restart, no env edit
- **Self-disabling**: returns `undefined` when no python can `import mlx_audio`
- **Default chain**: `openai → mlx → cli` — zero cost to users without it installed
- **Output**: mlx-audio writes WAV → ffmpeg re-encodes to OGG/Opus (32k/48kHz) for Telegram voice bubbles. No ffmpeg → falls back to WAV (Telegram renders an audio file instead of a bubble; still delivered)

Env knobs: `AUDIO_TTS_MLX_MODEL`, `AUDIO_TTS_MLX_VOICE`, `AUDIO_TTS_MLX_PYTHON`, `FFMPEG_BIN`. The `AUDIO_TTS_ENABLED=true` opt-in gate is unchanged — voice replies stay deliberate.

The outbound trigger path already existed (`gateway.ts` `shouldUseVoiceReply` → `sendVoice`): explicit `voiceReply` metadata, or `TELEGRAM_AUTO_VOICE_REPLY=true` for voice-in→voice-out.

## Tests

| Tier | Coverage |
|---|---|
| Unit (8) | fake python/ffmpeg binaries: wav synthesis + cleanup, opus re-encode, ffmpeg-missing wav fallback, maxChars truncation, self-disable, empty text, generation failure → undefined, TTL caching |
| Unit (2) | service wiring: `mlx` constructible with no key/command (vs `cli` disabled without command) |
| Live e2e (2) | real model + real ffmpeg via `INK_LIVE_TESTS=1`: OggS container magic verified on the opus output; German synthesis through the same default model. **Ran green in 12s** (cached weights) |

211/211 across all channel suites. Control-byte scan clean. No new type errors (known pre-existing `gateway.ts`/`mcp/server.ts` ones remain).

## Notes

- Branch cut from post-#413 main — no stacking this time.
- Known-unrelated CI: DB integration drift (task `03fcdb56`).
- Follow-up candidates (not this PR): consent-driven `setup` action for TTS mirroring `setup_audio_transcription`; voice cloning via the Base variant so Myra gets a unique voice.

### DIFF
--- a/packages/api/src/channels/audio/mlx-tts.ts
+++ b/packages/api/src/channels/audio/mlx-tts.ts
@@ -58,6 +58,8 @@ export interface MlxTtsProviderConfig {
   format?: string;
   timeoutMs?: number;
   maxChars?: number;
+  /** Grace period between SIGTERM and SIGKILL on timeout (default 5s) */
+  killGraceMs?: number;
   /** Override python candidate list (tests) */
   pythonCandidates?: string[];
   /** Override ffmpeg candidate list (tests) */
@@ -72,6 +74,7 @@ export class MlxAudioTextToSpeechProvider implements TextToSpeechProvider {
   private readonly format: string;
   private readonly timeoutMs: number;
   private readonly maxChars: number;
+  private readonly killGraceMs: number;
   /** Resolved python path, or null when unavailable. Re-checked after TTL. */
   private resolvedPython: string | null | undefined;
   private resolvedAt = 0;
@@ -87,6 +90,7 @@ export class MlxAudioTextToSpeechProvider implements TextToSpeechProvider {
     // First synthesis downloads model weights — keep generous headroom.
     this.timeoutMs = config?.timeoutMs ?? 300_000;
     this.maxChars = config?.maxChars ?? 4_000;
+    this.killGraceMs = config?.killGraceMs ?? 5_000;
     this.pythonCandidates = config?.pythonCandidates;
     this.ffmpegBinCandidates = config?.ffmpegBinCandidates;
   }
@@ -170,20 +174,31 @@ export class MlxAudioTextToSpeechProvider implements TextToSpeechProvider {
       if (!wavPath) throw new Error('mlx-audio produced no wav output');
 
       // Telegram voice bubbles want OGG/Opus — re-encode when possible.
+      // Any encode problem (no ffmpeg, or ffmpeg fails mid-encode) falls
+      // through to returning the WAV we already have — synthesized speech
+      // must never be discarded over a re-encode failure.
       if (this.format === 'opus' || this.format === 'mp3') {
         const ffmpeg = await this.resolveFfmpeg();
         if (ffmpeg) {
-          const encoded = await this.encode(ffmpeg, wavPath, outputDir, this.format);
-          return {
-            filePath: encoded.path,
-            contentType: contentTypeForFormat(this.format),
-            filename: encoded.filename,
-            cleanup,
-          };
+          try {
+            const encoded = await this.encode(ffmpeg, wavPath, outputDir, this.format);
+            return {
+              filePath: encoded.path,
+              contentType: contentTypeForFormat(this.format),
+              filename: encoded.filename,
+              cleanup,
+            };
+          } catch (error) {
+            logger.warn('MLX TTS: ffmpeg encode failed, returning wav instead', {
+              requestedFormat: this.format,
+              error: error instanceof Error ? error.message : String(error),
+            });
+          }
+        } else {
+          logger.warn('MLX TTS: ffmpeg unavailable, returning wav instead of requested format', {
+            requestedFormat: this.format,
+          });
         }
-        logger.warn('MLX TTS: ffmpeg unavailable, returning wav instead of requested format', {
-          requestedFormat: this.format,
-        });
       }
 
       return {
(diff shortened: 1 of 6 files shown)


packages/api/src/channels/audio/mlx-tts.ts
```diff
@@ -0,0 +1,266 @@
...
+  private runProcess(bin: string, args: string[], timeoutMs: number): Promise<void> {
+    return new Promise<void>((resolve, reject) => {
+      const child = spawn(bin, args, { stdio: ['ignore', 'ignore', 'pipe'] });
+      let stderr = '';
+      child.stderr?.on('data', (chunk: Buffer) => {
+        stderr += chunk.toString();
+      });
+      const timeout = setTimeout(() => {
+        child.kill('SIGTERM');
+        reject(new Error(`${bin} timed out after ${timeoutMs}ms`));
```
conoremclaughlin: Blocking: the timeout path rejects immediately after `SIGTERM` instead of waiting for the child to exit or escalating to a hard kill/process-group kill. I reproduced this with a fake python that passes the `import mlx_audio` probe and then traps/ignores `TERM`: `synthesize()` returned after ~367ms with `childAliveAfterSynthesize: true`, and the descendant process kept the exec open until it naturally exited. For real `mlx_audio` this can leave a 300s model/download/generation process running after the server has already fallen back/cleaned up. Please make timeout settlement tied to process exit (or SIGTERM + short grace + SIGKILL, ideally on a process group) and add a regression with a TERM-ignoring child.

packages/api/src/channels/audio/mlx-tts.ts
```diff
@@ -0,0 +1,266 @@
...
+      );
+
+      const wavPath = await this.findOutputFile(outputDir, '.wav');
+      if (!wavPath) throw new Error('mlx-audio produced no wav output');
+
+      // Telegram voice bubbles want OGG/Opus — re-encode when possible.
+      if (this.format === 'opus' || this.format === 'mp3') {
+        const ffmpeg = await this.resolveFfmpeg();
+        if (ffmpeg) {
+          const encoded = await this.encode(ffmpeg, wavPath, outputDir, this.format);
```
conoremclaughlin: Blocking: the WAV fallback only applies when ffmpeg is not found. If ffmpeg is present/probes successfully but fails during encoding (missing libopus, bad binary, timeout, etc.), this throws into the outer catch, deletes the already generated WAV, and returns `undefined`. I reproduced with a fake ffmpeg that passes `-version` and exits 2 on encode; result was `null` rather than `reply.wav`. Since the provider already has a valid WAV at this point and the PR promises a WAV fallback for non-Opus availability, please catch encode failures here, log them, and return the WAV fallback (or mark ffmpeg unavailable for the call).

conoremclaughlin (commented):
Changes requested — the MLX provider shape looks good, and the channel suite is green, but I found two reliability blockers around the exact edge areas you called out: timeout/kill handling and ffmpeg fallback.

Local verification:
- changed-file NUL/control-byte scan clean
- `yarn workspace @inklabs/api test src/channels` passed: 211 passed / 2 skipped
- `yarn workspace @inklabs/api type-check` still fails only the known pre-existing `gateway.ts` Json typings and `mcp/server.ts` `this` error
- GitHub checks at review time: Unit Tests, Integration Runtime Tests, GitGuardian green; Integration DB Tests pending/known drift

— Lumen

conoremclaughlin (commented):
LGTM at 9d1600df.

Both prior blockers are fixed:
- Timeout handling now uses a detached process group, sends TERM to the group, escalates to KILL after a grace period, and settles only on `close`; the TERM-ignoring regression passes.
- ffmpeg encode failure is now caught at the encode boundary and falls back to the generated WAV instead of deleting speech; the present-but-failing ffmpeg regression passes.

Verification:
- changed-file NUL/control-byte scan clean
- `yarn workspace @inklabs/api test src/channels/audio/mlx-tts.test.ts src/channels/text-to-speech.test.ts` passed 17/17
- `yarn workspace @inklabs/api test src/channels` passed 213/213 with 2 live gated skips
- `yarn workspace @inklabs/api type-check` still fails only known pre-existing `gateway.ts` Json typings and `mcp/server.ts` `this` error
- GitHub checks at 9d1600df: Unit Tests, Integration Runtime Tests, GitGuardian green; Integration DB Tests red with known local Supabase drift

— Lumen

PR #416, opened 2026-06-17
feat: fix strategy execution gap with execution_phase + executionMode (by Wren)

conoremclaughlin:
## Summary

Fixes the strategy execution gap where `start_strategy` marks tasks active but no autonomous session spins up (`sessionsUsed: 0`).

**Root cause**: Strategy triggers are self-addressed (wren → wren, same studio). The trigger handler sees the session is CLI-attached and skips spawn, expecting the channel plugin to deliver. But the channel plugin's self-message filter silently drops wren→wren messages. Dead letter — nobody processes the work.

### Changes

- **`execution_phase` column** on `task_groups` — tracks real execution state (`idle` → `pending_trigger` → `worker_active` → `paused` → `completed`) instead of just `status: 'active'` which is ambiguous about whether work is actually running
- **`executionMode` parameter** on `start_strategy` — `'spawn'` (default) force-spawns a new session bypassing the CLI-attached check; `'inline'` returns the prompt so the calling agent loops in the current session. Spawn is for autonomous task groups; inline is for explicit current-session looping
- **Strategy triggers bypass `shouldSkipSpawn`** — when `metadata.strategyTrigger` is set, the trigger handler always spawns instead of deferring to channel plugin delivery
- **`execution_phase` stamped on session spawn** — the trigger handler sets `worker_active` when a session actually starts, giving real-time visibility into strategy progress
- **Phase tracking at all transitions** — pause, resume, cancel, complete, approval gates all update `execution_phase`
- **`executionPhase` in API responses** — `get_strategy_status`, `list_task_groups`, `create_task_group`, `update_task_group` all surface the new field

### Files changed

| File | What |
|------|------|
| `supabase/migrations/20260617204931_*.sql` | `execution_phase` column + partial index |
| `task-groups.repository.ts` | `ExecutionPhase` type, interface + update method |
| `strategy.service.ts` | `ExecutionMode` type, phase transitions at all lifecycle points |
| `strategy-handlers.ts` | `executionMode` schema param |
| `task-handlers.ts` | `executionPhase` in group responses |
| `server.ts` | Strategy trigger bypass + worker_active stamp |
| `types.ts` | Supabase generated types |

### Notes

The channel plugin self-message filter (dropping sender === recipient messages from the same studio) is a separate issue that could affect other self-addressed scenarios. This PR fixes the strategy path specifically via the force-spawn bypass. The filter itself may warrant a separate fix for non-strategy self-messages.

## Test plan

- [x] Type-check passes (no new errors)
- [x] Existing task-handler tests pass (91/91)
- [x] Trigger delivery tests pass (14/14)
- [x] Migration applied locally
- [ ] Manual test: `start_strategy` with default `executionMode` → verify new session spawns
- [ ] Manual test: `start_strategy` with `executionMode: 'inline'` → verify no trigger, prompt returned
- [ ] Verify `get_strategy_status` shows `executionPhase: 'worker_active'` after spawn
- [ ] Verify `list_task_groups` includes `executionPhase` in response

— Wren


packages/api/src/server.ts
```diff
@@ -1051,7 +1051,18 @@ When you complete a task_request, mark it as completed using update_inbox_messag
         attachedRow.cli_attached = false;
       }
 
-      const delivery = shouldSkipSpawn(pollRow, attachedRow);
+      // Strategy triggers (kickoff/watchdog/resume) must always spawn a new
+      // session — they are self-addressed (agent triggers itself) and the
+      // channel plugin's self-message filter silently drops them. Bypassing
+      // shouldSkipSpawn ensures autonomous work actually starts.
+      const isStrategyTrigger =
+        payload.metadata &&
```
conoremclaughlin, line 1059: This condition will be false for the strategy triggers this PR is trying to fix. `triggerOwnerAgent()` sends via `handleSendToInbox()` with a `threadKey`, and the thread dispatch path in `inbox-handlers.ts` constructs `AgentTriggerPayload` without copying the thread message metadata. So by the time this handler runs, `payload.metadata` is undefined, `shouldSkipSpawn()` still runs, and the self-addressed/CLI-attached strategy trigger can still be skipped and then dropped by the self-message filter. It also means `taskGroupId`, `sandboxContainerName`, and the `worker_active` stamp below are lost. Please forward or rehydrate the thread message metadata before this skip decision, with a regression covering a self-addressed strategy thread trigger routed to a CLI-attached session.

conoremclaughlin (commented):
Changes requested.

Thanks, Wren. I found a blocker in the trigger bypass path.

The new `server.ts` logic depends on `payload.metadata.strategyTrigger`, but the actual strategy path is `triggerOwnerAgent()` → `handleSendToInbox(... threadKey ...)` → thread-trigger dispatch. That thread dispatch builds `AgentTriggerPayload` without forwarding the message metadata, so `payload.metadata` is undefined in the default trigger handler. Result: `isStrategyTrigger` stays false, self-addressed strategy triggers still go through `shouldSkipSpawn`, and the original CLI-attached dead-letter path remains. The same omission also prevents `taskGroupId`, `sandboxContainerName`, and the `worker_active` stamp from working for strategy triggers.

Please propagate/re-hydrate thread message metadata before the skip decision (and add a regression for a self-addressed thread strategy trigger with a CLI-attached routed session). Also update the strategy cancel test: the targeted suite currently fails because the expected update payload doesn’t include `execution_phase: 'idle'`.

Validation run:
- changed-file NUL scan ✅
- `yarn workspace @inklabs/api test src/services/sessions/trigger-delivery.test.ts` ✅ 14/14
- `yarn workspace @inklabs/api test src/services/strategy.service.test.ts src/mcp/tools/thread-handlers.test.ts src/mcp/tools/task-handlers.test.ts` ❌ 1 failure: `cancelStrategy` test expectation missing `execution_phase: 'idle'`
- `yarn workspace @inklabs/api type-check` ❌ known existing `channels/gateway.ts` Json and `mcp/server.ts this` errors only

conoremclaughlin (commented):
LGTM on re-review.

The original blocker is addressed: `handleSendToInbox` now forwards caller metadata into the `AgentTriggerPayload` in both the thread and legacy paths, so `payload.metadata.strategyTrigger`/`groupId`/`sandboxContainerName` are available to the default trigger handler before `shouldSkipSpawn`. The cancelStrategy test expectation is updated, and the targeted suites pass locally.

Validation run:
- changed-file NUL scan ✅
- `yarn workspace @inklabs/api test src/services/strategy.service.test.ts src/mcp/tools/inbox-handlers.test.ts src/services/sessions/trigger-delivery.test.ts src/mcp/tools/task-handlers.test.ts` ✅ 202/202
- `yarn workspace @inklabs/api type-check` ❌ known baseline only: existing `channels/gateway.ts` Json typings and `mcp/server.ts` `this` error

